PR #38, opened 2020-03-12
Add helm-template function

### LINKED ISSUE #24: Add a Helm expand source
@pwittrock FYI

prachirp:
- Add helm-template kpt function
- Add e2e test script for this repo at tests/e2e.sh
- Add publish-functions script to publish this repo's functions to catalog
- Copy example-configs from sdk
- Check in two helm charts under sh/test-data/helm for testing. Keep these separate from example-configs since they are not hydrated yaml.

Fixes #24


sh/src/helm-template
```diff
@@ -0,0 +1,100 @@
...
+
+create_tmp() {
+  tmp=$(mktemp -d "/tmp/helm-template.XXXXXXXX")
+  # If there is input through stdin then write it to tmp
+  if [[ ! -t 0 ]]; then
+    cat "-" | kpt fn sink "$tmp"
+  fi
+}
+
+delete_tmp() {
```
frankfarzan: why do need to delete this given it only exists in the ephemeral container?

sh/src/helm-template
```diff
@@ -0,0 +1,100 @@
...
+}
+
+run_main() {
+  read_arguments "$@"
+  create_tmp
+  expand_helm
+  delete_tmp
+}
+
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]
```
frankfarzan: why do you need this conditional?

sh/src/helm-template
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
```
frankfarzan: I suggest organzing this repo by function instead:

helm/
--  Dockerfile
--  helmp-template.sh
-- test

sh/build/helm-template.Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM alpine/helm:3.1.1
+
+RUN apk add bash curl git
+RUN apk update
+
+RUN curl -fsSL -o /usr/local/bin/kpt https://storage.googleapis.com/kpt-dev/latest/linux_amd64/kpt
+RUN chmod +x /usr/local/bin/kpt
+ENV PATH /usr/local/bin:$PATH
```
frankfarzan: why is this needed? What's the default value of PATH

sh/test-data/helm/mongodb/templates/NOTES.txt
```diff
@@ -0,0 +1,96 @@
+This Helm chart is deprecated
```
frankfarzan: ?

sh/test-data/helm/mongodb/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
frankfarzan: This needs to have a licnece? Also if you can avoid checking it in the repo, it's preferred. You can curl it for testing,

tests/e2e.sh
```diff
@@ -0,0 +1,81 @@
...
+
+function assert_dir_exists() {
+  [[ -d $1 ]] || fail "Dir not exist: $1"
+}
+
+############################
+# Docker Tests
+############################
+
+[[ -z ${NODOCKER} ]] || {
```
frankfarzan: Remove this and the variable as this is only does docker testing

tests/e2e.sh
```diff
@@ -0,0 +1,81 @@
...
+testcase "docker_helm_template_usage_too_few_args"
+docker run -v "$(pwd)/helm":/source gcr.io/kpt-functions/helm-template:"${TAG}" name=too-few-args 2>err.txt || true
+assert_contains_string err.txt "${HELM_USAGE_SNIPPET}"
+
+testcase "docker_helm_template_usage_too_many_args"
+docker run -v "$(pwd)/helm":/source gcr.io/kpt-functions/helm-template:"${TAG}" chart_path=/source/redis name=too-many-args extra=args_provided 2>err.txt || true
+assert_contains_string err.txt "${HELM_USAGE_SNIPPET}"
+
+testcase "docker_helm_template_correct_args"
+docker run -v "$(pwd)/helm":/source gcr.io/kpt-functions/helm-template:"${TAG}" name=correct-args chart_path=/source/redis >out.yaml
```
frankfarzan: Extract pwd/helm into a variable and reuse

tests/e2e.sh
```diff
@@ -0,0 +1,81 @@
...
+assert_dir_exists mongodb
+assert_dir_exists redis
+
+############################
+# kpt fn Tests
+############################
+
+testcase "kpt_helm_template_imperative"
+kpt fn source example-configs |
+  docker run -v "$(pwd)/helm":/source gcr.io/kpt-functions/helm-template chart_path=/source/mongodb name=my-mongodb |
```
frankfarzan: What about using --image?

Also, test using ConfigMap functionConfig?

sh/src/helm-template
```diff
@@ -0,0 +1,100 @@
...
+  fi
+}
+
+delete_tmp() {
+  rm -rf "$tmp"
+}
+
+expand_helm() {
+  # Overwrite files if they exist
+  if ! output="$(helm template "${name}" "${chartPath}" -f "${valuesPath}" --output-dir "${tmp}" )"; then
```
frankfarzan: Problem with wrapping helm cli like this is that you're preventing user from specify any other flag to template:

https://helm.sh/docs/helm/helm_template/

Instead, I think you want to only handle and redirect --output.

example-configs/README.md
```diff
@@ -0,0 +1 @@
+Example k8s configs.
```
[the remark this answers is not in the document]
prachirp: Done

sh/src/helm-template
prachirp: Good point, I was just following it as a best practice. Removed the delete_tmp function now.

sh/src/helm-template
prachirp: It's a way to tell if the script is being sourced. But this script will always be executed in a docker container, never sourced, so I removed the conditional.

sh/src/helm-template
prachirp: Changed, also removed test data based on below comment.

sh/build/helm-template.Dockerfile
prachirp: It's a best practice from Docker docs, "use ENV to update the PATH environment variable for the software your container installs"

sh/test-data/helm/mongodb/templates/NOTES.txt
prachirp: They moved to the bitnami charts repo 2 days ago. I now use the bitnami charts repo.

sh/test-data/helm/mongodb/Chart.yaml
prachirp: The bitnami charts use the Apache 2.0 license, but that's fair that we want to avoid checking it in. I now get it only for testing.

tests/e2e.sh
prachirp: Done

tests/e2e.sh
prachirp: It doesn't really make sense extracting the $(pwd) since we'd have to declare a new variable for each test case.  I extracted the common parts of the path to a variable.

tests/e2e.sh
prachirp: Using --image doesn't work, working with @pwittrock to investigate.

sh/src/helm-template
prachirp: I agree we prevent the user from specifying options. I'd rather check in this MVP first and address this in a separate change.

sh/build/helm-template.Dockerfile
frankfarzan: But you're using a well-known bin directory, not a user-defined one.

tests/e2e.sh
frankfarzan: This is about supporting running priviledged containers

sh/src/helm-template
frankfarzan: That's fine. I would look into parsing $@ and passing the rest to Helm. This script should ideally only know about the output flag. In fact, in the most general form, this script can be reused for wrapping any CLIs like Helm template.

sh/helm-template/helm-template
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
```
frankfarzan: 1. This file needs a .sh extension
2. Directory name already qualifies this script, so rename it to something else like `run.sh`.
3. Rename top-level `sh` dir to `functions`

frankfarzan: Remember to add LICENSE headers to all new files

sh/helm-template/helm-template.Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM alpine/helm:3.1.1
```
frankfarzan: rename this to just Dockerfile

sh/src/helm-template
frankfarzan: I'm fine with following up on a separate PR

tests/e2e.sh
prachirp: Added todo

sh/helm-template/helm-template
prachirp: Done

sh/helm-template/helm-template.Dockerfile
prachirp: Done

sh/helm-template/helm-template
prachirp: Done

scripts/publish-functions.sh
```diff
@@ -0,0 +1,30 @@
...
+
+REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. >/dev/null 2>&1 && pwd)"
+TAG=${TAG:-dev}
+FUNCS=(helm-template)
+
+for FN_NAME in "${FUNCS[@]}"; do
+  DOCKER_IMAGE="gcr.io/kpt-functions/${FN_NAME}:${TAG}"
+  DOCKER_FILE="${REPO}/functions/${FN_NAME}/Dockerfile"
+  docker build -t "$DOCKER_IMAGE" -f "$DOCKER_FILE" functions
+	docker push "$DOCKER_IMAGE"
```
frankfarzan: indentation issue

tests/e2e.sh
```diff
@@ -67,6 +81,7 @@ assert_dir_exists redis
 # kpt fn Tests
 ############################
 
+# TODO: Add kpt_helm_template_short test after kpt fn run --image fix <https://github.com/GoogleContainerTools/kpt/issues/359>
```
frankfarzan: Also a test for declarative run

scripts/publish-functions.sh
prachirp: Done

tests/e2e.sh
prachirp: Done

PR #56, opened 2020-06-16
Transfer kubeval function and add tests

prachirp:
Add unit tests and make small modifications to dockerfile
Add kpt e2e tests
Fixes GoogleContainerTools/kpt#578


tests/e2e.sh
```diff
@@ -100,3 +99,47 @@ docker run -u "$(id -u)" -v "$(pwd)/${CHARTS_SRC}":/source gcr.io/kpt-functions/
...
+  annotations:
+    config.k8s.io/function: |
+      container:
+        image:  gcr.io/kpt-functions/kubeval:dev
+        network:
+          required: true
+    config.kubernetes.io/local-config: 'true'
+EOF
+kpt fn source fixtures/valid* |
+kpt fn run --fn-path fc.yaml --network >out.yaml
```
frankfarzan, line 123: What's the exit code if there's an unexpected failure? In that case, wouldn't you exit here and never run subsequent lines?

functions/ts/src/kubeval.ts
```diff
@@ -0,0 +1,173 @@
+import {
```
frankfarzan, line 1: Note: this removes the Git history showing this was contributed by David. Maybe we should rethink extracting out functions into this repo. Instead we can have a demo-functions and real-functions in the SDK repo. WDYT? This would also simplify how many repos we have to touch.

tests/e2e.sh
```diff
@@ -100,3 +99,47 @@ docker run -u "$(id -u)" -v "$(pwd)/${CHARTS_SRC}":/source gcr.io/kpt-functions/
 assert_dir_exists default
 assert_contains_string default/secret_my-mongodb.yaml "my-mongodb"
 assert_contains_string default/secret_my-redis.yaml "my-redis"
+
+############################
```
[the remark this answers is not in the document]
prachirp: I intend for this section to hold all kpt tests. Each testcase name includes the name of the function.

tests/e2e.sh
prachirp: Yes, you error out with exit status 1 and the whole e2e test fails. See line 144 for the failure condition.

tests/e2e.sh
```diff
@@ -77,7 +76,7 @@ metadata:
   annotations:
     config.k8s.io/function: |
       container:
-        image:  gcr.io/kpt-functions/helm-template
+        image:  gcr.io/kpt-functions/helm-template:dev
```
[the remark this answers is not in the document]
prachirp: It's still blocked on https://github.com/GoogleContainerTools/kpt/issues/587 but I will add a TODO.

functions/ts/src/kubeval.ts
prachirp: I'd rather keep real functions in this repo, it seems easier to modularize this way and we can include not just typescript functions but any others like shell, starlark, etc.

prachirp:
cc @dflemstr We are considering that the catalog repo is a more natural place for functions. Can you please leave a comment for googlebot?

dflemstr:
@googlebot I consent.

PR #57, opened 2020-06-16
Set namespace function and e2e test

prachirp:
Users currently have no way to set the namespace of all resources when using Config Sync with unstructured layout.
This function provides a way for users to easily set the namespace field.
Fixes GoogleContainerTools/kpt#683

### DIFF
--- a/functions/starlark/set_namespace.star
+++ b/functions/starlark/set_namespace.star
@@ -1,0 +1,10 @@
+# set the namespace on each resource
+def run(r, ns_value):
+  for resource in r:
+    # mutate the resource
+    resource["metadata"]["namespace"] = ns_value
+
+# get the value of the annotation to add
+ns_value = ctx.resource_list["functionConfig"]["spec"]["namespace_value"]
+
+run(ctx.resource_list["items"], ns_value)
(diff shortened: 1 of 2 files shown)


prachirp:
cc @mortent @phani I made this function as a workaround for users to set namespaces until we decide on a longer term solution using kpt cfg or within Config Sync. PTAL

mortent:
So this would just add namespace to all resources, regardless of whether they have cluster or namespace scope? That is not ideal, but I don't think it will create any serious problems. Have we verified what the behavior is in that case?

prachirp:
Yes this just adds a namespace to all resources. This may cause tools to give warnings that cluster scoped resources have a namespace specified. In such a case users can use the remove namespace fn, which removes namespace from a hardcoded list of resources which it knows should be cluster scoped (https://github.com/kubeflow/kfctl/tree/master/kustomize-fns/remove-namespace).

functions/starlark/set_namespace.star
```diff
@@ -0,0 +1,10 @@
+# set the namespace on each resource
```
frankfarzan, line 1: Is this documented in the catalog? What prompted this? Also we've discussed adding a toolchain column to the catalog  (Can be follow up after the move).

prachirp: This is not documented in the catalog, I will add it once it's committed. This was prompted by [this issue](https://github.com/GoogleContainerTools/kpt/issues/683) a user filed noting that there was no easy way to add the namespace field to unstructured configs in order to use them with Config Sync.

The new PR will add a toolchain column to the catalog [here](https://github.com/prachirp/kpt/blob/update-general-docs/site/content/en/guides/consumer/function/catalog/_index.md)

PR #69, opened 2020-07-11
ARG instead of EVN to set component version

aodinokov:
Probably this will be a helpful thing:

This change allows to override the value
of the component versions, e.g.[1].
According to [2] ENV values persist in
Dockerfile and ARG can be redefinded with
--build-arg parameter, that probably would
be great to do without changes in the
Dockerfile itself.

[1]
functions/ts$ docker build -f build/kustomize_build.Dockerfile --build-arg KUSTOMIZE_VERSION=v3.8.0 ./

[2]
https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg

### DIFF
--- a/functions/ts/build/helm_template.Dockerfile
+++ b/functions/ts/build/helm_template.Dockerfile
@@ -2,7 +2,7 @@ FROM node:lts-alpine as builder
 
 RUN apk add bash curl git && apk update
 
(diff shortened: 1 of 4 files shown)


aodinokov:
@googlebot I signed it!

PR #71, opened 2020-07-16
Add example usage for helm template

prachirp:
Merge after kpt release uses kyaml v0.4.1

Fixes GoogleContainerTools/kpt#587


examples/helm-template/README.md
```diff
@@ -0,0 +1,36 @@
...
+The function is invoked by authoring a function configuration in local-configs/example.yaml
+with `metadata.annotations.[config.kubernetes.io/function]` set to
+`gcr.io/kpt-functions/helm-template`.
+
+Get this example and try it out by running the following commands:
+
+```sh
+kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/helm-template .
+cd helm-template; kpt cfg cat local-configs
+kpt fn run local-configs --mount type=bind,src=$(pwd)/helloworld-chart,dst=/source
```
frankfarzan, line 16: Can you simplify this to:

kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/helm-template .
kpt fn run helm-template/local-configs --mount type=bind,src=$(pwd)/helloworld-chart,dst=/source

you can do `kpt cfg cat` afterwards if reader is interested in learning what's in the package.

examples/helm-template/README.md
```diff
@@ -0,0 +1,36 @@
+# Helm Template
+
+This is an example of invoking the helm template function using declarative configuration.
+
+## Function invocation
+
+The function is invoked by authoring a function configuration in local-configs/example.yaml
+with `metadata.annotations.[config.kubernetes.io/function]` set to
```
frankfarzan, line 8: I'm not a fan of these of sentences:

Some words, with 'field X' in some other words in some 'file Foo'.

Might as well just show the YAML file instead of translating it into prose (which is less precise and more confusing).

examples/helm-template/README.md
```diff
@@ -0,0 +1,36 @@
+# Helm Template
+
+This is an example of invoking the helm template function using declarative configuration.
```
frankfarzan, line 3: What does this do? What's the use case? Why do I care about this?

examples/helm-template/README.md
```diff
@@ -0,0 +1,36 @@
+# Helm Template
```
frankfarzan, line 1: How's this meshing with this doc:

https://googlecontainertools.github.io/kpt/guides/consumer/function/catalog/docs/helm-template/usage/

examples/helm-template/helloworld-chart/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
```
frankfarzan, line 1: these should be used in the tests:

https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/tests/e2e.sh


We should create conventions for where each of the following lives for each function in this repo:

1. Function Source
2. Function Examples
3. Functions e2e tests

examples/helm-template/README.md
prachirp: Simplified

examples/helm-template/README.md
prachirp: Removed, hope it's less confusing.

examples/helm-template/README.md
prachirp: Give higher level motivation now.

examples/helm-template/README.md
prachirp: Will integrate in a separate docs PR to kpt repo.

examples/helm-template/helloworld-chart/Chart.yaml
prachirp: Test helm template using kpt in e2e script now.

PR #73, opened 2020-07-20
Declarative example of kustomize build and e2e tests

prachirp:


prachirp:
cc @aodinokov PTAL, this example is modified from yours in https://github.com/kubernetes-sigs/kustomize/pull/2709

PR #76, opened 2020-07-23
Remove `release-go-functions-` from the docker image tag

haiyanmeng:


haiyanmeng:
> Can you comment the file with more explanations? This file is confusing if I were to revisit it.

Done

PR #79, opened 2020-07-25
testing kustomize_build in ci

aodinokov:
do not merge
this is just to help with https://github.com/GoogleContainerTools/kpt/issues/856
later I'll abandon this change


aodinokov:
closing since we've proved that kpt actually works and https://github.com/GoogleContainerTools/kpt/issues/856 was an issue with permissions

PR #87, opened 2020-08-10
Add annotate-config blueprint

prachirp:


prachirp:
cc @frankfarzan PTAL

examples/annotate-config/README.md
```diff
@@ -0,0 +1,22 @@
+# Annotate Config
+
+The `annotate-config` KRM config function adds an annotation to configuration.
```
frankfarzan, line 3: configuration -> resources

examples/annotate-config/README.md
```diff
@@ -0,0 +1,22 @@
...
+This example invokes the annotate-config function using declarative
```
frankfarzan, line 4: This sentence is repeated in all examples and doesn't convey much to the user. Please remove everywhere.

examples/annotate-config/README.md
```diff
@@ -0,0 +1,22 @@
...
+configuration.
+
+## Function invocation
+
+Get this example and try it out by running the following commands:
+
+```sh
+kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/annotate-config .
+kpt fn run annotate-config/configs --fn-path annotate-config/functions
```
frankfarzan, line 13: Why do you need `--fn-path`. Try to show the simplest CLI usage.

prachirp: As with set-namespace, it's cleaner not to add an annotation to the function config. I only want add an annotation to the configs in the package dir so I put the function config in a separate fn-path.

examples/annotate-config/README.md
prachirp: Done in all outstanding blueprints PRs

examples/annotate-config/README.md
prachirp: Done in all outstanding blueprints PRs

examples/annotate-config/README.md
frankfarzan: This is trying to bypass a deeper issue here: You don't want the function to manipulate the function config.

prachirp: I agree it's a bypass but let's merge these examples now and update them later after there's consensus around a solution like the no-process annotation.

frankfarzan (changes requested):

prachirp:
@frankfarzan If I move the function config under annoate-config/configs then it's being mutated and gets the new annotation `configmanagement.gke.io/namespace-selector: sre-supported`. This issues is already covered by https://github.com/GoogleContainerTools/kpt/issues/630 and will be solved when we get consensus around no-process. In the meantime, let's merge this example with function config in a separate directory.

frankfarzan:
GoogleContainerTools/kpt#630 is a different issue, it's about the inventory object being processed. Function configs are different. Let's file a separate bug for this.

prachirp:
@frankfarzan Captured this in https://github.com/GoogleContainerTools/kpt/issues/1122

frankfarzan (approved):

PR #89, opened 2020-08-11
Add set-namespace blueprint

prachirp:


examples/set-namespace/README.md
```diff
@@ -0,0 +1,22 @@
...
+Get this example and try it out by running the following commands:
+
+```sh
+kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/set-namespace .
+kpt fn run set-namespace/configs --fn-path set-namespace/functions
+```
+
+## Expected result
+
+Check all resources have `metadata.namespace` set to `example-ns`:
```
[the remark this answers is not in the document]
prachirp: Documented limitation

prachirp:
cc @frankfarzan PTAL

examples/set-namespace/README.md
```diff
@@ -0,0 +1,26 @@
...
+cluster-scoped]. This example invokes the set-namespace function using
+declarative configuration.
+
+## Function invocation
+
+Get this example and try it out by running the following commands:
+
+```sh
+kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/set-namespace .
+kpt fn run set-namespace/configs --fn-path set-namespace/functions
```
frankfarzan, line 14: why do you need `fn-path`?

prachirp: I think it's cleaner not to set the namespace on the function config, only on the configs in the package dir set-namespace/configs. So I put the function config in a separate fn-path set-namespace/functions.

frankfarzan: Underlying issue is that function configs should not be processed by functions. That seems like a bug and may be related to https://github.com/GoogleContainerTools/kpt/issues/630

prachirp: Same as for annotate-config example. Let's merge these now and update them when there's consensus.

prachirp:
@frankfarzan PTAL

frankfarzan (approved):

PR #93, opened 2020-08-18
Avoid setting the namespace field for cluster-scoped resources

haiyanmeng:


functions/go/set-namespace/main.go
```diff
@@ -7,6 +7,43 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
+var notNamespaceableKinds = []string{
```
prachirp, line 10: Can you export the list and isNamespaceableKind function from https://github.com/kubernetes-sigs/kustomize/blob/master/api/resid/gvk.go and use it here?

prachirp:
cc @frankfarzan

functions/go/set-namespace/main.go
haiyanmeng: `notNamespaceableKinds` from https://github.com/kubernetes-sigs/kustomize/blob/master/api/resid/gvk.go is internal, and can only be used within the same package.
https://github.com/kubernetes-sigs/kustomize/blob/master/api/resid/gvk.go#L195 is a method on a structure, which is more than what we need here.

prachirp: The main motivation is to keep one source of truth list for `notNamespaceableKinds`. If we move the `isNamespaceableKind` implementation to some new file like kustomize/api/resid/util.go then it can use kustomize's `notNamespaceableKinds` and we don't have to maintain duplicate lists.

haiyanmeng: Make sense. Created https://github.com/kubernetes-sigs/kustomize/pull/2857 for this.

haiyanmeng: https://github.com/kubernetes-sigs/kustomize/pull/2859 is merged.
I updated this PR. PTAL.

haiyanmeng:
@prachirp , can this PR be merged? Thanks!

PR #95, opened 2020-08-25
Add --allow-cluster-scoped flag whose default value is true

haiyanmeng:
When the flag is set to false, this function should fail and return a list of cluster-scoped resources.

### DIFF
--- a/functions/go/set-namespace/main.go
+++ b/functions/go/set-namespace/main.go
@@ -1,19 +1,25 @@
 package main
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	"sigs.k8s.io/kustomize/kyaml/fn/framework"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
 var namespace string
+var allowClusterScoped bool
+
+var warning = "due to the difficulty of tracking all the cluster-scoped resource kinds, especially cluster-scoped CRDs."
 
 func main() {
 	resourceList := &framework.ResourceList{}
 	cmd := framework.Command(resourceList, func() error {
 		// cmd.Execute() will parse the ResourceList.functionConfig into cmd.Flags from
 		// the ResourceList.functionConfig.data field.
+		clusterScopedResources := []string{}
 		for i := range resourceList.Items {
 			// modify the resources using the kyaml/yaml library:
 			// https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml/yaml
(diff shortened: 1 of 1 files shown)


functions/go/set-namespace/main.go
```diff
@@ -33,11 +38,31 @@ func main() {
...
+					if err != nil {
+						continue
+					}
+					name := yaml.GetValue(nameNode)
+					resID := fmt.Sprintf("apiVersion: %s, kind: %s, name: %s", apiVersion, kind, name)
+					clusterScopedResources = append(clusterScopedResources, resID)
+				}
 			}
 		}
+		if len(clusterScopedResources) > 0 {
```
frankfarzan, line 58: what's the plan for cluster-scoped custom resources not hardcoded in implementation of isNamespaceable()?

functions/go/set-namespace/main.go
```diff
@@ -33,11 +38,31 @@ func main() {
...
+			return fmt.Errorf("the app config should only include namespace-scoped resources. "+
+				"But the following cluster-scoped resources are found:\n%s\n", strings.Join(clusterScopedResources, "\n"))
+		}
 		return nil
 	})
-	cmd.Flags().StringVar(&namespace, "namespace", "", "the namespace value")
+	cmd.Flags().StringVar(&namespace, "namespace", "", "the namespace to be added into namespaced resources")
```
frankfarzan, line 64: move flags to the top

functions/go/set-namespace/main.go
haiyanmeng: For these cluster-scoped custom resources, ACM will fail to apply the resources for the following error:
```
KNV1052: cluster-scoped resources MUST NOT declare metadata.namespace
```

functions/go/set-namespace/main.go
haiyanmeng: These flags can only be added after `cmd` is defined.

functions/go/set-namespace/main.go
haiyanmeng: After https://github.com/kubernetes-sigs/kustomize/pull/2620 is merged, we can also use the annotation to avoid adding non-application k8s resources (e.g., functions, constraints) into the hydrated config.

haiyanmeng: If ACM somehow does not cover all the cluster-scoped resource kinds, and the attempts to apply the cluster-scoped resources into a given namespace most likely will fail due to lack of permission.

frankfarzan: This check here is best-effort. There will be many cluster-scoped CRDs that are not hardcoded in kyaml, and won't be detected. We should be clear about that and document the UX to the user.

haiyanmeng: I clarified this in the command flags.

PR #96, opened 2020-08-25
Add tag to declarative example e2e tests

prachirp:


tests/helm_template.sh
```diff
@@ -36,7 +36,9 @@ assert_contains_string out.yaml "expected-args"
 
 testcase "kpt_helm_template_declarative_example"
 # TODO: Remove error handling once kpt pkg get shows errors gracefully https://github.com/GoogleContainerTools/kpt/issues/838
-kpt pkg get "$CATALOG_REPO"/examples/helm-template . || true
+kpt pkg get https://github.com/prachirp/kpt-functions-catalog.git/examples/helm-template@declarative-example-tags . || true
```
frankfarzan, line 39: This is pretty hacky. This is an example of transformation. Why not use use an imperative kpt function to change the tag?

Ultimately, perfect example of "search & replace" design doc.

frankfarzan: Also some comments about this file:

Why does NODOCKER option exist?

frankfarzan: Also how come the CI introducing the change passed?

https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/94

frankfarzan: As another solution, can you build :latest image (instead of :dev) and not push it before running the tests?

prachirp: Re: Using imperative kpt function, I tried replacing this with `kpt fn run istioctl-analyze/fn-config.yaml --image gcr.io/kpt-functions/annotate-config -- annotation_name=config.k8s.io/function annotation_value="container:\n  image: gcr.io/kpt-functions/istioctl-analyze:dev\n"` but couldn't get it working.

Re: NODOCKER option, we still need it for kpt tests otherwise the tests fail on macOS. See https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/77 specifically how CI failed on macOS tests before the last commit made it pass. We can 1.) rename NODOCKER for short-term, 2.) revisit how to set up tests in macOS for long-term

Re: CI introducing change, it passed because e2e test was testing against `GoogleContainerTools/kpt-functions-catalog` rather than `Shell32-Natsu/kpt-functions-catalog@helm-version`. Also it's a quirk of the function implementation that "data: chart: /source" becomes the same helm template arguments as
```
data:
  name: chart
  chart_path: /source
```
which is why this fix introduces the `name: chart` line.

prachirp: Re: another solution, it's better to use whatever image is denoted by $TAG since that's which tag is meant to be tested. Even if I built latest image and set $TAG as latest, we'd still want to  replace the image used in the declarative examples function config to use $TAG as it could be anything.

frankfarzan: Re: Imperative function. Why didn't it work? Did you investigate? We don't want to sweep issues under the rug. Having 3 lines of code to transform each package is not an acceptable solution.

Re: NODOCKER option: What's being tested on macOS? It seems like only Starlark? We should just remove MacOS as it's not adding value. Please follow up with this.

Re: CI passing for #94. Not sure I parsed this.

frankfarzan: Also another item to fix:

4. All tests should use the kpt CLI. There are still directly `docker run` test cases.

prachirp: 1.) Yes, I have to add double quotes to the annotation value arg in order to get $TAG to resolve correctly but this means that there are quotes around the value of the `config.k8s.io/function` annotation. So kpt does not parse the result as a function config anymore. It's also simpler and more understandable for our users to do it the current way.

2.) Sure created an issue and will follow up with a change after the blockers are fixed https://github.com/GoogleContainerTools/kpt/issues/976

3.) #94 e2e test tested the blueprint from `"$CATALOG_REPO"/examples/helm-template` so it did not actually test the blueprint change which was broken. #97 fixes the blueprint and also changes the e2e test to test the fixed blueprint at `https://github.com/prachirp/kpt-functions-catalog.git/examples/helm-template@helm-fix`.

4.) Yes the remaining `docker run` cases are blocked on https://github.com/GoogleContainerTools/kpt/issues/824. I added it as a blocker for 2.) above.

frankfarzan:
Is this PR still current?

prachirp:
@frankfarzan It's not current as I need to include new examples. We still need this change to ensure we can test the newest tag when running the examples. I will update this PR later this week.

frankfarzan:
@prachirp Please either complete this or delete the PR.

PR #97, opened 2020-08-26
Name chart in helm blueprint

prachirp:


tests/helm_template.sh
```diff
@@ -36,7 +36,7 @@ assert_contains_string out.yaml "expected-args"
 
 testcase "kpt_helm_template_declarative_example"
 # TODO: Remove error handling once kpt pkg get shows errors gracefully https://github.com/GoogleContainerTools/kpt/issues/838
-kpt pkg get "$CATALOG_REPO"/examples/helm-template . || true
+kpt pkg get https://github.com/prachirp/kpt-functions-catalog.git/examples/helm-template@helm-fix . || true
```
frankfarzan, line 39: Add a TODO that this is a temporary fix and follow up with real fixes

prachirp: Done

PR #98, opened 2020-08-27
Update helm template usage

prachirp:
Clarify that chart name is optional and will have a default value of RELEASE-NAME if none is provided.

### DIFF
--- a/functions/ts/src/helm_template.ts
+++ b/functions/ts/src/helm_template.ts
@@ -20,7 +20,7 @@ import { spawnSync } from 'child_process';
 
 const CHART_NAME = 'name';
 const CHART_PATH = 'chart_path';
-const VALUES_PATH = '--values';
+const VALUES = '--values';
 
 // Render chart templates locally using helm template.
 export async function helmTemplate(configs: Configs) {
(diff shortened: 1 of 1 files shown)


functions/ts/src/helm_template.ts
```diff
@@ -83,11 +83,11 @@ Render chart templates locally using helm template. If input a list of configs i
 addition to arguments will overwrite any chart objects that already exist in the list.
 
 Configured using a ConfigMap with keys for ${CHART_NAME}, ${CHART_PATH}.
-Works with arbitrary helm template flags like --values:
+Works with arbitrary helm template flags like ${VALUES}:
 
-${CHART_NAME}: Name of helm chart.
+${CHART_NAME}: Name of helm chart. Default RELEASE-NAME.
```
frankfarzan, line 88: What's the significance of `RELEASE-NAME`, where is it defined/consumed? Maybe link to helm docs.

prachirp: This is not documented anywhere in helm docs.

[This doc](https://helm.sh/docs/topics/v2_v3_migration/#overview-of-helm-3-changes) says helm "requires release name or --generate-name argument" but this is not accurate as helm uses "RELEASE-NAME" as a default if the name is not provided.

frankfarzan: Let me rephrase. If someone just provides CHAR_PATH, what's the expected behavior? In which situation the user needs to explicitly provide the CHART_NAME?

prachirp: If someone just provides chart-path then helm template will output expanded configs that use the name RELEASE-NAME. The user should provide a name or use the --generate-name flag if they do not want to use the default chart name.

functions/ts/src/helm_template.ts
```diff
@@ -83,11 +83,11 @@ Render chart templates locally using helm template. If input a list of configs i
...
 ${CHART_PATH}: Chart templates directory.
-${VALUES_PATH}: [Optional] Path to values file.
+${VALUES}: [Optional] Path to values file.
```
frankfarzan, line 90: nit: Consistent use of `[optional]`

PR #100, opened 2020-09-01
Release ts functions 0.1.0

prachirp:


functions/ts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ts-functions",
-  "version": "0.0.4",
```
frankfarzan, line 3: I think you want 0.5.0

prachirp: Done

PR #101, opened 2020-09-10
Sops kpt-function

aodinokov:
Would you be interested in the [SOPS](https://github.com/mozilla/sops/blob/master/README.rst) function that allows to decrypt yaml documents?

It allows to keep all yaml documents in encrypted form in a git repo and decrypt them right before sending it to k8s or performing some other actions. Please refer to the example to get more details on how to use it.
This function can work as kpt-function and as kustomize fn-plugin. I've tested it [here](https://github.com/aodinokov/noctl-airship-poc/blob/master/packages/test_fn_sops/run.sh) in both modes.

I've created an example in my PR as well.

Please let me know if you're interested or not.
Couple of additional comments for a case if you're interested:

- this function doesn't decrypt comments, because comments are not included in ResourceList. I think it's not a big issue
- it's necessary to avoid using 'sops': fields in the root of document, because it's the way how this function understands that the document is encrypted
- when use with kustomize build, it's necessary to keep in mind that it's not possible to encrypt GVK+name fields. for that reason this documents must be encrypted only partially. need to use any of the following options: `--encrypted-regex`, `--unencrypted-suffix`, `--encrypted-suffix`.
- since kustomize build changes the field order it's necessary to add `ignore-mac: true` as and additional option. See [here](https://github.com/aodinokov/noctl-airship-poc/blob/master/packages/test_fn_sops/manifests/base/sops.yaml).

PS
It may be that it should be named as `sops-decrypt` instead of `sops`. I just think if there is also a need in some additional 
functionality, like encryption. But it may be in future - so far I didn't plan to add this functionality. What do you think about the name?

I used rw module to write to the temp file. This required an additional module, that I added and because of that I ended up with adding to all dockerfiles the construction listed below. I'm not sure it's a good thing, but `package.json` required that. Maybe we can use a separate `package.json` and etc for some of functions? where should we put it?

```
# Make rw package work
COPY --chown=node:node @types @types
```

and the last: To run the example, please substitute the container url with quay.io/aodinokov/sops:v0.0.2, since gcr image isn't built yet 

Cheers,
Alexey


functions/ts/build/helm_template.Dockerfile
```diff
@@ -19,6 +19,8 @@ WORKDIR /home/node/app
 
 # Install dependencies and cache them.
 COPY --chown=node:node package*.json ./
+# Make rw package work
```
frankfarzan, line 22: These work fine right now. Why is this change necessary?

examples/sops/local-configs/example-use.yaml
```diff
@@ -0,0 +1,80 @@
+apiVersion: v1
```
frankfarzan, line 1: Separate out this multi-object file into multiple files for clarity:

function.yaml
encrypted.yaml

examples/sops/local-configs/example-use.yaml
```diff
@@ -0,0 +1,80 @@
...
+metadata:
+  name: my-config
+  annotations:
+    config.k8s.io/function: |
+      container:
+        image: gcr.io/kpt-functions/sops
+    config.kubernetes.io/local-config: "true"
+data:
+---
+myapp1: ENC[AES256_GCM,data:zlGNmhTYX5xol4ZZFsiaoGkD73nn,iv:ql9mkhoU1I64E/FJi3iA0HaAe2U3kQVFee2ZLwPnBik=,tag:SqVSfu/JkRrwqidAT/i0pg==,type:str]
```
frankfarzan, line 12: Any reason this is not a K8S resources, i.e. doesn't have GVKN?

examples/sops/sops_functional_tests_key.asc
```diff
@@ -0,0 +1,174 @@
+-----BEGIN PGP PRIVATE KEY BLOCK-----
```
frankfarzan, line 1: Don't check in the private key in the package as it sets a bad precedence for readers. Maybe add it in a code block in the README with note explaining it's only given for demo purposes.

functions/ts/src/sops.ts
```diff
@@ -0,0 +1,206 @@
...
+  return args;
+}
+
+sops.usage = `
+Sops function (see https://github.com/mozilla/sops).
+So far supports only decrypt operation:
+runs sops -d for all documents that have field 'sops:' and put the decrypted result back.
+
+Can be configured using a ConfigMap with the following flags:
+ignore-mac: true [Optional: default empty]
```
frankfarzan, line 185: explain what these parameter do.

functions/ts/src/sops.ts
```diff
@@ -0,0 +1,206 @@
+import {
```
frankfarzan, line 1: can you add some e2e tests for this to ensure it doesn't bit rot?

functions/ts/src/sops.ts
```diff
@@ -0,0 +1,206 @@
...
+}
+
+function detachConfigKubernetesIoAnnotations(
+  object: SopsKubernetesObject
+): Map<string, string> {
+  const annotations: string[] = [
+    'config.kubernetes.io/index',
+    'config.kubernetes.io/path',
+    'config.k8s.io/id',
+    'kustomize.config.k8s.io/id',
```
frankfarzan, line 90: Why not just remove any annotations with prefix `config.kubernetes.io`?


@monopole What's `kustomize.config.k8s.io/id` and how's it different from `config.k8s.io/id`?
Same with `config.kubernetes.io` vs `config.k8s.io`. Why do these different prefixes co-exist?

examples/sops/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Verify the updated configuration:
+
+```sh
+kpt cfg cat sops/local-configs
+```
+
+## Note:
+
+Starting kustomize 3.8.1 this function can be also used by kustomize as a plugin, but
```
frankfarzan, line 44: I'd avoid this Note given this user guide is not concerned about kustomize binary.

frankfarzan (commented):
Thanks for contributing this. Overall approach is reasonable to me, have some comments.

functions/ts/src/sops.ts
aodinokov: first 3 used by kustomize/kpt fn run. 
the last one used by kustomize build to check keep information about the document.

>Why not just remove any annotations with prefix config.kubernetes.io?


hmm. really good catch. that makes total sense if we're confident that all annotations with that prefixes are added by kustomize/kpt and were not part of the original documents. waiting for **monopole** to chime it

examples/sops/local-configs/example-use.yaml
aodinokov: just to demonstrate that it's possible to decrypt any documents.
I can add both GVKN and non-GVKN,
or change it to GVKN-only.
Can do it along with splitting the multi-doc

functions/ts/build/helm_template.Dockerfile
aodinokov: I needed rw.writeFile to put yamls into /tmp/ directory (sops works with files).
I looked up this function in [kpt-functions-sdk](https://github.com/GoogleContainerTools/kpt-functions-sdk).
It didn't work until I modified [package.json](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/101/files#diff-6e89b8d725b5789ac81ff297fac241c1) and [package-lock.json](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/101/files#diff-0613cd7774d07870a90667aa23032291) by adding `file:@types/rw`.  But since these files are shared across all ts-functions, it's necessary to have `@types/rw` in all dockerfiles, because it's referenced. I wasn't able to find a better solution. Maybe separate package.json for each ts-function? but I'm not sure where to put it.

functions/ts/src/sops.ts
aodinokov: sure - I can :) . I wasn't sure if you were interested in this function. Now when I see the interest, I will make e2e based on the already existing tests for other functions, along with resolving other comments.

examples/sops/README.md
aodinokov: will remove it. Thank you for the review!

functions/ts/src/sops.ts
aodinokov: @frankfarzan , I've checked -
There are at least 3 more annotations with the same prefixes that we don't want to remove accidentally:

- config.kubernetes.io/local-config
- config.kubernetes.io/function
- config.k8s.io/function

I think for now it's better to keep this implementation. If we're not sure what will happen with this annotations I can add a parameter to override this list in the function configuration.

aodinokov: done

functions/ts/src/sops.ts
aodinokov: added tests/sops.sh

functions/ts/package.json
```diff
@@ -30,6 +30,7 @@
     "@types/jasmine": "^3.5.11",
     "@types/js-yaml": "^3.12.1",
     "@types/node": "^14.0.23",
+    "@types/rw": "file:@types/rw",
```
frankfarzan, line 33: rw should be part of `dependencies` right?

  "dependencies": {
    "rw": "^1.3.3"
  },

tests/sops.sh
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env bash
```
frankfarzan, line 1: @prachirp FYI, this needs to be cleaned up https://github.com/GoogleContainerTools/kpt/issues/976

functions/ts/build/helm_template.Dockerfile
frankfarzan: Filed https://github.com/GoogleContainerTools/kpt/issues/1026

To unblock you, you can add TODO in Dockerfiles linking to that issue.

functions/ts/package.json
aodinokov: correct. without that it wasn't possible to use this module.

functions/ts/build/helm_template.Dockerfile
aodinokov: done

tests/sops.sh
prachirp: Cleaning up depends on features in the next kpt release which I expect to happen sometime this week.

examples/sops/local-configs/example-use.yaml
frankfarzan: The spec requires that items in ResourceList be K8S resources:
https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md

Currently, non-K8S resources not leading to validation failures is an implementation detail that shouldn't be depended on. So, please use GVKN only

functions/ts/src/sops.ts
frankfarzan: Filed an issue: https://github.com/kubernetes-sigs/kustomize/issues/2996

frankfarzan: Let's add a TODO to this issue.

examples/sops/local-configs/example-use.yaml
aodinokov: agree, even though it's working now, it's better to follow the specs so it would continue work in future :). I'll remove non-GVKN item.

functions/ts/src/sops.ts
monopole: I don't think there's a canonical list of well-known annotation keys or `apiVersion` values, or policies around using the latter as the former, that one can reliably use for any given goal.  So providing a means to specify such a "detach or ignore" list, and offering a good guess as a default, looks good to me.

e.g. this is interesting, but incomplete :  https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints

examples/sops/local-configs/example-use.yaml
aodinokov: done

functions/ts/src/sops.ts
aodinokov: I was thinking about the same: created the list of annotations that should be detached before the decryption, but I also added a parameter called [override-detached-annotations](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/101/files#diff-11b1a6e5f9cc9e2bb106e55858164b37R199) that allows to specify comma-separated list of annotations to be detached if the behavior should be changed for any reason.

functions/ts/package.json
frankfarzan: Clarify: `dependencies` section above currently doesn't declare `rw`. How's this working?

aodinokov: sorry, just misunderstood the question, thank you for clarification 👍 
You're right - I missed that.
I've corrected and added:
"dependencies": {
...
"rw": "^1.3.3"
...
},

I think that it is working without that, because it has  `"kpt-functions": "^0.14.5"` in `dependencies` and
`kpt-functions` depends on rw: see [here](https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/ts/kpt-functions/package.json#L34)

frankfarzan (approved):
Thanks for contributing this function.

functions/ts/src/sops.ts
```diff
@@ -0,0 +1,223 @@
...
+To decrypt the documents use:
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-config
+  annotations:
+    config.k8s.io/function: |
+      container:
+        image: (put here the image url)
```
prachirp, line 219: I'll add this into a PR and make a release of the package.

functions/ts/src/sops.ts
```diff
@@ -0,0 +1,223 @@
...
+  }
+  configMap.forEach((value: string, key: string) => {
+    if (key === OVERRIDE_PREEXEC_CMD) {
+      preExecCmd = value;
+    } else if (key === OVERRIDE_DETACHED_ANNOTATIONS) {
+      detachedAnnotations = value.replace(/\s/g, '').split(',');
+    } else if (key === VERBOSE || key === IGNORE_MAC) {
+      args.push('--' + key);
+    } else {
+      args.push('--' + key);
```
prachirp, line 181: Nice way to handle arbitrary flags! We've had issues with this in the past (https://github.com/GoogleContainerTools/kpt/issues/824)

prachirp (commented):
Thanks for contributing this! I'll release it this week.

PR #103, opened 2020-09-20
Adding a simple function to test env vars

aodinokov:
This function was created based the request from kustomize team:
this is for kubernetes-sigs/kustomize#2999

Please pay attention that go functions don't get built, so to make my e2e work I've done this:
https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/103/commits/84f2c5702d021f925437633b9db442486dcaf86c#diff-60fe268f71ae7f8409aff40d0a3e51d8


aodinokov:
this is for https://github.com/kubernetes-sigs/kustomize/pull/2999
cc: @monopole , @Shell32-Natsu

.github/workflows/ci.yaml
```diff
@@ -67,6 +67,8 @@ jobs:
       run: |
         cd functions/ts
         npm run kpt:docker-build
+        cd ../go
```
aodinokov, line 70: a little hack to make e2e work.
probably should be done in more generic way

prachirp (changes requested):
Thanks, this makes sense in the context of testing env variables.

functions/go/templater/README.md
```diff
@@ -0,0 +1,46 @@
+# Templater function
+
+This is an example of implementing a templater function.
```
frankfarzan, line 3: what's a "templater". This paragraph doesn't actually explain what this does.

functions/go/templater/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This example is written in `go` and uses the `kyaml` libraries for parsing the
```
frankfarzan, line 5: This is an implementation detail that should not be exposed to the user. Instead explain the use case, why would anyone want to use this function? Why is it useful?

frankfarzan (changes requested):

functions/go/templater/function/filter_test.go
```diff
@@ -0,0 +1,206 @@
...
+		{
+			cfg: `
+apiVersion: v1alpha1
+kind: Templater
+metadata:
+  name: notImportantHere
+data:
+  template: |
+    {{ range .hosts -}}
+    ---
```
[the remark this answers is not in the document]
aodinokov: this `---` is needed because the intent was to generate a yaml document for each objects in `.hosts`. 
kio.Pipeline ignores empty documents, when reads and converts buffer to RNode [here](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/103/files/84f2c5702d021f925437633b9db442486dcaf86c#diff-a281396d074fd7355fe0e872747be5e1R63).
that's why we see only 2 generated documents.
I could put `---` near `{{ end -}}`, the result should be the same.

.github/workflows/ci.yaml
aodinokov: done

functions/go/templater/function/filter_test.go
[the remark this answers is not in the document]
prachirp: I see thanks!

prachirp (approved):
LGTM

functions/go/templater/README.md
aodinokov: I've updated the text below

functions/go/templater/README.md
```diff
@@ -0,0 +1,60 @@
+# Templater function
+
+This is an example of implementing a templater function.
+Templater allows generating documents using go-template
+engine and Sprig library.
```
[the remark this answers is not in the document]
aodinokov: right naming is always a question :)
Sprig is a name of the library. Template syntax is from go-template. I'm not sure that sprig-template will resonate.
Looks like the function you found can only generate nginx service and deployment templates and used just as an example to demonstrate what functions can do. The image name for that function is:

```
gcr.io/kustomize-functions/example-nginx
```

(see [here](https://github.com/kubernetes-sigs/kustomize/blob/master/functions/examples/template-go-nginx/local-resource/example-use.yaml#L11)), even though it's located in `template-go-nginx` directory.
Templater function is more universal, it can simulate `example-nginx` behavior and generate the same yamls if the appropriate template is included.

I can rename if you insist, but I'm not sure we should. Just let me know if you still fill that you need that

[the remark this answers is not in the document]
prachirp: Thanks the templater name makes sense based on your explanation.

functions/go/templater/README.md
frankfarzan: Thanks for the clarification. Even though we CAN create a templating function like this, it goes against the Config as Data principles:

https://googlecontainertools.github.io/kpt/concepts/architecture/#configuration-as-data

Which is an important motivator for why Kpt/Kustomize projects exist.

frankfarzan: I don't think this is resolved. Nothing preventing users from using a templater function, but having it in this catalog is against the design philosophies of Config as Data

aodinokov: oh.. ok, It seems I missed the last comment and was just waiting... I'll look into that and will try to address it. Sorry for the delay.

aodinokov: @frankfarzan 
I've tried to modify the function, so it could employ go-template engine + sprig as a scripting lang to perform transformation.
To be able to work with yaml-resources, I've introduced [KYAML go-template library](https://github.com/aodinokov/kpt-functions-catalog/tree/function-with-env/functions/go/templater#kyaml-go-template-library). The new version of README contains some [examples](https://github.com/aodinokov/kpt-functions-catalog/tree/function-with-env/functions/go/templater#kyaml-go-template-library-usage-examples) how this function can be used.
The result is pretty universal and powerful. Hope you'll like this concept, since it's based on kyaml primitives. 
@monopole asked to create a simple function to be able to test that env variables pass. Even though this function isn't absolutely simple, it's still pretty short for the functionality it can provide. I think this function can be used for a big amount of different cases: filtering, annotating/lableling/modification of resources. It may be useful  especially for non-standard cases.

Let me know if it's too complicated or you have other concerns.
If it's not something you want to support I can make something really simple for the testing purposes of https://github.com/kubernetes-sigs/kustomize/pull/2999 . I just wanted to make something meaningful and useful, not just yet-another-example of set-label fn, that takes env var for that :)

aodinokov:
@frankfarzan sorry for the delay with the new version.
I've completely removed the template mechanism and have made a really simple function, but since it works with env variables it can be used when the value isn't knows. E.g. timestamping example.

while I was working on this I found several little issues in the repo:

1. https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/116 - forgot about functions/ts/build/kustomize_build.Dockerfile - didn't fix, since it doesn't fail
2. https://github.com/helm/charts/commit/2f54b7c00c9e72b210dbcfb6254a68746672a6a7#diff-4ea099515951a32bd22484cbcc15d627597b97b43f6df64437b8ca64a2ff6676 - chart is deprecated and that causes error - partially fixed, partially commented
3. https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/tests/annotate_config.sh#L53 - this branch has been removed and that causes error - fixed

PR #106, opened 2020-09-24
Added docker-cli package to kustomize_build fn

aodinokov:
This is to make docker-in-docker possible for kustomize_build fn:

kustomize build command may work with kpt-function
based plugins, but in order to make it work
kustomize_build function must contain docker-cli
to be able to run docker in docker.

### DIFF
--- a/functions/ts/build/kustomize_build.Dockerfile
+++ b/functions/ts/build/kustomize_build.Dockerfile
@@ -35,7 +35,7 @@ RUN npm run build && \
 
 FROM node:lts-alpine
 
-RUN apk add git
+RUN apk add git docker-cli
 
 # Run as non-root user as a best-practices:
 # https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md


aodinokov:
FYI @Shell32-Natsu

tests/kustomize_build.sh
```diff
@@ -79,3 +79,61 @@ kpt fn source examples |
   kpt fn sink .
 assert_contains_string configmap_the-map.yaml "app: hello"
 assert_dir_exists knative-serving
+
+testcase "kpt_kustomize_build_fn_plugins"
+mkdir fn
+mkdir -p source/base
+mkdir -p source/kustomize/plugin/
+curl -fsSL -o key.asc https://raw.githubusercontent.com/mozilla/sops/master/pgp/sops_functional_tests_key.asc || echo "couldn't create key.asc"
```
frankfarzan, line 87: what does 'kustomize build' have to do with sops? There's a sops function with its own e2e tests.

frankfarzan (changes requested):

tests/kustomize_build.sh
```diff
@@ -79,3 +79,61 @@ kpt fn source examples |
   kpt fn sink .
 assert_contains_string configmap_the-map.yaml "app: hello"
 assert_dir_exists knative-serving
+
+testcase "kpt_kustomize_build_fn_plugins"
+mkdir fn
+mkdir -p source/base
+mkdir -p source/kustomize/plugin/
```
prachirp, line 86: Where do you use the plugin dir? Is it implicit when setting KUSTOMIZE_PLUGIN_HOME?

tests/kustomize_build.sh
```diff
@@ -79,3 +79,61 @@ kpt fn source examples |
...
+EOF
+
+cat >source/base/kustomization.yaml <<EOF
+resources:
+- encrypted-gvk.yaml
+transformers:
+- sops.yaml
+EOF
+
+DOLLAR=$
```
prachirp, line 119: Nice way to get this resolving inside quotes!

tests/kustomize_build.sh
```diff
@@ -79,3 +79,61 @@ kpt fn source examples |
   kpt fn sink .
 assert_contains_string configmap_the-map.yaml "app: hello"
 assert_dir_exists knative-serving
+
+testcase "kpt_kustomize_build_fn_plugins"
+mkdir fn
+mkdir -p source/base
+mkdir -p source/kustomize/plugin/
+curl -fsSL -o key.asc https://raw.githubusercontent.com/mozilla/sops/master/pgp/sops_functional_tests_key.asc || echo "couldn't create key.asc"
+curl -fsSL -o source/base/encrypted-gvk.yaml \
```
prachirp, line 88: We generally use `kpt pkg get "$CATALOG_REPO"/examples/sops/...` for this.

prachirp (approved):
LGTM

tests/kustomize_build.sh
aodinokov: 'kustomize build' may use kpt-functions as plugins.
Before adding `docker-cli` to the dockerfile it wasn't possible.

The intent of the test was just to make sure that this worked now.
I could use any of existing container-based functions (ts or go) to write a case that covers running kpt-fn-based plugins inside kustomize_build function. Since I anyway needed to select one function - it was easier for me to make this test with sops, because I knew it would work.

I had another option: to use kustomize_build function as kustomize plugin inside another kustomize_build function that is ran by kpt was an another example, but I just though it was too artificial. 

I can try to rewrite the test with another function. Or may completely remove the test, since the change is trivial.
Or I can move this test to sops e2e test file.

What option seems reasonable to you?

tests/kustomize_build.sh
aodinokov: yes, correct: [here](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/106/files#diff-cddb7e3d19c09d9f69d411848ad17f9cR103) That's because kpt-function based plugins can only work when `--enable_alpha_plugins` is set in kustomize options. And if this flag is set it's necessary to have a directory with plugins, even though it is empty.

tests/kustomize_build.sh
frankfarzan: My vote is to completely remove the test.

Larger context: We are in the process of conceptually aligning kpt and kustomize projects. Having 'kustomize build' as a KRM function doesn't really make sense in an e2e flow and will be removed in the near future. 

Filed https://github.com/GoogleContainerTools/kpt/issues/1078

Sorry for the confusion caused here.

aodinokov: Thank you for the larger context! That sounds really interesting - looking forward to see what it will be.
Meanwhile - removing the test 👍

frankfarzan (approved):

PR #107, opened 2020-09-24
Remove broken function catalog links.

etefera:
Addresses part of https://github.com/GoogleContainerTools/kpt/issues/1066


frankfarzan:
we need a top-level index page for all the category pages. Currently it's at the end of this page: https://googlecontainertools.github.io/kpt/guides/consumer/function/
So every time I want to find a function, I have to scroll past the user guide.

frankfarzan (approved):

PR #111, opened 2020-10-16
Support remote chart in helm-template

Shell32-Natsu:
https://github.com/GoogleContainerTools/kpt/issues/945

- Support remote chart in helm-template
- e2e test should work with kpt >= 0.35.0
- Use [`nyc`](https://github.com/istanbuljs/nyc) (istanbul) to provide a unit test coverage report for TS functions.


functions/ts/package.json
```diff
@@ -13,7 +13,7 @@
     "lint-license": "license-checker --onlyAllow 'Apache-2.0;MIT;BSD;BSD-2-Clause;BSD-3-Clause;ISC;CC-BY-3.0;CC0-1.0;Unlicense'",
     "format": "prettier \"src/**\" \"*.json\" --write",
     "pretest": "npm run build",
-    "test": "jasmine --config=jasmine.json",
+    "test": "nyc -x 'src/gen/**' -x '**/*_test.ts' jasmine --config=jasmine.json",
```
frankfarzan, line 16: Let's keep the test framework consistent across functions in the catalog and the sdk repo. Any reason you're changing this?

functions/ts/package.json
```diff
@@ -30,12 +30,15 @@
   "devDependencies": {
     "@types/jasmine": "^3.5.14",
     "@types/js-yaml": "^3.12.5",
-    "@types/node": "^14.11.2",
+    "@types/node": "^14.11.10",
```
frankfarzan, line 33: If you haven't already take a look at the guideline for adding dependencies (even if they're devDependencies):

https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/CONTRIBUTING.md#adding-an-npm-dependency

Even though it's very easy to add dependencies with tools like NPM, each new dependency needs to be scrutinized.

functions/ts/src/helm_template.ts
```diff
@@ -105,6 +237,24 @@ metadata:
     config.kubernetes.io/local-config: "true"
 data:
   ${CHART_NAME}: my-chart
-  ${CHART_PATH}: ../path/to/helm/chart
-  ${VALUES}: ./values.yaml
+  ${LOCAL_CHART_PATH}: ../path/to/helm/chart
+  ${VALUES_PATH}: ./values.yaml
+
+2. To expand a chart named 'my-chart' at remote chart 'stable/chart'
```
frankfarzan, line 243: I would put this example first. Based on offline discussions, #1 is not a best practice for using functions.

functions/ts/src/helm_template.ts
```diff
@@ -17,82 +17,214 @@
...
+
+  getData() {
+    const data = this.configs.getFunctionConfigMap();
+    if (data !== undefined) {
+      this.data = data;
+    }
+    this.validData();
+  }
+
+  validData() {
```
frankfarzan, line 61: this doesn't need to be a method. In general prefer using first class functions instead of classes and methods. This is a major difference between more functional languages like TS vs something like Java.

This should be:

validateXXX(param)

I'd also avoid use of "data" as it's too generic to be meaningful

functions/ts/src/helm_template.ts
```diff
@@ -17,82 +17,214 @@
...
 }
 
-function readArguments(configs: Configs) {
-  const args: string[] = [];
-  let nameArg: string | undefined;
-  let pathArg: string | undefined;
-  const configMap = configs.getFunctionConfigMap();
-  if (!configMap) {
-    return args;
+class HelmTemplate {
```
frankfarzan, line 43: You don't really need to use a class here, see my comment below about methods vs functions

functions/ts/src/helm_template.ts
```diff
@@ -17,82 +17,214 @@
...
+    }
+    if (this.data.get(CHART_NAME) !== undefined) {
+      args.unshift(this.data.get(CHART_NAME)!);
+    }
+
+    return args;
+  }
+
+  // run helm template command
+  async template() {
```
frankfarzan, line 109: Functions/method names should be a verb. e.g.:

runHelmTemplate

frankfarzan (changes requested):

functions/ts/package.json
Shell32-Natsu: I want to know the coverage of my unit test codes. Our current unit tests are really simple and I even have no idea how simple they are.

functions/ts/src/helm_template.ts
Shell32-Natsu: I use a class because it makes the variable management easier here.

functions/ts/src/helm_template.ts
Shell32-Natsu: Will do

functions/ts/src/helm_template.ts
Shell32-Natsu: Will do

functions/ts/package.json
Shell32-Natsu: [`nyc`](https://www.npmjs.com/package/nyc) is the CLI interface for unit test coverage tool [`istanbuljs`](https://istanbul.js.org/). `nyc` is a big project and I cannot tell that all dependencies are high quality but it is one of the best tool that I can found for showing coverage report. It's a builtin component in [`jest`](https://github.com/facebook/jest) as well.
[`proxyquire`](https://www.npmjs.com/package/proxyquire) is used to do mock in the unit test. It's very simple and has only seldom dependencies.
Their licenses are OK.

I need to add these dependencies since `jasmine` doesn't have enough capabilities (coverage report and function/class mock).

functions/ts/package.json
Shell32-Natsu: By this tool I can see a report of the unit test coverage. Like this
```
---------------------|---------|----------|---------|---------|--------------------------
File                 | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s        
---------------------|---------|----------|---------|---------|--------------------------
All files            |   67.98 |    56.98 |    62.5 |    68.5 |                          
 helm_template.ts    |   98.68 |    93.33 |     100 |   98.67 | 121                      
 istioctl_analyze.ts |   55.26 |    33.33 |      60 |   55.26 | 46-89,103,111-112        
 kubeval.ts          |    54.1 |    54.17 |      20 |      55 | 50,61-93,115-116,140-152 
 kustomize_build.ts  |   32.14 |        0 |      50 |   33.33 | 26-39,51-62              
---------------------|---------|----------|---------|---------|--------------------------
```

functions/ts/src/helm_template.ts
```diff
@@ -17,82 +17,226 @@
...
+const CHART_REPO = 'chart-repo';
+// CHART_REPO_URL is the repo remote URL
+const CHART_REPO_URL = 'chart-repo-url';
 
-// Render chart templates locally using helm template.
-export async function helmTemplate(configs: Configs) {
-  // Validate config data and read arguments.
-  const args = readArguments(configs);
-  args.unshift('template');
+class HelmTemplateError extends Error {
```
prachirp, line 34: Let's name this HelmError, keep it more generic.

functions/ts/src/helm_template.ts
```diff
@@ -17,82 +17,226 @@
...
+Configured using a ConfigMap with keys for ${CHART_NAME}, ${LOCAL_CHART_PATH}, ${CHART}, ${CHART_REPO} or ${CHART_REPO_URL}.
+Works with arbitrary helm template flags like --values:
 
-${CHART_NAME}: [Optional] Name of helm chart. Default RELEASE-NAME.
-${CHART_PATH}: Chart templates directory.
-${VALUES}: [Optional] Path to values file.
+${CHART_NAME}: Name of helm chart.
+${LOCAL_CHART_PATH}: Chart templates directory.
+${VALUES_PATH}: [Optional] Path to values file.
+${CHART}: A url to pull the remote chart instead of using local templates. This flag only works with remote charts.
```
prachirp, line 214: Note optional keys.

functions/ts/src/helm_template.ts
```diff
@@ -17,82 +17,226 @@
...
+${CHART_REPO}: Repo name that helm should pull the templates from. Only used when chart is provided.
+${CHART_REPO_URL}: Repo list URL which will be added to the helm repo list with repo name chart-repo. Only used when chart is provided.
 ...
 
-Example:
+Examples:
+
+1. To expand a chart named 'my-chart' at remote chart 'stable/chart'
```
prachirp, line 221: Need to specify CHART_NAME: my-chart

prachirp (changes requested):
Nice, the unit test coverage will help with testing other functions too.

functions/ts/package.json
frankfarzan: Couple of things:

1. I understand, but we want to have a consistent dev environment across our NPM packages. So we should either use jasmine everywhere or nyc everywhere.

2. Also keep in mind my other comment about evaluating new dependencies to avoid pulling in a transitive deps. For maintainability and security reasons, you want as few transitive deps as possible.

For comparison:

http://npm.broofa.com/?q=jasmine

vs

http://npm.broofa.com/?q=nyc

The latter pulls in A LOT of deps. That's the reason we decided to stick with jasmine.

functions/ts/package.json
frankfarzan: proxyrequire has fewer deps:

http://npm.broofa.com/?q=proxyquire

It should be fine.

frankfarzan (approved):

prachirp (approved):

PR #119, opened 2020-11-06
fix release e2e test

Shell32-Natsu:
The helm template example need to be updated to release latest images.


prachirp (approved):
We can update the kpt website to use remote chart now https://googlecontainertools.github.io/kpt/guides/ecosystem/helm/

frankfarzan (approved):

PR #121, opened 2020-11-11
fix ci issue

Shell32-Natsu:
We haven't release the helm-inflator image so we have to use helm-template function in examples.

### DIFF
--- a/tests/helm_inflator.sh
+++ b/tests/helm_inflator.sh
@@ -28,10 +28,11 @@ assert_contains_string out.yaml "expected-args"
 
 testcase "kpt_helm_inflator_declarative_example"
 # TODO: Remove error handling once kpt pkg get shows errors gracefully https://github.com/GoogleContainerTools/kpt/issues/838
-# TODO: update helm-template to helm-inflator
-kpt pkg get "$CATALOG_REPO"/examples/helm-template . || true
-kpt fn run helm-template/local-configs --mount type=bind,src="$(pwd)"/helm-template/helloworld-chart,dst=/source --as-current-user
-assert_contains_string helm-template/local-configs/deployment_chart-helloworld-chart.yaml "name: chart-helloworld-chart"
+kpt pkg get "$CATALOG_REPO"/examples/helm-inflator . || true
+# Use proper image tag
+sed -e "s/image: gcr.io\/kpt-functions\/helm-inflator/image: gcr.io\/kpt-functions\/helm-inflator:${TAG}/" -i helm-inflator/local-configs/fn-config.yaml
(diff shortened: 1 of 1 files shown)


examples/helm-inflator/local-configs/fn-config.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   annotations:
     config.kubernetes.io/function: |
       container:
-        image: gcr.io/kpt-functions/helm-inflator
+        image: gcr.io/kpt-functions/helm-template
```
prachirp, line 9: Can you manually push the helm-inflator image and avoid this change?

Shell32-Natsu: Do I have the permission to directly push it?

prachirp: @frankfarzan Can you give Donny permissions for this?

frankfarzan: There should be different flows:

1- CI workflow should be testing the unreleased images not the latest.
2- CD should be testing the release candidate images

Why do you need to manually release an image? It seems like an antipattern, images should be released through automated CD.

frankfarzan (changes requested):

tests/helm_inflator.sh
```diff
@@ -28,10 +28,11 @@ assert_contains_string out.yaml "expected-args"
 
 testcase "kpt_helm_inflator_declarative_example"
 # TODO: Remove error handling once kpt pkg get shows errors gracefully https://github.com/GoogleContainerTools/kpt/issues/838
-# TODO: update helm-template to helm-inflator
-kpt pkg get "$CATALOG_REPO"/examples/helm-template . || true
-kpt fn run helm-template/local-configs --mount type=bind,src="$(pwd)"/helm-template/helloworld-chart,dst=/source --as-current-user
-assert_contains_string helm-template/local-configs/deployment_chart-helloworld-chart.yaml "name: chart-helloworld-chart"
+kpt pkg get "$CATALOG_REPO"/examples/helm-inflator . || true
+# Use proper image tag
+sed -e "s/image: gcr.io\/kpt-functions\/helm-inflator/image: gcr.io\/kpt-functions\/helm-inflator:${TAG}/" -i helm-inflator/local-configs/fn-config.yaml
```
frankfarzan, line 33: This is pretty hacky, can you file an issue to address this generally for all functions instead of in an adhoc fashion?

frankfarzan (approved):

tests/helm_inflator.sh
Shell32-Natsu: https://github.com/GoogleContainerTools/kpt/issues/1224

PR #123, opened 2020-11-20
Update set-namespace function

Shell32-Natsu:
Change set-namespace function to kustomize based version


functions/go/set-namespace/Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM golang:1.14
+FROM golang:1.13-stretch
```
frankfarzan, line 1: What are the binary files (set-namespace-linux) being checked in?

functions/go/set-namespace-legacy/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.14
```
frankfarzan, line 1: Why keep the legacy function around?

functions/go/set-namespace-legacy/Makefile
```diff
@@ -0,0 +1,21 @@
+.PHONY: fix vet fmt test lint
```
frankfarzan, line 1: We should have a top-level Makefile instead of having each function have its own copy. How's the new function being built?

tests/set_namespace_go.sh
```diff
@@ -41,17 +41,18 @@ metadata:
...
-assert_contains_string example-configs/gatekeeper.yaml "namespace: example-ns"
+kpt fn source example-configs |
+  kpt fn run --fn-path fc.yaml >out.yaml
+assert_contains_string out.yaml "namespace.*:.*example-ns"
 
 # Use the linux executable to test using kpt exec runtime on linux
 testcase "kpt_set_namespace_go_exec_imperative_linux"
 kpt pkg get "$SDK_REPO"/example-configs example-configs
 kpt pkg get "$CATALOG_REPO"/functions/go ./
 kpt fn run example-configs --enable-exec --exec-path "$(pwd)"/go/set-namespace/set-namespace-linux -- namespace=example-ns
```
frankfarzan, line 52: - We should get rid of any non-containerized tests, also generally standardize the tests as part of support for curated functions. They're pretty adhoc and hack right now.
- Should test for idempotency and other properties as discusse don the doc.

@mengqiy

functions/go/set-namespace/main.go
```diff
@@ -1,75 +1,97 @@
...
+	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/fn/framework"
-	"sigs.k8s.io/kustomize/kyaml/yaml"
+	"sigs.k8s.io/yaml"
 )
 
-var namespace string
-var allowClusterScoped bool
+//nolint
+func main() {
```
frankfarzan, line 18: What's the --help string for this command?

The usage message should a) provide a description of this function b) Describe the function config required.

Example:

https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/ts/hello-world/src/label_namespace.ts#L29

frankfarzan (changes requested):

functions/go/set-namespace/Dockerfile
Shell32-Natsu: That follows the old set-namespace function. It has 2 binary files checked in, one for linux and one for mac.

functions/go/set-namespace-legacy/Dockerfile
Shell32-Natsu: @mengqiy

functions/go/set-namespace/Dockerfile
frankfarzan: There's a functions/go/set-namespace/set-namespace-linux which is being added in the new function, no? In general, we should build from source, and not check in any prebuilts in Git.

Shell32-Natsu: I guess those old binary files are checked in for exec test. Since we will avoid non-containerized tests so I will remove it.

tests/set_namespace_go.sh
Shell32-Natsu: The test files used in e2e test, (`"$SDK_REPO"/example-configs`) cannot keep idempotency. Because after applying the namespace function, there will be multiple resources with duplicate ID (group, kind, version, namespace and name tuple). I will use a subdirectory (`"$SDK_REPO"/example-configs/shipping-dev`) in the test.

functions/go/set-namespace/Dockerfile
mengqiy: IMO binary should not be checked in here.
And even if we want to check in some big files, GIt LFS should be used.

functions/go/set-namespace-legacy/Dockerfile
mengqiy: Initially I thought we were doing this in 2 steps:
1) we rename the old one as "legacy" and meanwhile setup the new one.
2) remove the legacy after the new one is ready (including code, test and example and README).

We also remove the old one in this PR if you want to do it.

functions/go/set-namespace/Dockerfile
mengqiy, line 1: Why using an older go version? 1.15 is the latest.

functions/go/set-namespace/README.md
```diff
@@ -1,28 +0,0 @@
-# Set Namespace Config Function
```
mengqiy, line 1: We should have a README, but I'm OK adding the doc and example later.

functions/go/set-namespace/go.mod
```diff
@@ -1,5 +1,12 @@
-module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-namespace
+module main
 
 go 1.14
```
mengqiy, line 3: We should use 1.15

functions/go/set-namespace/go.mod
```diff
@@ -1,5 +1,12 @@
...
 
-require sigs.k8s.io/kustomize/kyaml v0.6.1-0.20200819135012-e6ee03e3e3e0
+require (
+	github.com/spf13/cobra v1.0.0
+	sigs.k8s.io/kustomize/api v0.6.4
+	sigs.k8s.io/kustomize/kyaml v0.9.4
+	sigs.k8s.io/yaml v1.2.0
+)
+
+replace sigs.k8s.io/kustomize/api v0.6.4 => sigs.k8s.io/kustomize/api v0.0.0-20201120230733-052a6b4e967b
```
mengqiy, line 12: I'm a little confused here. If 0.6.4 is a released version, we don't need this line.

functions/go/Makefile
```diff
@@ -0,0 +1,33 @@
+GOPATH := $(shell go env GOPATH)
+
+FUNCTIONS := \
```
mengqiy, line 3: I imagine we should have a docker build target for building image: `gcr.io/kpt-functions-trusted/`+fn_name

When there are fn from external contributors, we may need to have a `curated_function` list and a `contrib_function`list. The contrib fn will go to `gcr.io/kpt-functions-contrib/`

mengqiy (commented):
Thanks for helping with this.
Depending on if you want to or have cycle to write the example and README for this fn, we can add them in this PR or later. If you don't want or don't have cycle to work on the doc, I can take it over.

I agree with Frank that current tests are adhoc and hacking and we should build a testing framework and have a united way to test functions. But IMO it should not block this PR from merging.

functions/go/set-namespace/go.mod
Shell32-Natsu: We need some new changes in api module and it hasn't released. We can wait for api 0.6.5. It should be released this week or next week.

functions/go/Makefile
Shell32-Natsu: Currently functions in this catalog are named in this way: `gcr.io/kpt-functions/`+fn name.

@frankfarzan How do you think about this proposal.

frankfarzan: We're discussing this in @mengqiy's document, I'm advocating for a delineation of community contributed functions. I'll defer to you whether to make this change in a future PR though

frankfarzan, line 3: Add comments to variables you expect humans to edit in the future.

functions/go/publish-functions.sh
```diff
@@ -25,13 +25,14 @@ image_tag=${TAG:-latest}
 prefix="release-go-functions-"
 [[ "${image_tag}" = "${prefix}"* ]] && image_tag="${image_tag#$prefix}"
 
+make
```
frankfarzan, line 28: This should probably the either way around, Makefile has a command for also building and publisher docker image. It can execute helper bash scripts as implementation detail. The GH workflows can then invoke make for everything.

This can be done in a future PR.

functions/go/set-namespace/main.go
```diff
@@ -1,75 +1,139 @@
...
+
+Example:
+
+To add a namespace 'color' to all resources:
+
+apiVersion: v1
+kind: ConfigMap
+data:
+  namespace: color
+metadata:
```
frankfarzan, line 95: by convention, 'metadata' field should come before 'data'.

functions/go/set-namespace/main.go
```diff
@@ -1,75 +1,139 @@
...
+  name: my-config
+
+You can use key "fieldSpecs" to specify the resource selector you
```
frankfarzan, line 98: describe the default behavior. What happens when 'fieldSpecs' is not specified?

functions/go/set-namespace/main.go
```diff
@@ -1,75 +1,139 @@
...
+	})
+	cmd.Long = usage()
 	if err := cmd.Execute(); err != nil {
+		fmt.Println(err)
 		os.Exit(1)
 	}
 }
+
+func usage() string {
+	return `Update or add namespace.
```
frankfarzan, line 81: How does handle any CRD which may be cluster-scoped?
Does it handle refs to namespaces (i.e. hardcoded mapping in Kustomize today)?

In general, the function documentation has to be very precise and comprehensive with behavior and limitations.

functions/go/set-namespace/Dockerfile
```diff
@@ -1,9 +1,8 @@
...
 ENV CGO_ENABLED=0
 WORKDIR /go/src/
 COPY . .
-RUN go build -v -o /usr/local/bin/set-namespace ./
-
+RUN go build -v -o /usr/local/bin/function ./
 FROM alpine:latest
```
frankfarzan, line 6: We should standardize this base image across Go and TS functions.

functions/go/set-namespace/main.go
```diff
@@ -1,75 +1,139 @@
...
-		if len(clusterScopedResources) > 0 {
-			return fmt.Errorf("the app config should only include namespace-scoped resources. "+
-				"But the following cluster-scoped resources are found:\n%s\n", strings.Join(clusterScopedResources, "\n"))
+		dataField, err := getDataFromFunctionConfig(resourceList.FunctionConfig)
+		if err != nil {
+			return err
+		}
+		dataValue, err := yaml.Marshal(dataField)
+		if err != nil {
+			return err
```
frankfarzan, line 52: General comment about error handling, you're currently just `return err` everywhere. Have you thought about what those error messages look like, and whether it requires more context (wrapping error message)?

functions/go/set-namespace/main.go
```diff
@@ -1,75 +1,139 @@
...
-var namespace string
-var allowClusterScoped bool
+//nolint
+func main() {
+	var plugin *plugin = &KustomizePlugin
+	defaultConfigString := `
+- path: metadata/namespace
+  create: true
+- path: subjects
+  kind: RoleBinding
```
frankfarzan, line 24: Why is this the default? Why RoleBinding/ClusterRolebinding?

functions/go/set-namespace/main.go
```diff
@@ -1,75 +1,139 @@
+// This file will be processed and embedded to pluginator.
```
frankfarzan, line 1: where does unit tests for this logic live?

frankfarzan (changes requested):

functions/go/set-namespace/main.go
Shell32-Natsu: The main functions are tested in the pluginator. But I will add a test for this *function* as well.

functions/go/set-namespace/main.go
Shell32-Natsu: They are from the default config in kustomize. I will remove RoleBinding/ClusterRoleBinding

functions/go/set-namespace/main.go
Shell32-Natsu: This metadata is for function config. Should it be from data as wel?

functions/go/set-namespace/Dockerfile
mengqiy: `alpine:latest` may not be great
IMO we should use `alpine:3.12` for go functions and `node:lts-alpine3.12` for TS functions.
It's probably safe to use the latest patch releases. `alpine:3.12` points to the latest 3.12.* versions.

functions/go/set-namespace/go.mod
mengqiy: Please add a TODO comment here to remind us to do that.

functions/go/set-namespace/Dockerfile
Shell32-Natsu: @mengqiy It will be good to create an issue to discuss it.

functions/go/set-namespace/main.go
frankfarzan: Yes, this is general convention for any KRM that GVKNN should come before the spec.

functions/go/set-namespace/main.go
Shell32-Natsu: Updated the usage message.

functions/go/set-namespace/README.md
Shell32-Natsu: https://github.com/GoogleContainerTools/kpt/issues/1245

functions/go/Makefile
Shell32-Natsu: https://github.com/GoogleContainerTools/kpt/issues/1246

tests/set_namespace_go.sh
```diff
@@ -41,17 +41,22 @@ metadata:
...
+  kpt fn run --fn-path fc.yaml >out.yaml
+assert_contains_string out.yaml "namespace: example-ns"
 
 testcase "kpt_set_namespace_go_declarative_example"
 kpt pkg get "$CATALOG_REPO"/examples/set-namespace .
 kpt fn run set-namespace/configs --fn-path set-namespace/functions
 assert_contains_string set-namespace/configs/example-config.yaml "namespace: example-ns"
+
+testcase "kpt_set_namespace_go_docker_imperative_idempotency"
+kpt pkg get "$SDK_REPO"/example-configs example-configs
```
frankfarzan, line 54: @mengqiy We should have all the example packages in the catalog repo. This is currently using examples in SDK repo which tightly couples the SDK and functions in the catalog.

frankfarzan (changes requested):

functions/go/set-namespace/Dockerfile
Shell32-Natsu: https://github.com/GoogleContainerTools/kpt/issues/1247

functions/go/set-namespace/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+      config.kubernetes.io/function: |
+        container:
+          image: gcr.io/example/foo:v1.0.0
+  data:
+    namespace: foo
+    fieldSpecs:
+    - path: metadata/namespace
+      create: true
+`
+	cmd := exec.Command("go", "run", ".")
```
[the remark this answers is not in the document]
Shell32-Natsu: There are some [issues](https://stackoverflow.com/a/55324723/3317502) to test main functions with unit tests in go.

tests/set_namespace_go.sh
Shell32-Natsu: The first 2 tests for set-namespace are using sdk repo.

functions/go/set-namespace/main.go
```diff
@@ -3,73 +3,145 @@ package main
...
 	if err := cmd.Execute(); err != nil {
+		fmt.Println(err)
 		os.Exit(1)
 	}
 }
+
+func usage() string {
+	return `Update or add namespace.
+
+Configured using a ConfigMap with the following keys:
```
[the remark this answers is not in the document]
Shell32-Natsu: There are only `namespace` and `fieldSpecs` in data field. Will add fieldSpec.

functions/go/set-namespace/main.go
```diff
@@ -3,73 +3,145 @@ package main
...
+kind: ConfigMap
+metadata:
+  name: my-config
+data:
+  namespace: color
+
+You can use key "fieldSpecs" to specify the resource selector you
+want to use. By default, the function will use this field spec:
+
+- path: metadata/namespace
```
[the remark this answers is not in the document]
Shell32-Natsu: 1. `path` is a field in `fieldSpec` and it is reused by multiple different transformers.
2. Is it impossible that user's CRDs won't have namespace in different paths?

functions/go/set-namespace/main.go
```diff
@@ -3,73 +3,145 @@ package main
...
+  create: true
+
+This means a "metadata/namespace" field will be added to all resources
+with namespaceable kinds. This information is collected from Kubernetes
+openAPI schema. "create: true" means a field 'metadata/namespace' will
+be created if it doesn't exist. To support your own CRDs you will beed
```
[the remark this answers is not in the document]
Shell32-Natsu: I don't get your point. What do you mean 'what the native k8s resources are covered'? Are you talking about the API schema version?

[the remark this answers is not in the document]
frankfarzan: "This information is collected from Kubernetes openAPI schema"

Currently, when you run this function with kpt, which resources are covered in the OpenAPI spec?

functions/go/set-namespace/main_test.go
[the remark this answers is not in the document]
frankfarzan: To clarify, my point is that this should either be a:
- e2e tests which invokes the function using KPT CLI
- unit tests for underlying go package that implements the function

main_test.go is kind of a weird hybrid.

tests/set_namespace_go.sh
frankfarzan: They shouldn't. We should have all the example packages checked in the catalog repo. Otherwise, a PR in SDK repo can break tests here.

functions/go/publish-functions.sh
frankfarzan: @mengqiy @Shell32-Natsu Probably want to add a name to the TODO or track all these issues in a bug. Anti pattern would be adding a TODOs that are not addressed.

functions/go/set-namespace/main.go
frankfarzan: How was this addressed?

functions/go/set-namespace/main.go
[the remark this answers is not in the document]
frankfarzan: I think this would be the too low-level an abstraction to provide to the user.  Otherwise, we can just provide a generic function called `transformer` that has two fields:

1. fieldSpec
2. Value to set on fields matched by fieldSpec

This is similar to Search and Replace functionality. 

Instead, a user should just able to use a function without any configuration. The function should be smart enough to change a namespace across resource. In other words, the complexity should be taken care by the function instead of pushed as a burden to end users. We're not there yet, this requires two things:
1. Ability to configure the OpenAPI document used by kpt (design in progress)
2. OpenAPI document capturing references to namespaces.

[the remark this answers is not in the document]
Shell32-Natsu: User can use the function without configuration. That's why we have a default fieldSpec. We can get rid of fieldSpec after the OpenAPI implementation. If there is no `path`, user cannot customize the function.

functions/go/set-namespace/main.go
Shell32-Natsu: `usage` function provides the --help message.

tests/set_namespace_go.sh
Shell32-Natsu: https://github.com/GoogleContainerTools/kpt/issues/1248

functions/go/set-namespace/main.go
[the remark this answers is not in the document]
Shell32-Natsu: Added more details about API schema.

functions/go/set-namespace/main_test.go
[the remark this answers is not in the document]
Shell32-Natsu: Removed main_test.go and add NamespaceTransformer_test.go

functions/go/set-namespace/Dockerfile
mengqiy: I have added a section in my doc and you guys should get an email notification for that. PTAL.

functions/go/publish-functions.sh
frankfarzan: Ping. Who's following up on this? Is that a follow up PR?

functions/go/set-namespace/main.go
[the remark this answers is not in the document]
frankfarzan: I think we're in agreement this this is a crutch until the function itself can automatically handle the use case.

@mengqiy We may want to capture this in a backlog issue and add TODOs here.

frankfarzan (approved):
approved once few comments are addressed. Please wait for @mengqiy as well.

mengqiy (approved):
I'm going to file issues to track the open issues shortly.

functions/go/set-namespace/main.go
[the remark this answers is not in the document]
mengqiy: Filed https://github.com/GoogleContainerTools/kpt/issues/1252.

functions/go/publish-functions.sh
mengqiy: filed https://github.com/GoogleContainerTools/kpt/issues/1251

PR #124, opened 2020-11-20
Add set-label function

Shell32-Natsu:


Shell32-Natsu:
cc @mengqiy

Shell32-Natsu:
CI failed due to Go package 4d63.com/gochecknoglobals/checknoglobals not available

frankfarzan:
I'll review this after https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/123 is approved since many comments will be applicable to both.

Shell32-Natsu:
@frankfarzan Ready for review

Shell32-Natsu:
e2e test is now using new driver.

@mengqiy @frankfarzan

functions/go/set-label/LabelTransformer_test.go
```diff
@@ -0,0 +1,195 @@
+package main
```
mengqiy, line 1: Are these tests generated from the tests in kustomize or manually written?

functions/go/set-label/main.go
```diff
@@ -0,0 +1,191 @@
...
+Example:
+
+To add a label 'color: orange' to all resources:
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-config
+data:
+  label_name: color
```
mengqiy, line 105: Do we support adding more than 1 labels at the same time?

mengqiy (commented):
Overall looks good.

functions/go/set-label/LabelTransformer_test.go
Shell32-Natsu: I added them manually to avoid importing bunch of kustomize internal test utilities.

functions/go/set-label/main.go
Shell32-Natsu: No, this function only supports one label a time.

mengqiy: Comparing with the functionality in kustomize, user can specify multiple labels under `commonLabels` in `kustomization.yaml`. 
When a kustomize user tries to use this function, he/she may expect it to be able to change multiple labels at once.

We need to decide this for set-label and set-annotation. @frankfarzan WDYT?

If we want to do multiple, we may have something like this:
```
type functionConfigData struct {
	Labels  []label `json:"labels,omitempty" yaml:"labels,omitempty"`
}

type label struct {
	LabelName  string `json:"label_name,omitempty" yaml:"label_name,omitempty"`
	LabelValue string `json:"label_value,omitempty" yaml:"label_value,omitempty"`
}
```

Shell32-Natsu: Updated the PR to support multiple labels.

examples/set-label/functions/fn-config.yaml
```diff
@@ -8,5 +8,12 @@ metadata:
...
     config.kubernetes.io/local-config: 'true'
 data:
-  label_name: color
-  label_value: orange
+  specs:
+  - label_name: color
+    label_value: orange
+  - label_name: fruit
+    label_value: apple
+    fieldSpecs:
```
mengqiy, line 16: It seems this is per-label config. Why? Should it apply to all of the labels?

examples/set-label/functions/fn-config.yaml
```diff
@@ -8,5 +8,12 @@ metadata:
         image: gcr.io/kpt-functions/set-label
     config.kubernetes.io/local-config: 'true'
 data:
-  label_name: color
-  label_value: orange
+  specs:
```
mengqiy, line 11: I saw you use `rn.Pipe(kyaml.Lookup("data", "specs"))` below.
Why don't we define a corresponding go struct and simply use unmarshal? IMO the latter approach is easier to reasonable about.

examples/set-label/functions/fn-config.yaml
Shell32-Natsu: The default config is applied to all labels. Each label can have it's own custom spec.

examples/set-label/functions/fn-config.yaml
Shell32-Natsu: I don't think we should implement our own struct for function config given this part may be used by other functions and they have already had well-defined utilities in kyaml.

examples/set-label/functions/fn-config.yaml
mengqiy: I see. Thanks for clarification.

examples/set-label/functions/fn-config.yaml
mengqiy: Sorry for not being very clear.
I meant `data` is standard field for `ConfigMap`, but `data.specs` is not. `kyaml.Lookup("data", "specs")` this feels like hardcoding the path.
IMO it's better to treat whatever under `.data` as a blob and unmarshal the whole blob to `addLabelSpecs`.

`addLabelSpecs` may look like this:
```
type addLabelSpecs struct {
	Labels []addLabelSpec `json:"labels,omitempty" yaml:"labels,omitempty"`
	FieldSpec types.FieldSpec `json:"fieldSpec,omitempty" yaml:"fieldSpec,omitempty"` // Field config for all labels
}

type addLabelSpec struct { // you already defined this in functions/go/set-label/main.go
	LabelName  string            `json:"label_name,omitempty" yaml:"label_name,omitempty"`
	LabelValue string            `json:"label_value,omitempty" yaml:"label_value,omitempty"`
	FieldSpecs []types.FieldSpec `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"` // Field config for one label.
}
```
Then you have something like:
```
d := kyaml.Lookup("data")
yaml.Unmarshal(d, &ads)
```

Does it make sense to you?

Shell32-Natsu: done

mengqiy (approved):
LGTM

mengqiy:
Let's wait for some time to see if Frank want to take a look.

examples/set-label/functions/fn-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+  annotations:
+    config.k8s.io/function: |
+      container:
+        image: gcr.io/kpt-functions/set-label
+    config.kubernetes.io/local-config: "true"
+data:
+  labels:
+  - label_name: color
+    label_value: orange
+  - fieldSpecs:
```
frankfarzan, line 14: @droot We need to rationalize these with `filters` at the Pipeline level.

@Shell32-Natsu @mengqiy For context, instead of each function implementing this filtering mechanism, we can provide this at a higher level in the `Pipeline` which makes function implementation simpler.

examples/set-label/.expected/diff.patch
```diff
@@ -0,0 +1,55 @@
...
++++ b/configs/example-config.yaml
+@@ -2,7 +2,11 @@ apiVersion: v1
+ kind: ConfigMap
+ metadata:
+   name: the-map
++  labels:
++    color: orange
++    fruit: apple
+ data:
++  fruit: apple
```
frankfarzan, line 13: what's this being changed?

examples/set-label/.expected/diff.patch
```diff
@@ -0,0 +1,55 @@
+diff --git a/configs/example-config.yaml b/configs/example-config.yaml
+index 1c43349..02e8ab4 100644
+--- a/configs/example-config.yaml
++++ b/configs/example-config.yaml
+@@ -2,7 +2,11 @@ apiVersion: v1
+ kind: ConfigMap
+ metadata:
+   name: the-map
++  labels:
```
frankfarzan, line 9: Do tests currently check for idempotency?

frankfarzan (changes requested):

examples/set-label/.expected/diff.patch
Shell32-Natsu: Because there is a custom field spec for this.

examples/set-label/.expected/diff.patch
Shell32-Natsu: The test driver doesn't check it. However, the unit test covers the idempotency of the transformer and I have updated the tests to 
```
  - label_name: color
    label_value: orange
  - label_name: color
    label_value: orange
  - fieldSpecs:
    - kind: ConfigMap
      create: true
      path: data
    label_name: fruit
    label_value: apple
```

examples/set-label/.expected/diff.patch
frankfarzan: Not following, what is a function for setting labels changing fields that are not labels?

examples/set-label/.expected/diff.patch
frankfarzan: @mengqiy This should be done by the e2e test harness.

examples/set-label/.expected/diff.patch
Shell32-Natsu: This is used to test the custom field specs. Users can use it to add labels to any field in any resource.

examples/set-label/.expected/diff.patch
Shell32-Natsu: https://github.com/GoogleContainerTools/kpt/issues/1396
 Will do it. But that doesn't block this PR.

mengqiy: Yes, our test driver should cover idempotency.
Thanks for bring this up!

examples/set-label/functions/fn-config.yaml
droot: After looking at this and offline discussions, here is the summary:
`fieldSpecs` provided here is the schema information that `set-label` needs to support different types. So this completely orthogonal to the `filtering` mechanism in the pipeline.

Another bit that confused me is that currently schema and input args (label/value in this case) is mixed together that makes the usage look very different from the case where schema is not provided, so there is an opportunity to unify that as well. 

so, TL;DR is we need a mechanism to make schema information available to all KRM functions in an easy way. And while doing so, we should also unify the UX for taking input for the functions with or without schema info.

frankfarzan: To expand on this, I consider the `fieldSpec` to be a low-level crutch until we address the general problem of augmenting the OpenAPI document with such semantics. A general solution would address all the following:

- Which field in a type is a `label`?
- Where is a field references from other resources?
- etc.

Kustomize has hardcoded these for k8s native types, we need to generalize this to address other resources types, e.g. Cloud Connector types.

examples/set-label/.expected/diff.patch
frankfarzan: Even though this is technically true, I think saying that ConfigMap.data is a label creates confusion. Instead use a custom resource with a label field.

frankfarzan (changes requested):

examples/set-label/.expected/diff.patch
Shell32-Natsu: Updated the test case.

frankfarzan (approved):

frankfarzan:
Some post-commit comments:
- Go files should use c_style_naming and not CamelCase naming
- Optional: Consider having directory organization for each function. Currently, everything is flat. `TS` functions have `build` to contain `Dockerfile` and `src` for sources. For Golang, we may need to also have `build` for `Dockerfile` and `pkg` for source files. I don't feel strongly about this though.

Shell32-Natsu:
Will update go file names in separate pr

PR #125, opened 2020-11-21
Encryption support for sops function

aodinokov:
Added support for encrypt operation for the function.
That extends the functionality and list of potential use-cases for this function:
it's possible to encrypt the newly generated secrets or rotate them.

The description, set of examples and e2e tests were update accordingly

### DIFF
--- a/examples/sops/local-configs/function.yaml
+++ b/examples/sops/local-configs/function.yaml
@@ -1,10 +1,32 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: my-config
+  name: my-decrypt-config
   annotations:
     config.k8s.io/function: |
       container:
         image: gcr.io/kpt-functions/sops
+        envs:
+        - SOPS_IMPORT_PGP
     config.kubernetes.io/local-config: "true"
 data:
+  cmd: 'decrypt'
+  cmd-json-path-filter: '$[?(@.metadata.name=="toDecrypt")]'
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-encrypt-config
+  annotations:
+    config.k8s.io/function: |
+      container:
+        image: gcr.io/kpt-functions/sops
+        envs:
+        - SOPS_IMPORT_PGP
+    config.kubernetes.io/local-config: "true"
+data:
+  cmd: 'encrypt'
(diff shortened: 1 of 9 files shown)


examples/sops/local-configs/function.yaml
```diff
@@ -1,10 +1,33 @@
...
+kind: ConfigMap
+metadata:
+  name: my-encrypt-config
+  annotations:
+    config.k8s.io/function: |
+      container:
+        image: gcr.io/kpt-functions/sops
+        envs:
+        - SOPS_IMPORT_PGP
+        #- SOPS_PGP_FP
```
frankfarzan, line 26: Why is this commented?

functions/ts/src/sops.ts
```diff
@@ -187,23 +248,49 @@ function readSopsArguments(configs: Configs) {
 
 sops.usage = `
 Sops function (see https://github.com/mozilla/sops).
-So far supports only decrypt operation:
-runs sops -d for all documents that have field 'sops:' and put the decrypted result back.
 
-Can be configured using a ConfigMap with the following flags:
-ignore-mac: true [Optional: default empty] Ignore Message Authentication Code during decryption.
+The fucntion supports 2 operations: encrypt and decrypt.
```
frankfarzan, line 252: Can you provide a typical function config example for each mode of operation? (Think of this as example in a MAN page so the user can get something working quickly without reading all of the docs).

frankfarzan (changes requested):

examples/sops/local-configs/function.yaml
aodinokov: there are 2 supported options how to provide this parameter:
1. via function parameter pgp, e.g.:
pgp: 'FBC7B9E2A4F9289AC0C1D4843D16CEE4A27381B4'

2. or via SOPS_PGP_FP env variable.

I commented env variable because I used the first option.

functions/ts/src/sops.ts
aodinokov: Done. I appreciate any feedback on how clear the examples I provided.

frankfarzan (approved):

mengqiy:
Is this PR ready to merge?

aodinokov:
> Is this PR ready to merge?
@mengqiy, 
yes. and it has been approved already.

PR #130, opened 2020-12-21
add dependency update for helm inflator

Shell32-Natsu:
`zookeeper` chart used in the test is updated to include a dependency. `helm-inflator` needs to update the dependency before running `template` command.

### DIFF
--- a/functions/ts/src/helm_inflator.ts
+++ b/functions/ts/src/helm_inflator.ts
@@ -162,11 +162,28 @@ function getTemplateArgs(configMapData: Map<string, string>): string[] {
   return args;
 }
 
+// run help dependency update command
+async function runHelmDepUpdate(configMapData: Map<string, string>) {
+  if (configMapData.get(LOCAL_CHART_PATH) === undefined) {
+    return;
+  }
+
+  const args: string[] = [
+    'dependency',
+    'update',
+    configMapData.get(LOCAL_CHART_PATH)!,
+  ];
+
+  runHelmCommand(args);
+}
+
 // run helm template command
 async function runHelmTemplate(
   configs: Configs,
   configMapData: Map<string, string>
 ) {
+  // Update helm dependency
+  await runHelmDepUpdate(configMapData);
   // Validate config data and read arguments.
   const args = getTemplateArgs(configMapData);
   args.unshift('template');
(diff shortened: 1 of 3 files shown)


functions/ts/src/helm_inflator.ts
```diff
@@ -228,8 +248,7 @@ metadata:
     config.kubernetes.io/function: |
       container:
         image: gcr.io/kpt-functions/helm-inflator
-        network:
-          required: true
+        network: true
```
mengqiy, line 251: Why format change?

Shell32-Natsu: The `network` field in function config has been updated months ago but this example hasn't updated.

tests/helm_inflator.sh
```diff
@@ -56,7 +56,7 @@ assert_contains_string out.yaml "name: extra-args-redis-master"
 
 helm_testcase "kpt_helm_inflator_imperative_pipeline"
 kpt fn source example-configs |
-  kpt fn run --mount type=bind,src="$(pwd)/${CHARTS_SRC}",dst=/source --image gcr.io/kpt-functions/helm-inflator:"${TAG}" --as-current-user -- local-chart-path=/source/zookeeper name=my-zookeeper | 
+  kpt fn run --mount type=bind,src="$(pwd)/${CHARTS_SRC}",dst=/source,rw=true --image gcr.io/kpt-functions/helm-inflator:"${TAG}" --as-current-user --network -- local-chart-path=/source/zookeeper name=my-zookeeper |
```
frankfarzan, line 59: The UX here doesn't make sense IMO. At this point, this should fetch everything from the network instead of mounting host volume.

frankfarzan, line 59: The tests here need to be refactored per @mengqiy doc.

Currently, these tests use test data from the SDK repo and charts from https://github.com/bitnami/charts.git which is very brittle. Test data and dependencies should either live in the repo or if remote (e.g. Helm chart), needs to use a particular version. Otherwise tests will break randomly like this.

frankfarzan (changes requested):
Helm

tests/helm_inflator.sh
Shell32-Natsu: This test case is for using local charts in pipeline. There is another case for remote charts in pipeline.

Shell32-Natsu: Agree, all e2e tests are needed to be re-written after tests framework finalized.

frankfarzan: Understood. I don't see someone using this flow since it's too cumbersome, instead they would either:

1) Use Helm CLI for local charts
2) Use function for fetching remote packages
3) Use the hermetic function when it's available.

frankfarzan (approved):

PR #133, opened 2021-01-06
add a simple e2e test driver

Shell32-Natsu:
close https://github.com/GoogleContainerTools/kpt/issues/1259


Shell32-Natsu:
CI failed. Should be fixed after #131 merged.

tests/Makefile
```diff
@@ -14,11 +14,14 @@
...
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 .PHONY: test
 
-test: ## Run all e2e tests
+test: install-test-driver ## Run all e2e tests
 	TAG=$(TAG) ./e2e.sh
+
+install-test-driver: ## Install e2e test driver
+	cd e2e-driver && go test ./... && go install
```
mengqiy, line 27: Why you choose to use the test driver a binary instead of a library?

mengqiy (commented):
What are you guys opinions about using [ginkgo](http://onsi.github.io/ginkgo/).
We have been using it a lot in kubebuilder and k/k also uses it heavily.

tests/Makefile
Shell32-Natsu: How do we use the driver if it's library?

Shell32-Natsu:
ginkgo looks like some JS test framework, e.g. Jest. IMO it's not very important in this simple test driver.

tests/e2e-driver/internal/runner/runner.go
```diff
@@ -0,0 +1,180 @@
...
+		resultsFile := l[0].Name()
+		actualResults, err := ioutil.ReadFile(filepath.Join(resultsPath, resultsFile))
+		if err != nil {
+			return fmt.Errorf("failed to read actual results: %w", err)
+		}
+		expectedResults, err := ioutil.ReadFile(filepath.Join(tmpPkgPath, expectedDir, expectedResultsFile))
+		if err != nil {
+			return fmt.Errorf("failed to read expected results: %w", err)
+		}
+		actual := strings.TrimSpace(string(actualResults))
```
frankfarzan, line 146: Why is this necessary?

Either:
1- we care to test the exact raw text of results YAML -> Compare files as is
2- we just want to compare the YAML content -> test the parsed content of the yaml instead of raw text

Given we already assert exact git diff for resources, then I think you want #1?

tests/e2e-driver/internal/runner/runner.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		return nil
+	}
+
+	// compare diff
+	err = gitAddAll(tmpPkgPath)
+	if err != nil {
+		return err
+	}
+	err = gitCommit(tmpPkgPath, "second")
```
frankfarzan, line 160: I don't think you need to commit to get the diff?

tests/e2e-driver/internal/runner/config_test.go
```diff
@@ -0,0 +1,49 @@
+package runner_test
```
frankfarzan, line 1: Is unit testing this necessary?

tests/e2e-driver/internal/runner/config.go
```diff
@@ -0,0 +1,33 @@
+package runner
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"gopkg.in/yaml.v3"
+)
+
+// TestConfig is the config for a test
```
frankfarzan, line 10: These comments are kind of tautological and don't help the reader

examples/set-namespace/.expected/diff.patch
```diff
@@ -0,0 +1,51 @@
+diff --git a/configs/example-config.yaml b/configs/example-config.yaml
+index c3ca9ba..6dc6427 100644
+--- a/configs/example-config.yaml
++++ b/configs/example-config.yaml
+@@ -1,20 +1,21 @@
+ apiVersion: v1
++data:
```
frankfarzan, line 7: there's diff here due to YAML formatting that's not related to this function. Please run 'kpt cfg fmt' on the input package first.

tests/e2e-driver/internal/runner/config.go
```diff
@@ -0,0 +1,33 @@
...
+	Network bool   `json:"network,omitempty" yaml:"network,omitempty"`
+}
+
+// TestConfigs contains a list of TestConfig
+type TestConfigs struct {
+	Configs []TestConfig `json:"configs,omitempty" yaml:"configs,omitempty"`
+}
+
+// ConfigFromFile returns a list of TestConfig read from path
+func ConfigFromFile(path string) (*TestConfigs, error) {
```
frankfarzan, line 22: Where's the actual config file passed to this?

frankfarzan (changes requested):

tests/e2e-driver/internal/runner/runner.go
Shell32-Natsu: You need to commit to include untracked file (new files). Although I am not sure is it possible that kpt fn creates new files but commit can handle it even it does.

tests/e2e-driver/internal/runner/config_test.go
Shell32-Natsu: Will remove it.

tests/e2e-driver/internal/runner/runner.go
Shell32-Natsu: Yes. The reason that I add `TrimSpace` here is to avoid false positive test failure caused by newline char in the end of expected files. Some users will add \n in the end of file while some will not.

tests/e2e-driver/internal/runner/config.go
Shell32-Natsu: Sorry I don't get your point. Can you clarify?

examples/set-namespace/.expected/diff.patch
Shell32-Natsu: `set-namespace` function uses the library from kustomize/api and the output mapping order doesn't match the result of `kpt cfg fmt`. IMO we should use the real output from the diff.

mortent:
Overall this looks good to me. Just two questions:
* Could we set this up as a golang test instead of using a separate binary? So we could run the tests with `go test ./...`?
* If there is a defined structure for the packages in the repo, could we just scan for the test cases instead of requiring users to specify them in a config file?

mengqiy:
> Could we set this up as a golang test instead of using a separate binary? So we could run the tests with go test ./...?

I asked the same question in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/133#discussion_r553047418.
Here's what we did in kubebuilder:
- We have a small test framework in https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/utils
- We have the e2e test framework as golang tests. e.g. https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/v2

>If there is a defined structure for the packages in the repo, could we just scan for the test cases instead of requiring users to specify them in a config file?

+1

Shell32-Natsu:
> * If there is a defined structure for the packages in the repo, could we just scan for the test cases instead of requiring users to specify them in a config file?

I want to have functionality to control the behavior of each test case. For example, the network access.

Shell32-Natsu:
Updated the PR, now one can run the e2e tests by running `go test ./...` in `tests/e2e-test/tests`

tests/e2e-test/internal/runner/config.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// Example of a test config:
+//
+// 	pkgPath: foo/bar
+//	network: true
+type TestConfig struct {
+	// PkgPath is the path to the package which will be tested.
+	PkgPath string `json:"pkgPath,omitempty" yaml:"pkgPath,omitempty"`
+	// Network indicates whether enable network access for function or not
+	Network bool `json:"network,omitempty" yaml:"network,omitempty"`
```
frankfarzan, line 21: This is the only test data that's not contained within the package itself. Why not move this into the package so all it takes to test a package is the path?

You can then get rid of TestConfig all together.

tests/e2e-test/internal/runner/runner.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil, fmt.Errorf("path %s is not a directory", pkg)
+	}
+	return &Runner{
+		pkgPath: pkg,
+		pkgName: filepath.Base(pkg),
+		network: network,
+	}, nil
+}
+
+// Run runs the test
```
frankfarzan, line 43: Can this run multiple e2e tests concurrently?

tests/e2e-test/internal/runner/runner.go
```diff
@@ -0,0 +1,180 @@
...
+func (r *Runner) compareResult(exitErr error, tmpPkgPath, resultsPath string) error {
+	// get exit code
+	exitCode := 0
+	if e, ok := exitErr.(*exec.ExitError); ok {
+		exitCode = e.ExitCode()
+	} else if exitErr != nil {
+		return fmt.Errorf("cannot get exit code from %w", exitErr)
+	}
+	// compare exit code
+	b, err := ioutil.ReadFile(filepath.Join(tmpPkgPath, expectedDir, expectedExitCodeFile))
```
frankfarzan, line 111: I suggest extracting a type for encapsulating reading and unmarshalling data from package under test (put):

put.exitCode -> int
put.results -> Results

This would make this logic more readable.

frankfarzan (changes requested):

tests/e2e-test/internal/runner/runner.go
Shell32-Natsu: I think it can. Will try it.

tests/e2e-test/tests/e2e_test.go
```diff
@@ -0,0 +1,71 @@
...
+		r, err := runner.NewRunner(c)
+		if err != nil {
+			return fmt.Errorf("failed to run test: %w", err)
+		}
+		go r.Run(retErr[i])
+	}
+	for i := range retErr {
+		err := <-retErr[i]
+		if err != nil {
+			return fmt.Errorf("%s: %w", (*cases)[i], err)
```
mengqiy, line 67: Should we collect the status of other tests instead of only returning the first failing test?

Shell32-Natsu: Good idea

examples/set-namespace/.expected/diff.patch
mengqiy: It's a little hard to read when the diff contains reordering.
I wonder if we should run `kpt cfg fmt` as part of `kpt fn` to enforce the order.

mengqiy (commented):
Only one comment regarding https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/133/files#r553530985.
Other pieces LGTM

examples/set-namespace/.expected/diff.patch
Shell32-Natsu: From the test's perspective, we don't really care about the content in `diff.patch`. What we really care is the diff between expected diff and actual diff. So the readability in `diff.patch` is not too important here.

+1 for running `fmt` in `fn run`. Will consider it in the pipeline implementation.

mengqiy:
@Shell32-Natsu A test is failing in the CI. Please take a look.

tests/e2e-test/internal/runner/runner.go
```diff
@@ -0,0 +1,237 @@
...
+
+	// run function
+	// TODO: change to pipeline when it's ready
+	kptArgs := []string{"fn", "run", tmpPkgPath, "--results-dir", resultsPath}
+	if r.network {
+		kptArgs = append(kptArgs, "--network")
+	}
+	o, fnErr := runCommand("", "kpt", kptArgs)
+	// run formatter
+	_, err = runCommand("", "kpt", []string{"cfg", "fmt", tmpPkgPath})
```
mengqiy, line 93: Please add a TODO here that links to the GH issue. It should remind us to drop this line after the issue is fixed.

mengqiy (commented):
One nit. Otherwise LGTM

tests/e2e-test/internal/runner/runner.go
Shell32-Natsu: I think this line is ok to be left here. Some function may use its own implementation and doesn't respect the original map order. It's better to explicitly run `kpt cfg fmt` after running function.

tests/e2e-driver/internal/runner/runner.go
frankfarzan: but why is the user editing the results file directly. Shouldn't be auto-generated by kpt?

examples/set-namespace/.expected/diff.patch
```diff
@@ -0,0 +1,38 @@
...
+ spec:
+   ports:
+   - name: etcd-server-ssl
+diff --git a/functions/fn-config.yaml b/functions/fn-config.yaml
+deleted file mode 100644
+index 47eab5e..0000000
+--- a/functions/fn-config.yaml
++++ /dev/null
+@@ -1,11 +0,0 @@
+-apiVersion: v1
```
frankfarzan, line 28: what's with this diff?

tests/e2e-driver/internal/runner/runner.go
Shell32-Natsu: If we can guarantee that the results file will be copied (not the content) then we can avoid trimming spaces. And if someone open the expected results file with editor that will modify the `\n` in the file end (at least VS Code has such option, some git tool may do it as well) the difference between the expected and actual results is very hard to identify.

frankfarzan (approved):

examples/set-namespace/.expected/diff.patch
Shell32-Natsu: kpt will remove the fn-config.yaml after running. I am not sure is this expected behacior.

PR #135, opened 2021-01-07
use node-forge >=0.10.0

mengqiy:


mengqiy:
It seems it doesn't work. I'm no expert in JS/TS. 
I'm not sure what would be the right thing to do here. @frankfarzan @Shell32-Natsu suggestions?

Shell32-Natsu:
Could you remind me why do we need to add this?

mengqiy:
https://app.snyk.io/org/kpt-fn-bot/project/d53cb2b3-2d15-4e59-bf5d-74574c5b898d

Shell32-Natsu:
I guess the 0.8.5 version is included via `Introduced through: ts-functions@0.1.2 › create-kpt-functions@0.17.0 › @kubernetes/client-node@0.10.3 › openid-client@2.5.0 › node-jose@1.1.4 › node-forge@0.8.5` so we cannot simply update the node-forge version in the top level package.json

frankfarzan:
create-kpt-functions@0.17.0 is the library in the SDK repo here: 

https://github.com/GoogleContainerTools/kpt-functions-sdk/tree/master/ts/create-kpt-functions

We should enable Snyke on the SDK repo. It may be the case that we can upgrade its dependency on the k8s's node client (@kubernetes/client-node@0.10.3) to fix this. @mengqiy WDYT?

PR #139, opened 2021-01-09
Add licenses checker for go

mengqiy:


Shell32-Natsu:
It will be good to have a target in the top level Makefile which checks license for both Go and TS functions.

functions/go/Makefile
```diff
@@ -61,6 +63,29 @@ push: $(FUNCTION_PUSH) ## Push images to registry. WARN: This operation should o
...
 
+.PHONY: check-licenses
+check-licenses: go-licenses $(FUNCTION_CHECKLICENSES)
+
+.PHONY: $(FUNCTION_CHECKLICENSES)
+$(FUNCTION_CHECKLICENSES):
+	$(MAKE) CURRENT_FUNCTION=$(subst -CHECKLICENSES,,$@) func-check-licenses
+
+# find or download go-licenses
+go-licenses:
```
frankfarzan, line 74: This seems like a hack. Can you just curl the pre-built binary? Note that unlike other tests, you don't need to run this across platforms/arch.

frankfarzan (changes requested):

functions/go/Makefile
mengqiy: > Can you just curl the pre-built binary?

No. Unfortunately, they don't release pre-built binary.

Using `go get` is their suggested way to install it: https://github.com/google/go-licenses#installation

frankfarzan: OK, could you extract lines 76 to 83 into a bash script? ./download_go_license.sh

In general mixing bash and Makefile makes things unreadable.

frankfarzan (changes requested):

Shell32-Natsu:
Will this license check make the CI fail?

mengqiy:
Yes, it has been added to `.github/workflows/ci.yaml`.

frankfarzan (approved):

PR #147, opened 2021-01-15
Support building functions with multiple versions for go and ts

mengqiy:
For example, we can run the following to build the images for a function:
`GIT_TAG=go/set-namespace/v1.2.3 ./scripts/function-release.sh build`
We will get the image tagged as
gcr.io/kpt-fn/set-namespace:v1.2.3
gcr.io/kpt-fn/set-namespace:v1.2
gcr.io/kpt-fn/set-namespace:v1

CD system can run the following to push the images.
`GIT_TAG=go/set-namespace/v1.2.3 ./scripts/function-release.sh push`

For ts functions, we can run something like:
`GIT_TAG=ts/kubeval/v1.2.3 ./scripts/function-release.sh build|push`

fixes: https://github.com/GoogleContainerTools/kpt/issues/1358


scripts/get-versions.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+VERNUM='0|[1-9][0-9]*'
+
+# An example is "set-namespace/v1.2.3"
```
frankfarzan, line 5: Why do you need the name of the function instead of just `v.1.2.3`?

scripts/go-function-build.sh
```diff
@@ -2,9 +2,12 @@
...
+source ../../scripts/get-versions.sh
+
 image_tag=${TAG:-dev}
+tags=$(validate_version "${image_tag}")
 prefix="release-go-functions-"
 [[ "${image_tag}" = "${prefix}"* ]] && image_tag="${image_tag#$prefix}"
-image_name=gcr.io/kpt-functions/"${CURRENT_FUNCTION}"
-image="${image_name}":"${image_tag}"
-docker build -t "${image}" -t "${image_name}" -f "${CURRENT_FUNCTION}"/Dockerfile "${CURRENT_FUNCTION}"
+for tag in ${tags}; do
```
frankfarzan, line 11: you probably want to invoke `docker build` once and then use `docker tag` instead of building N times.

frankfarzan (changes requested):

scripts/get-versions.sh
mengqiy: The release is going to be git tag driven.
If we push a git tag `v1.2.3`, we can't tell which function this version is for.
The prefix in the git tag can be useful when there are multiple release targets in one repo.
This is similar to [what kustomize does](https://github.com/kubernetes-sigs/kustomize/tags).

functions/go/Makefile
```diff
@@ -54,7 +55,7 @@ build: $(FUNCTION_BUILDS) ## Build all function images. 'TAG' env is used to spe
 
 .PHONY: $(FUNCTION_BUILDS)
 $(FUNCTION_BUILDS):
-	$(MAKE) CURRENT_FUNCTION=$(subst -BUILD,,$@) TAG=$(TAG) func-build
+	$(MAKE) CURRENT_FUNCTION=$(subst -BUILD,,$@) GCR_REGISTRY=$(GCR_REGISTRY) TAG=$(TAG) func-build
```
Shell32-Natsu, line 58: Do you really need to set `GCR_REGISTRY` again?

scripts/git-tag-parser.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+VERNUM='0|[1-9][0-9]*'
+
+# An example is "set-namespace/v1.2.3"
+GIT_TAG_SEMVER_REGEX="\
+^([a-zA-Z]*(-[a-zA-Z]+)*)\/\
```
Shell32-Natsu, line 7: You may need to accept every valid docker image (`_`, `.` are also valid in image name), or make it less strict.

mengqiy:
The e2e test is failing https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/147/checks?check_run_id=1711106740, but I can't reproduce it locally. @Shell32-Natsu Any ideas?

Shell32-Natsu:
I cannot reproduce either. Will continue investigating.

Shell32-Natsu:
Please rebase this PR.

mengqiy:
I can't reproduce the failure locally. I have kpt 0.37.1 locally.

The diff exists in `fn-config.yaml`. I suspect it's because the kpt version is too old.

The following is what we got:
```
            diff --git a/functions/fn-config.yaml b/functions/fn-config.yaml
            index 47eab5e..418bd0e 100644
            --- a/functions/fn-config.yaml
            +++ b/functions/fn-config.yaml
            @@ -2,6 +2,7 @@ apiVersion: v1
             kind: ConfigMap
             metadata:
               name: my-config
            +  namespace: example-ns
               annotations:
                 config.k8s.io/function: |
                   container:
```
The following is the expected:
```
            diff --git a/functions/fn-config.yaml b/functions/fn-config.yaml
            deleted file mode 100644
            index 47eab5e..0000000
            --- a/functions/fn-config.yaml
            +++ /dev/null
            @@ -1,11 +0,0 @@
            -apiVersion: v1
            -kind: ConfigMap
            -metadata:
            -  name: my-config
            -  annotations:
            -    config.k8s.io/function: |
            -      container:
            -        image: gcr.io/kpt-functions/set-namespace
            -    config.kubernetes.io/local-config: 'true'
            -data:
            -  "namespace": "example-ns"
```

mengqiy:
We are using the latest kpt in CI test: https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/.github/workflows/ci.yaml#L90

Shell32-Natsu:
What is interesting is only this PR has this failure.

Shell32-Natsu:
The latest version of kpt is 0.37.1. I tried in a brand new container with same method to install kpt and still cannot reproduce.

Shell32-Natsu:
You need to add a step to build `latest` tag images. Because some tests are using latest tag instead of dev.

scripts/go-function-build.sh
```diff
@@ -2,9 +2,26 @@
...
+fn_name=${fn_name:-"${CURRENT_FUNCTION}"}
+GCR_REGISTRY=${GCR_REGISTRY:-gcr.io/kpt-functions}
+
+cd "${scripts_dir}"/../functions/go
+docker build -t "${GCR_REGISTRY}/${fn_name}:${DEV_TAG}" -f "${fn_name}"/Dockerfile "${fn_name}"
+
+# TODO(https://github.com/GoogleContainerTools/kpt/issues/1368): stop building
+# docker images with tag "latest".
+echo tagging "${GCR_REGISTRY}/${fn_name}:latest"
+docker tag "${GCR_REGISTRY}/${fn_name}:${DEV_TAG}" "${GCR_REGISTRY}/${fn_name}:latest"
```
Shell32-Natsu, line 23: Any special reason to hardcode the latest tag here? Can't we add a step in the `.github/ci.yaml` to build latest tag images?

mengqiy: >Any special reason to hardcode the latest tag here? 

No special reason. But I prefer to be explicit.

> Can't we add a step in the .github/ci.yaml to build latest tag images?

We have a step `make build` in `ci.yaml`. `make build` invokes `go-function-build.sh` underlying.
Why suggesting add the step in `ci.yaml`?

Shell32-Natsu: Add a `make TAG=latest build` so that we don't need to add a un-skippable latest tag here.

mengqiy: I assume the build cache is shared across steps in GitHub Actions. It should not double the build time.

Shell32-Natsu: It's not about the build time. When I build images locally I may only want to build dev tag, or only latest tag. This can be controlled by make command. If latest is hardcoded in the script, I cannot avoid building it if I don't want it.

Shell32-Natsu (approved):
lgtm

mengqiy:
@frankfarzan PTAL

mengqiy:
@Shell32-Natsu I pushed another commit to support pushing images. PTAL

mengqiy:
I'm going to make some changes to make it easier for CD pipeline.

functions/go/Makefile
```diff
@@ -11,9 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-SHELL=/bin/bash
+
```
Shell32-Natsu, line 14: If you are trying to avoid using `bash`, why do you still use `bash` in the scripts below?

mengqiy: Makefile by default uses `sh`.
I'm trying to avoid setting `SHELL` unless necessary, since in our CD pipeline we may use [docker image](https://hub.docker.com/_/docker) which is alpine-based and it doesn't have `bash` but it has `sh`.
I saw it complained when we set `SHELL=/bin/bash`. Things seems to be working fine after removing this line.

Shell32-Natsu: If our environment may not have bash, then the shell scripts below shouldn't use bash as well.

mengqiy: I have bring this line back. Let's just install bash in alpine.

mengqiy:
@Shell32-Natsu Start to see the test failure related to the helm-inflator:
In the test, we got `helm.sh/chart: nginx-8.4.1`, but we expect `helm.sh/chart: nginx-8.4.0`.
Our test should not fail when a new chart version is released. The test should be pinned to a specific version of a chart.

Shell32-Natsu:
https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/154

hack/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM docker
```
frankfarzan, line 1: Add a comment describing what this is for. Otherwise it's not clear what a file called `hacker/Dockerfile` is for.

scripts/git-tag-parser.sh
```diff
@@ -0,0 +1,65 @@
...
+
+    echo v"${major}"."${minor}"."${patch}"
+    echo v"${major}"."${minor}"
+    echo v"${major}"
+  else
+    echo "${version}"
+  fi
+}
+
+function parse_git_tag {
```
frankfarzan, line 37: it's good to document all exported functions with example input/output.

e.g.

Invocation: `parse_git_tag ARG1 ARG2`
Returns `X`

scripts/go-function-release.sh
```diff
@@ -0,0 +1,46 @@
...
+
+set -euo pipefail
+
+scripts_dir="$(dirname "$0")"
+# git-tag-parser.sh has been shell-checked separately.
+# shellcheck source=/dev/null
+source ${scripts_dir}/git-tag-parser.sh
+
+DEV_TAG=dev
+versions=$(get_versions "${TAG}")
```
frankfarzan, line 25: wheres's TAG coming from? Each bash script should have a top-level description describing its usage including any  env vars it expects.

scripts/ts-function-release.sh
```diff
@@ -0,0 +1,56 @@
...
+source ${scripts_dir}/git-tag-parser.sh
+
+DEV_TAG=dev
+versions=$(get_versions "${TAG}")
+GCR_REGISTRY=${GCR_REGISTRY:-gcr.io/kpt-functions}
+
+cd "${scripts_dir}/../functions/ts/${CURRENT_FUNCTION}"
+
+# This make it work for npm 6.*.*
+sed -i.bak "s|gcr.io/kpt-functions|${GCR_REGISTRY}|g" package.json
```
frankfarzan, line 31: Why are we editing package.json like this?

Can you just change registry in the functions from kpt-functions to new one?

mengqiy: This is the reason: https://github.com/GoogleContainerTools/kpt/issues/1394

scripts/go-function-release.sh
mengqiy: > wheres's TAG coming from?

In `function-release.sh`, we set TAG. `CURRENT_FUNCTION="${fn_name}" TAG="${fn_ver}" make func-build`.

mengqiy:
Added more comments. PTAL

scripts/ts-function-release.sh
frankfarzan: Can you add a TODO here and follow up on that issue. There must be an alternative provided by NPM.  Hack here should be removed soon.

frankfarzan (approved):

PR #149, opened 2021-01-19
test for suggest-psp

Shell32-Natsu:


tests/e2e-test/internal/runner/util.go
```diff
@@ -42,7 +42,7 @@ func gitCommit(d, msg string) error {
 	if err != nil {
 		return fmt.Errorf("git config error: %w, output: %s", err, o)
 	}
-	o, err = runCommand(d, "git", []string{"commit", "-m", msg})
+	o, err = runCommand(d, "git", []string{"commit", "-m", msg, "--allow-empty"})
```
mengqiy, line 45: In what scenario, there will be no diff?

Shell32-Natsu: I am not familiar with this function. But if a function is used to validate the resources then there might be no diff.

mengqiy: When the hydration pipeline work is done, ideally, this function can be specified in the validation section and then the test harness doesn't need to do `--allow-empty`.
We can improve that later.

mengqiy (approved):

mengqiy:
The test is failing with `Error:  Error: Helm command template extra-args /source --values /source/values-production.yaml results in error: Error: open /source/values-production.yaml: no such file or directory`.
I thought this was fixed in the other PR.

Shell32-Natsu:
Yes I need to rebase

tests/e2e-test/internal/runner/util.go
Shell32-Natsu: Another possibility is the transformer is conditional.

frankfarzan (approved):

PR #151, opened 2021-01-22
add e2e test for helm inflator

Shell32-Natsu:
Use bitnami/nginx instead of bitnami/redis to avoid inconsistent results.


mengqiy:
> Use bitnami/nginx instead of bitnami/redis to avoid inconsistent results.

I don't see you delete anything related to `redis ` in this PR. Did I missing anything?

Shell32-Natsu:
I will create a separate PR to remove old shell script test.

mengqiy (approved):

PR #152, opened 2021-01-25
cleanup e2e tests

Shell32-Natsu:
Tests for sops is remained since it cannot use test driver.


mengqiy (approved):
Awesome!

mengqiy:
Snyk is pending forever in this PR. I guess that's because I re-config snyk just now after https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/148 is merged.
It's OK to ignore it, since it doesn't change any dependencies.

PR #153, opened 2021-01-26
SOPS: Adding cmd-tolerate-failures flag

aodinokov:
This change just introduces some degree of flexibility to control error handling

This flag allows to skip failure if
some documents were not decrypted/encrypted.
This may be useful for serveral use-cases:
1. several attempts to decrypt with different
   secrets sets: use the chain of sops fn
   and set cmd-tolerate-failures: false in
   all of them except the last one. the
   last one will fail if there are some
   undecrypted/unencrypted docs left.
2. ability to use that documents values
   for futher processing for people
   who don't have secrets to decrypt.
   May be useful for debugging perposes.

Also adjusted the code in a way
so the fucntion doesn nothing if
cmd is unknown: 'no-op' behavior.

### DIFF
--- a/functions/ts/sops/src/sops.ts
+++ b/functions/ts/sops/src/sops.ts
@@ -1,5 +1,6 @@
 import {
   Configs,
+  Severity,
   KubernetesObject,
   isKubernetesObject,
   generalResult,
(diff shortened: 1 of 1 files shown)


aodinokov:
failing on this.. looks like it's not sops-related, but will look into that later
--- FAIL: TestE2E (0.16s)
    --- FAIL: TestE2E/../../../tests/helm-inflator (10.13s)
        e2e_test.go:56: failed to run test: actual diff doesn't match expected

Shell32-Natsu:
Don't mind, this error is irrelevant to your PR. Will fix it soon.

mengqiy:
It seems Synk test is somehow stuck when process `functions/ts/helm-inflator/package.json`. Let me try to close and reopen to see if it resolves it.

frankfarzan (approved):

PR #154, opened 2021-01-26
add option to specify remote chart version

Shell32-Natsu:
Fix CI. Now the helm-inflator e2e test will pin to a specific version of chart which can avoid issues from chart updating.


mengqiy (approved):
Thanks for fixing the CI!

PR #155, opened 2021-01-27
improve e2e driver to run function multiple times

Shell32-Natsu:
fixes: https://github.com/GoogleContainerTools/kpt/issues/1396


examples/annotate-config/.krmignore
```diff
@@ -0,0 +1 @@
+.expected
```
mengqiy, line 1: What if we don't have this line here?
What if there are no yaml files in .expected? When do we have to have this line?

examples/suggest-psp/.expected/config.yaml
```diff
@@ -0,0 +1,2 @@
+exitCode: 0
```
mengqiy, line 1: This line is not necessary?

tests/e2e-test/internal/runner/config.go
```diff
@@ -2,13 +2,52 @@ package runner
...
+		config.RunTimes = 1
+	}
+	return config, nil
+}
+
 // TestCase contains the information needed to run a test. Each test case
 // run by this driver is described by a `TestCase`.
-type TestCase string
+type TestCase struct {
+	Path   string
```
mengqiy, line 48: There is field called `pkgPath` in `Runner`. These 2 become a little confusing. Please add some comments.

tests/e2e-test/tests/e2e_test.go
```diff
@@ -13,18 +13,22 @@ import (
...
+//  - 'config.yaml' has the configuration information for this test. It should contain a
+//    map of following fields:
+//    - 'exitCode': Contains a single number which is expected exit code of command
+//      'kpt fn run'. If this field is missed, driver will assume the expected exit code
+//      is 0.
+//    - 'network': A boolean which indicates whether network should be enabled
+//    	for this test. If this field exists and the content in it is 'true' then the
+//    	network is accessible. Otherwise the function cannot access network.
+//    - 'runTimes': A number which sepcifies the times of function running. If this field
+//      missing then the function will be run once.
```
mengqiy, line 25: Should we run more than once by default?

examples/annotate-config/.krmignore
Shell32-Natsu: kpt will read the yaml files from .expected (results.yaml and config.yaml) and process them just like they are regular k8s resources. It's better to always have this line but it can be ignored when there is no yaml file in .expected.

examples/suggest-psp/.expected/config.yaml
Shell32-Natsu: done

tests/e2e-test/internal/runner/config.go
Shell32-Natsu: removed `pkgPath` in `Runner`

tests/e2e-test/tests/e2e_test.go
Shell32-Natsu: Updated the default value to 2.

examples/suggest-psp/.expected/config.yaml
```diff
@@ -0,0 +1 @@
+runTimes: 2
```
mengqiy, line 1: nit: 2 is the default now. So this line is no longer needed.

mengqiy (approved):

PR #156, opened 2021-02-01
make test harness importable

Shell32-Natsu:
Make the e2e test harness importable so it can be reused in kpt pipeline e2e test.

@natasha41575


tests/e2etest/tests/e2e_test.go
```diff
@@ -3,7 +3,7 @@ package e2etest
 import (
 	"testing"
 
-	"github.com/GoogleContainerTools/kpt-functions-catalog/tests/e2etest/internal/runner"
```
natasha41575, line 6: is it possible to fix the typo that causes the difference between the name of the module here and the name of the directory?
https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/tests/e2e-test/go.mod

Shell32-Natsu: The name of the directory has been updated.

natasha41575: Ah, my mistake, I didn't see that. Thank you

mengqiy:
I expect the end states are:
- The e2e test harness in the kpt-functions-catalog repo is merged with the test harness in the kpt repo.
- The test harness in the kpt repo is stable enough. IOW, it is not updated very often.
- kpt-functions-catalog repo depends on the test harness in the kpt repo for e2e tests.

Making it exportable can be the 1st step to transition to the end states mentioned above.

mengqiy (approved):

PR #157, opened 2021-02-02
Split curated and contrib functions in the build and release

mengqiy:
We use `unstable` as image tag name until we have stable release for these kpt functions.
This PR also starts to use gcr.io/kpt-fn and gcr.io/kpt-fn-contrib repos for the functions that live in this repo.


mengqiy:
@Shell32-Natsu @frankfarzan This is ready for review.

Shell32-Natsu (approved):
I think we should have a release after this otherwise the example will not work for users.

mengqiy:
Yes. Tag `unstable` will be built from HEAD on every new commit. The CD pipeline is expected to be ready later this week.

PR #168, opened 2021-02-23
[HOLD] Add function set-name-prefix

Shell32-Natsu:
close https://github.com/GoogleContainerTools/kpt/issues/1474


droot:
I want to ensure we have following scenario covered. Let say we have a nested package scenario:

app/
       Kptfile
	   deployment.yaml
	   db/
	       Kptfile
		   statefulset.yaml
		  
```yaml
# app/Kptfile

pipeline:
  mutators:
  - image: gcr.io/kpt-fn/set-nameprefix
    configMap:
      prefix: app-
```
```yaml
# app/db/Kptfile

pipeline:
  mutators:
  - image: gcr.io/kpt-fn/set-nameprefix
    configMap:
      prefix: db-
```
Now when the pipeline is applied on `app` package, the name of the statefulset in `db` package will have prefix `app-db-`. The results of the operation will be saved `in-place`. The subsequent runs of the pipeline operation should not lead to repeated prefixing of `app-db-`.

Shell32-Natsu:
@droot I think this is how this function works. For idempotency what I am concern is the function doesn't know a name prefix is added by previous function or it's a part of the original name. Currently I treat these 2 cases in the same way: they are all considered as a prefix.

mengqiy:
It is a good point to bring up this nested package in-place hydration example.
It seems we may have to use something like settes' comments. This is also something Brian mentioned in another discussion.
I imagine it may looks like:
```
metadata:
  name: app-my-deployment  # prefix added by kpt fn set-name-prefix: ["app-"]
```

```
metadata:
  name: app-db-my-statefulset  # prefix added by kpt fn set-name-prefix: ["app-", "db-"]
```

Shell32-Natsu:
kustomize does this via annotations. https://github.com/kubernetes-sigs/kustomize/blob/6518393f5d0e2c6dced984ee377bc8ff1fa71911/api/resource/resource.go#L276

mengqiy:
This way also work.  We have to leave some breadcrumbs to solve this idempotency issue for nested packages.
The difference is that if using setter-like comments, the prefix information is stored only in Git, while if using annotation, the prefix information is stored in both Git and kubeapiserver.

examples/set-name-prefix/.expected/diff.patch
```diff
@@ -0,0 +1,51 @@
...
+@@ -23,17 +23,17 @@ spec:
+ apiVersion: autoscaling/v2beta2
+ kind: HorizontalPodAutoscaler
+ metadata:
+-  name: nginx-hpa
++  name: dev-nginx-hpa
+ spec:
+   scaleTargetRef:
+-    name: nginx-deployment
++    name: dev-nginx-deployment
```
mengqiy, line 40: Why this field get changed when the [nameprefix config only have `.metadata/name`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/nameprefix.go#L9)?

functions/go/set-name-prefix/main.go
```diff
@@ -0,0 +1,207 @@
...
+data:
+  prefix: dev-
+  fieldSpecs:
+  - path: data/selector/name
+    kind: MyOwnKind
+`
+}
+
+func getDefaultConfig() (transformerConfig, error) {
+	defaultConfigString := builtinpluginconsts.GetDefaultFieldSpecsAsMap()["nameprefix"]
```
mengqiy, line 168: It seems currently we only use kustomize's [default config for nameprefix](https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/nameprefix.go#L7).
I was expecting this function also update the name reference. If not, IMO this function won't provide a lot of value.
@frankfarzan WDYT?

Shell32-Natsu: The name references are updated and that's the answer for your previous comment. The name references are updated in `nameref/nameref.go` file.

frankfarzan:
Could you come up with a brief design doc on this function before getting into implementation?

Shell32-Natsu:
Will do it

functions/go/set-name-prefix/main.go
mengqiy: Got it.
This confused me a bit because this function is called `getDefaultConfig` and it's in `main.go`. I thought it's the only place for loading all config. But `nameref/nameref.go` appear to have a function `getDefaultConfig` too.
I'd suggest load all configs in a places for better readability.

PR #169, opened 2021-02-24
Add tests for function failures

Shell32-Natsu:
Also updated the function to return errors through fn framework.

close https://github.com/GoogleContainerTools/kpt/issues/1488

### DIFF
--- a/functions/go/set-annotation/main.go
+++ b/functions/go/set-annotation/main.go
@@ -58,49 +58,67 @@ func setAnnotation(spec setAnnotationSpec,
 
 //nolint
 func main() {
-	var plugin *plugin = &KustomizePlugin
-	tc, err := getDefaultConfig()
-	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
-	}
-
-	resmapFactory := newResMapFactory()
-
-	pluginHelpers := newPluginHelpers(resmapFactory)
 
 	resourceList := &framework.ResourceList{}
 	resourceList.FunctionConfig = map[string]interface{}{}
 
 	cmd := framework.Command(resourceList, func() error {
-		resMap, err := resmapFactory.NewResMapFromRNodeSlice(resourceList.Items)
+		err := run(resourceList)
 		if err != nil {
-			return err
-		}
-		annotations, err := getAnnotations(resourceList.FunctionConfig)
-		if err != nil {
-			return fmt.Errorf("failed to get data.specs field from function config: %w", err)
-		}
-		for _, a := range annotations.Annotations {
-			err := setAnnotation(a, resMap, tc, pluginHelpers, plugin)
-			if err != nil {
-				return fmt.Errorf("failed to add annotation [%s: %s]: %w",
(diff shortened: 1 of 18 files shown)


functions/go/set-namespace/main.go
```diff
@@ -15,59 +15,73 @@ import (
...
+	dataValue, err := yaml.Marshal(dataField)
+	if err != nil {
+		return errors.Wrap(err, "error when marshal data values")
+	}
+
+	err = plugin.Config(pluginHelpers, dataValue)
+	if err != nil {
+		return errors.Wrap(err, "failed to config plugin")
+	}
+	if len(plugin.FieldSpecs) == 0 {
```
mengqiy, line 70: Like set-annotations and set-labels, we should check if `plugin.Namespace` is empty. If so, we should fail with an error clearly saying that namespace should not be empty.
Returning an error complaining `missing metadata.name` can be confusing for users.

functions/go/set-annotation/main.go
```diff
@@ -58,51 +58,69 @@ func setAnnotation(spec setAnnotationSpec,
...
-			err := setAnnotation(a, resMap, tc, pluginHelpers, plugin)
-			if err != nil {
-				return fmt.Errorf("failed to add annotation [%s: %s]: %w",
-					a.AnnotationName, a.AnnotationValue, err)
+			resourceList.Result = &framework.Result{
+				Name: "set-annotation",
+				Items: []framework.Item{
+					{
+						Message:  err.Error(),
+						Severity: "error",
```
mengqiy, line 73: I guess we can use this:
https://github.com/kubernetes-sigs/kustomize/blob/9a94c5ecd382dd4d9877f679fba7f91d9fb9645e/kyaml/fn/framework/types.go#L30

functions/go/set-label/main.go
```diff
@@ -56,52 +56,67 @@ func setLabel(spec setLabelSpec,
...
-			err := setLabel(l, resMap, tc, pluginHelpers, plugin)
-			if err != nil {
-				return fmt.Errorf("failed to add label [%s: %s]: %w",
-					l.LabelName, l.LabelValue, err)
+			resourceList.Result = &framework.Result{
+				Name: "set-label",
+				Items: []framework.Item{
+					{
+						Message:  err.Error(),
+						Severity: "error",
```
mengqiy, line 70: Same here

functions/go/set-namespace/main.go
```diff
@@ -15,59 +15,76 @@ import (
...
-
-		resourceList.Items, err = resMap.ToRNodeSlice()
-		if err != nil {
-			return errors.Wrap(err, "failed to convert resource map to items")
+			resourceList.Result = &framework.Result{
+				Name: "set-namespace",
+				Items: []framework.Item{
+					{
+						Message:  err.Error(),
+						Severity: "error",
```
mengqiy, line 29: Same here

mengqiy (commented):
One last minor comment. Otherwise LGTM

mengqiy (approved):

PR #170, opened 2021-02-24
Update json|yaml tag naming

mengqiy:
We should not repeat ourselves when the top level already has "annotatioins" or "labels".


Shell32-Natsu (approved):
lgtm. @droot You may need to update the pipeline tests.

PR #172, opened 2021-02-25
update the namespace name in set-namespace

Shell32-Natsu:


functions/go/set-namespace/main.go
```diff
@@ -164,11 +169,9 @@ them.
 }
 
 //nolint
-func getDefaultConfig() ([]types.FieldSpec, error) {
-	defaultConfigString := `
-- path: metadata/namespace
-  create: true`
-	var defaultConfig []types.FieldSpec
+func getDefaultConfig() (transformerConfig, error) {
+	defaultConfigString := builtinpluginconsts.GetDefaultFieldSpecsAsMap()["namespace"]
```
mengqiy, line 173: > builtinpluginconsts.GetDefaultFieldSpecsAsMap()["namespace"]

There seems to be a bug after looking at the configuration it points to:
```
namespace:
- path: metadata/namespace
  create: true
- path: metadata/name
  kind: Namespace
  create: true
- path: subjects      # it should be `subjects.namespace`
  kind: RoleBinding
- path: subjects      # it should be `subjects.namespace`
  kind: ClusterRoleBinding
- path: spec/service/namespace
  group: apiregistration.k8s.io
  kind: APIService
  create: true
```
For reference, the following is the definitions for RoleBinding.

```
type RoleBinding struct {
	metav1.TypeMeta
	metav1.ObjectMeta
	Subjects []Subject
	RoleRef RoleRef
}

type Subject struct {
	Kind string `json:"kind" protobuf:"bytes,1,opt,name=kind"`
	APIGroup string `json:"apiGroup,omitempty" protobuf:"bytes,2,opt.name=apiGroup"`
	Name string `json:"name" protobuf:"bytes,3,opt,name=name"`
	Namespace string `json:"namespace,omitempty" protobuf:"bytes,4,opt,name=namespace"`
}
```

Shell32-Natsu: No it's not a bug. `subjects` is specially handled by the transformer.

mengqiy: Got it.

mengqiy (approved):

PR #176, opened 2021-03-03
Add a catalog landing page

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1465

Allows for an index `README` to be generated from markdown of functions in release branches.
image

Adds a site that access functions' versions and examples from a dropdown, which is currently temporarily viewable at https://kpt-dev-catalog.web.app
image

Contributes to https://github.com/GoogleContainerTools/kpt/issues/1489


examples/index.html
```diff
@@ -0,0 +1,192 @@
+<!DOCTYPE html>
```
mengqiy, line 1: Do you have something like this for the main kpt site?
Ideally we should create a folder called `site` and store it there if possible.

mengqiy:
I tried to click one of the examples. It brings me to https://kpt-dev-catalog--functions-wta49njt.web.app/mutators/set-annotation/simple/, but with a blank page.

examples/README.md
```diff
@@ -0,0 +1,15 @@
+# KPT Function Catalog
```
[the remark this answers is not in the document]
etefera: It's currently a manual step but can be added to the Makefile and run on site generation PRs once that's finalized.

scripts/generate_catalog.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2021 The Kubernetes Authors.
```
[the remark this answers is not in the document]
etefera: Didn't notice the license differed here. Changed.

scripts/generate_catalog.go
```diff
@@ -0,0 +1,190 @@
...
+// Usage: generate_catalog SOURCE_MD_DIR/ DEST_GO_DIR/
+//
+// The command will create a README.md file under DEST_GO_DIR/ containing tables
+// of functions separated by function type.
+// <!--mdtogo:<VARIABLE_NAME>
+// ..some content..
+// -->
+//
+// <VARIABLE_NAME> must be Name, Type or Description.
+package main
```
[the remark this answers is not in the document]
etefera: Done.

scripts/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/GoogleContainerTools/kpt-functions-catalog
```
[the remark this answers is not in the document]
etefera: This was autogenerated while it was in the scripts directory but can change if need be

examples/index.html
etefera: We could do that in the long term, but that would break running it locally without something like nginx, since the site expects its contents like examples to be in child directories.

etefera:
> I tried to click one of the examples. It brings me to https://kpt-dev-catalog--functions-wta49njt.web.app/mutators/set-annotation/simple/, but with a blank page.

That's because there are currently no `README`s in the `simple` or `advanced` directories, even though that was the specified structure. We expect that to change once that content is created for each example, correct?

examples/index.html
mengqiy: > that would break running it locally without something like nginx

How do we bring up the catalog site locally? Why nginx can help here?

mengqiy:
> That's because there are currently no READMEs in the simple or advanced directories, even though that was the specified structure. We expect that to change once that content is created for each example, correct?

Thanks for pointing this out. It somehow thought we already have that. We will update it to match the specified structure. I filed https://github.com/GoogleContainerTools/kpt/issues/1522 to track it.

examples/index.html
etefera: Currently, you can do something like `cd examples/ && python -m SimpleHTTPServer` to start running. If the index and catalog README are in a directory outside of `examples`, we'd need something rerouting `/` to `../examples/`

frankfarzan:
@mengqiy Is the plan to host the catalog separate from the CLI given independent versioning? Doesn't need to block this, just making sure the end-goal is clear.

examples/mutators/set-namespace/README.md
```diff
@@ -1,5 +1,8 @@
 # Set Namespace
```
frankfarzan, line 1: @mengqiy @etefera I suggest authoring a doc using one of the functions as an example that includes all the content we have talked about:
- Function reference generated from --help
- One or more examples
- Category of function
- Link to source code (if open source)
- etc.

Once we have the desired UX hashed out for one function, then things fall into place. As it stands, the example README is being shown as the entirety of documentation for a function which is confusing.

frankfarzan (dismissed):

mengqiy:
> Is the plan to host the catalog separate from the CLI given independent versioning?

Yes, it will be hosted separately under `catalog.kpt.dev`. We will make the catalog site look like the kpt CLI site.

I created https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/178 to standarlize the documentaion for function examples. We will have one README.md per example.

The landing page for each function is generated from the help text of the function. We plan to use `kpt fn doc --image=` to get it, but the command is NOT implemented yet. (This part has changed a bit from our last discussion which is using a README.md) I will prioritize this task to unblock you.

About the landing page for the catalog, since the top level README (e.g. examples/mutators/set-namespace/README.md) is gone, we need to do something different. One way is to use the first paragraph of the help text as the short description.

examples/mutators/set-namespace/README.md
etefera: Updated site according to doc.

mengqiy:
The GitHub icon on the usage doc page should link to the source of the function e.g. https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/set-annotation/v0.1/functions/go/set-annotation/metadata.yaml#L5

etefera:
> The GitHub icon on the usage doc page should link to the source of the function e.g. https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/set-annotation/v0.1/functions/go/set-annotation/metadata.yaml#L5

Fixed.

mengqiy:
We have discussed offline is that we need to have a way to allow user to go back to a function's usage doc page for an example.

Since doing it using partial title is going to be hard, we can alternatively have something in the navigation bar on the left. e.g. list all functions and their examples in a tree structure. Only expand it for the function that the user is looking at. I'm not sure how hard to implement this way.

scripts/generate_catalog/generate_catalog.go
```diff
@@ -0,0 +1,321 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func parseMetadata(f function, metadataPath string, version string, versionDest string) (function, error) {
+	var buf bytes.Buffer
+	// Get the content of metadata.yaml from the appropriate release branch.
+	cmd := exec.Command("git", "show", metadataPath)
```
mengqiy, line 234: Why we use `git show` to get the content of a file? Is using `cat` more straight forward?

examples/README.md
```diff
@@ -0,0 +1,7 @@
...
```
mengqiy, line 1: It's OK to check this in for now.
I want to understand the long term plan. We will eventually generate `examples/README.md` in GH action and no longer need to check in this file, right?

etefera: That's correct; we can also remove it here.

scripts/generate_catalog/generate_catalog.go
```diff
@@ -0,0 +1,321 @@
...
+	var buf bytes.Buffer
+	cmd := exec.Command("git", "branch", "-a")
+	cmd.Stdout = &buf
+	err := cmd.Run()
+	if err != nil {
+		return verBranches, err
+	}
+
+	for _, branch := range strings.Split(buf.String(), "\n") {
+		segments := strings.Split(branch, "/")
```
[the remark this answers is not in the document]
etefera: Switched to full branch matching.

scripts/generate_catalog/generate_catalog.go
```diff
@@ -0,0 +1,321 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Usage: generate_catalog SRC_REPO_DIR/ DEST_MD_DIR/
+//
+// The command will create a README.md file under DEST_MD_DIR/ containing tables
+// of functions separated by function type. Source files for the catalog will
```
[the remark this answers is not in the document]
etefera: Fixed

examples/kubeval/simple/README.md
```diff
@@ -1,5 +1,8 @@
 # Kubeval
 
+<!--catalog:Name Kubeval-->
```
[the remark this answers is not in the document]
etefera: Yes, removed.

scripts/generate_catalog/generate_catalog.go
etefera: Switched to `git cat-file` to retrieve the file in a different branch.

examples/kubeval/simple/README.md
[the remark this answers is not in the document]
mengqiy: Why am I still seeing this change?

examples/README.md
mengqiy: Great! Then let's remove it here

mengqiy (approved):

mengqiy:
Next steps:
- Automate the deployment process with GH action
- Bring the site live

PR #177, opened 2021-03-09
implement starlark function

mengqiy:
fixes: https://github.com/GoogleContainerTools/kpt/issues/1496


examples/runtime/starlark/using-path/fn-config.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kpt.dev/v1beta1
```
mengqiy, line 1: This example is not tested since our test harness doesn't support mount yet.

examples/runtime/starlark/README.md
```diff
@@ -0,0 +1,17 @@
+# Starlark Runtime
+
+The `starlark` function provides a runtime to execute a starlark script that
+updates or validate KRM resources.
+
+Starlark source can be provided using `source` field.
```
frankfarzan, line 6: Talked about this bunch of times now: but we need to have a clear delineation between function help and examples.

The README is an example of using Starlark, but is describing a field called `source`, so it seems like a help page. But even then, it's not actually talking what custom resource is being described here (i.e. `StarlarkFunction`), so it's incomplete and ambigious.

Please make sure:
a) Examples README should talk about example. It's fine to explain how this particular example works (e.g. it could use the inline method).
b) Help page should be comprehensive and describe all the details and sematics of each field.

examples/runtime/starlark/inline/README.md
```diff
@@ -0,0 +1,21 @@
...
+kpt fn run inline
+```
+
+## Expected result
+
+Check the `.metadata.namespace` field has been set to `baz` for every resource.
+
+```sh
+kpt cfg cat inline
+```
```
frankfarzan, line 21: @phanimarupaka Regarding CLI usability work, `kpt cfg cat` is useful in documentation since it works across different OS'es.

examples/runtime/starlark/using-path/README.md
```diff
@@ -0,0 +1,21 @@
...
+In this example, starlark source is provided using path in the function
+configuration.
+
+## Function invocation
+
+Get the config example and try it out by running the following commands:
+
+```sh
+kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/runtime/starlark/using-path .
+kpt fn run --mount type=bind,src="$(pwd)/using-path/star-source",dst=/starlark/ using-path
```
frankfarzan, line 12: we should avoid using mount. Otherwise, how does this work for declarative functions?

examples/runtime/starlark/using-path/README.md
```diff
@@ -0,0 +1,21 @@
+# Starlark Path Example
+
+In this example, starlark source is provided using path in the function
```
frankfarzan, line 3: General note: I suggest expanding such sentences and being one level more precise and specific.

examples/runtime/starlark/using-url/.expected/config.yaml
```diff
@@ -0,0 +1,2 @@
+runCount: 2
+network: true
```
frankfarzan, line 2: why do you need network?

examples/runtime/starlark/using-url/README.md
```diff
@@ -0,0 +1,23 @@
+# Starlark URL Example
+
+In this example, starlark source is provided using URL in the function
```
frankfarzan, line 3: Let's avoid this pattern. For now only provide hermetic, inline starlark.

functions/go/starlark/config.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
+	"sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const (
+	fnConfigGroup      = "kpt.dev"
```
frankfarzan, line 13: Let's follow offline to standardize the group and version for all function configs. Using the kpt CLI group doesn't seem right.

@phanimarupaka

functions/go/starlark/main.go
```diff
@@ -0,0 +1,92 @@
...
+  inline: |
+    # set the namespace on each resource
+    def run(r, ns_value):
+      for resource in r:
+        # mutate the resource
+        resource["metadata"]["namespace"] = ns_value
+    # get the value to add
+    ns_value = ctx.resource_list["functionConfig"]["keyValues"]["foo"]
+    run(ctx.resource_list["items"], ns_value)
+keyValues:
```
frankfarzan, line 33: `data` for brevity? Is this used already and would break things?

functions/go/starlark/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+COPY go.mod go.mod
+COPY go.sum go.sum
+RUN go mod download
+COPY main.go main.go
+COPY config.go config.go
+RUN go build -o /usr/local/bin/star ./
+
+#############################################
+
+FROM alpine:3.12
```
frankfarzan, line 13: where' the interpreter for the python code?

frankfarzan (dismissed):
I like this!

functions/go/starlark/Dockerfile
mengqiy: Though Starlark is a dialect of Python, I don't think we should support Python in this function.
To clarify, this function and `kpt fn run --enable-star` both use the lib from kyaml to support Starlark. The kyaml library uses [`go.starlark.net/starlark`](https://github.com/google/starlark-go) as the Starlark interpreter.

functions/go/starlark/main.go
mengqiy: > Is this used already and would break things?

No. 

I will change it to `data`.

functions/go/starlark/config.go
mengqiy: Maybe we can use `fn.kpt.dev`?

functions/go/starlark/main.go
frankfarzan: Actually, what's the use case for this?

In this example, isn't it easier to just inline `baz` instead of referring to `ctx.resource_list["functionConfig"]["keyValues"]["foo"]`

examples/runtime/starlark/README.md
Shell32-Natsu: +1

Shell32-Natsu (commented):
Please add a test for unhappy path.

examples/runtime/starlark/using-path/README.md
mengqiy: Do you mean that we don't advocate using mount and URL by remove these 2 examples? Or we should remove the capability from the function?
If the latter, `StarlarkFunction` CRD can be simplified to
```
apiVersion
kind
metadata
source string
data map[string]string
```

frankfarzan: We definitely should support running remote scripts so let's remove that from the function.

It can be useful to avoid wrapping a script in KRM, but note:
1) Gatekeeper constraints already do this, so there's a precedence
2) We can shift the problem left and providing tooling to automate wrapping the script in KRM. The user shouldn't have to author KRM as that's a poor UX.

So don't advertise mount through examples for now.

frankfarzan: * should NOT support remote scripts

functions/go/starlark/main.go
mengqiy: These key-value pairs can be considered as parameters to the starlark script.
The user journey: A user writes a starlark script. The user wants to use this script in different pipelines to achieve slightly different customizations without changing the starlark script. The user can provide different values in these key-value pairs to achieve that.

frankfarzan: It sounds like you want reusability of scripts which is at odds with how inline works. One way to address this is to have two KRM resources:

StarlarkInvocation: This can either inline the script or have a reference to the StarlarkScript resource with key/value params.
StarlarkScript: Reusable script.

Maybe start with just StarlarkInvocation with inline script (not separate key/value). Later on you can introduce StarlarkScript.

Meta point: This kind of design should be done as a brief doc.

mengqiy:
Updated the error message according to the guideline. PTAL

functions/go/starlark/main.go
mengqiy: > start with just StarlarkInvocation with inline script (no separate key/value field)

It seems ConfigMap can do what we want here.

frankfarzan: We shouldn't abuse ConfigMap unless we really need simple key/value pairs. In this case, we need a custom resource which can be extended in the future when you introduce `StarlarkScript`. I wouldn't recommend using ConfigMap here. `StarlarkRun` may be a better name than `StarlarkInvocation`.

examples/runtime/starlark/inline/README.md
```diff
@@ -0,0 +1,43 @@
+# Starlark Inline Example
+
+In this example, we are going to demonstrate how to execute a starlark script
+in starlark function against the input KRM resources to update or validate them.
```
frankfarzan, line 4: This paragraph is too generic/ambiguous. What exactly is this example doing?

Keep in mind we don't need this to be a general introduction to Starkark, so lead by example.

examples/runtime/starlark/inline/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+The starlark script is provided inline in the function configuration like below:
+
+```
+apiVersion: fn.kpt.dev/v1beta1
```
frankfarzan, line 9: As in other PR, this should be followed by 1 or 2 sentences explain what this does.

examples/runtime/starlark/inline/fn-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: StarlarkFunction
+metadata:
+  name: my-star-fn
+  annotations:
+    config.k8s.io/function: |
+      container:
+        image: gcr.io/kpt-fn/starlark:unstable
+source: |
+  # set the namespace on each resource
+  def run(r, ns_value):
```
frankfarzan, line 11: do you need a function definition? Can you inline this?

examples/runtime/starlark/inline/README.md
frankfarzan: https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/183

examples/runtime/starlark/inline/fn-config.yaml
mengqiy: Echo from https://github.com/bazelbuild/starlark/blob/master/spec.md#for-loops:
> In Starlark, a for loop is permitted only within a function definition. A for loop at top level results in a static error.

examples/runtime/starlark/inline/README.md
mengqiy: Updated the example to follow the template.
Link to reference doc part has been skipped since we don't have it live.

examples/runtime/starlark/inline/fn-config.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: fn.kpt.dev/v1beta1
+kind: StarlarkFunction
+metadata:
+  name: my-star-fn
+  annotations:
+    config.k8s.io/function: |
+      container:
+        image: gcr.io/kpt-fn/starlark:unstable
```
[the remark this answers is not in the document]
mengqiy: Yes, we will keep using tag `unstable` for master branch. But we will use versioned tags on the release branches which are the branches that the site will use.

mengqiy:
Discussed with @droot offline today. Sunil is waiting for it and want to use this in some of his tests.
@Shell32-Natsu I have addressed all comments. Let's get this merged today.
If @frankfarzan has more comments, let's address them in a followup PR.

Shell32-Natsu (approved):

PR #181, opened 2021-03-11
use fn.kpt.dev instead of kpt.dev for Group name

mengqiy:
Also reorganize the `tests/` directory structure.

Original discussed in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/177#discussion_r590571449
We should get this in before trying to release these functions.

fixes: https://github.com/GoogleContainerTools/kpt/issues/1539


examples/mutators/set-annotation/advanced/fn-config.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: kpt.dev/v1beta1
+apiVersion: fn.kpt.dev/v1beta1
```
frankfarzan, line 1: I like the group name.

What's the plan for evolving the version? Why start at `v1beta1` instead of `v1alpha1`? When does it graduate to `v1`? How does this interact with semver for the function? I'm guessing once a function image semver goes to v1.0, then it should be taking a v1 of the function config?

mengqiy: At the beginning, I thought using `v1beta1` may make the users think that this config's API is more stable than using `v1alpha1`. But it's just my guessing, the user may not even care about that given that it's a client-side CRD.
Thinking a little more about it, using `v1alpha1` may be better when we want to deprecate the old schema.

Here's the plan now:
Let's use `v1alpha1` for the current client-side CRD. When we have OpenAPI stuff ready, let's make the new CRD `v1beta1`. After one month or two, we promote the `v1beta1` API to `v1` and make necessary changes based on user feedback. When a function goes to v1.0.0, ideally its config CRD should also reached `v1`.

frankfarzan: Sounds good

frankfarzan (approved):

PR #182, opened 2021-03-12
Add `httpbin-gen` kpt function

droot:
This is a PR to figure out how/where to place the code for `httpbin-gen` a test generator kpt function.

Pl. do not merge as is.


droot:
I am not comfortable having this generator function under `contrib` directory in the current form. I think we should create `test` directory under `functions` as a place for keeping all tests related kpt functions.  /cc @frankfarzan @Shell32-Natsu

frankfarzan:
@mengqiy At some point we had demo/test functions in the catalog (e.g. `team-crd`). Were they deleted?

I'm not opposed to having a third category of functions:

kpt-fn: Officially maintained
kpt-contrib: Contributed, questionable quality
kpt-test: For testing/demp purposes

mengqiy:
> I am not comfortable having this generator function under contrib directory in the current form.

+1 For functions that are not intended to be used by end-users should have their own home e.g. kpt-fn-test.

mengqiy:
> At some point we had demo/test functions in the catalog (e.g. team-crd). Were they deleted?

`team-crd` example is here:
https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/ts/demo-functions/src/expand_team_cr.ts

frankfarzan:
@droot This is an example of a generator function with client-side CRD:

https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/ts/demo-functions/src/expand_team_cr.ts

Image is here:

https://pantheon.corp.google.com/gcr/images/kpt-functions/GLOBAL/expand-team-cr?project=kpt-functions&organizationId=433637338589&gcrImageListsize=30

droot:
Awesome! Thanks @frankfarzan 
I will try to use `expand-team-cr` for my tests and if it works, then we won't need this PR.

Overall, we are in agreement in creating `test/demo` category of these tests functions.

droot:
So I gave `expand-team-cr` function today and observed the following:

- It isn't taking the input as `functionConfig`. I am guessing it is expecting the team-cr to be in resourcelist because when I did that it generated the expected output
- Its not idempotent. When I re-ran the pipeline, it  failed with error: `may not add resource with an already registered id: rbac.authorization.k8s.io_v1_RoleBinding|staging|developer`

@mengqiy as I am planning to add some more test functions to simulate `deleting of resources` etc. as well, we should add support for `test` category of the functions.

frankfarzan:
@droot Right, `expand-team-cr` doesn't have a function config. Instead it hydrated every instance of `Team` resource in the input. This actually makes sense for this use case.

Re idempotency: Can you dig into that error? Where is it being thrown. Note that this function is written in Typescript which provides an `insert` method using tuple of (apiVersion, kind, namespace, name) as the primary key.

https://googlecontainertools.github.io/kpt-functions-sdk/api/classes/_types_.configs.html#insert 

We need to figure out the general story around pruning and primary key invariant at the orchestrator level.

droot:
> @droot Right, `expand-team-cr` doesn't have a function config. Instead it hydrated every instance of `Team` resource in the input. This actually makes sense for this use case.

Re-read the documentation and yes, it makes sense.

> 
> Re idempotency: Can you dig into that error? Where is it being thrown. Note that this function is written in Typescript which provides an `insert` method using tuple of (apiVersion, kind, namespace, name) as the primary key.
> 
> https://googlecontainertools.github.io/kpt-functions-sdk/api/classes/_types_.configs.html#insert
> 
> This will handle prunning if the primary key doesn't change across invocations.

You are right, `expand-team-cr` is idempotent in this case. I created a new test package that uses just the `expand-team-cr` and repeated in-place runs are working perfectly fine. Actually, I accidentally created a pkg use-case without much thinking that resulted in this error. Shown below is the strange use-case where I am using this function with two other functions `set-namespace` and `set-label` as shown below, and I get the error where I am guessing `set-namespace` function is complaining about duplicate resource issue (many such[ instances](https://github.com/kubernetes-sigs/kustomize/issues/2747) have been reported in kustomize repo.

```yaml
apiVersion: kpt.dev/v1alpha2
kind: Kptfile
metadata:
  name: app-with-generator
pipeline:
  mutators:
  - image: gcr.io/kpt-functions/expand-team-cr
  - image: gcr.io/kpt-fn/set-namespace:unstable ## I clearly don't know what I am trying to do here with two namespaces already in the repo.
    configMap:
      namespace: staging
  - image: gcr.io/kpt-fn/set-label:unstable
    configMap:
      tier: db
```

> 
> We need to figure out the general story around pruning and primary key invariant at the orchestrator level.

Yes, this is a big gap. I will file an issue to track this. Second thing I learnt that debugging story is very weak currently, Ideally I would like to see the function that threw this error.

frankfarzan (commented):
> Yes, this is a big gap. I will file an issue to track this. Second thing I learnt that debugging story is very weak currently, Ideally I would like to see the function that threw this error.

For both logging and pruning, are you comparing the current behavior of kpt with the new `kpt fn render`? 
Current implementation should log what function is running. I'm not sure the current pruning logic (or lack of) top of my head, let's dig into it.

droot:
This PR is no longer needed as I used `starlark` to implement the same functionality https://github.com/GoogleContainerTools/kpt/pull/1554

PR #185, opened 2021-03-15
set-label v0.1.0 doc

mengqiy:


examples/mutators/set-label/v0_1_0/advanced/README.md
```diff
@@ -32,7 +32,7 @@ added to field `.spec.selector.labels` as well. We specify it in field
 Get the example config and try it out by running the following commands:
```
frankfarzan, line 32: is there a reason the directory uses `v0_1_0` instead of `v0.1.0`?

mengqiy: Initially I thought it will cause problem on Windows. But it appears not to be true after googling it.
I can change it to use dot.

mengqiy:
Updated it to use v0.1.0 as directory name

Shell32-Natsu (approved):

PR #189, opened 2021-03-17
use Entrypoint instead of CMD in Dockerfiles

mengqiy:
Also tweak Dockerfile to speed up build speed by avoiding repeatly downloading go modules.

This will make implementing `kpt fn doc` command easier. We don't need to specify the CMD. It now becomes `docker run <image-name> --help`.

### DIFF
--- a/functions/go/set-annotation/Dockerfile
+++ b/functions/go/set-annotation/Dockerfile
@@ -1,8 +1,15 @@
 FROM golang:1.15-alpine3.12
 ENV CGO_ENABLED=0
 WORKDIR /go/src/
+
+COPY go.mod go.sum ./
+RUN go mod download
+
 COPY . .
 RUN go build -o /usr/local/bin/function ./
+
+#############################################
+
 FROM alpine:3.12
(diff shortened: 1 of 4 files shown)


functions/go/set-annotation/Dockerfile
```diff
@@ -1,8 +1,15 @@
 FROM golang:1.15-alpine3.12
 ENV CGO_ENABLED=0
 WORKDIR /go/src/
+
+COPY go.mod go.sum ./
```
Shell32-Natsu, line 5: I don't think this can avoid duplicated downloading because each function is in a different container.

mengqiy: To be clear, this is optimized for the following workflow for a single container:
- go build
- change go source code
- go build again

Before:
Because we use `COPY . .`, if we update anything (e.g. help text in main.go), the cache for this layer will not be reused. It will download all go modules again.

After:
`go mod download` is in a earlier layer and cache can be reused if we don't change `go.mod` and `go.sum`.

It will be 10+s faster on subsequent builds.

Shell32-Natsu (approved):

PR #191, opened 2021-03-18
docs for set-annotation v0.1.1

mengqiy:
We don't keep the directory for v0.1.0 because the corresponding image doesn't work with kpt fn doc command.

In the catalog site, we won't display a page for v0.1.0.


mengqiy:
/assign @Shell32-Natsu 
cc: @etefera

Shell32-Natsu (approved):
It's a little bit annoying to update these docs every time when we release a new version.

PR #192, opened 2021-03-18
docs for set-label v0.1.1

mengqiy:
We don't keep the directory for v0.1.0 because the corresponding
image doesn't work with kpt fn doc command.


mengqiy:
/assign @Shell32-Natsu 
cc: @etefera

Shell32-Natsu (approved):

PR #197, opened 2021-03-24
Migrate policy-controller-validate to use kyaml

mengqiy:
The PR migrate policy-controller-validate to use kyaml lib from the kustomize repo instead of using the go SDK in the kpt-functions-sdk repo.
The code is ready for review.
But the documents are NOT ready now. I will update docs when #196 merges.

The old fn is at https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/pkg/functions/gatekeeper/validate.go

fixes: https://github.com/GoogleContainerTools/kpt/issues/1199


frankfarzan (changes requested):
Where's the existing function? Can you preserve the Git history and show the diffs?

frankfarzan:
> > Where's the existing function?
> 
> https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/pkg/functions/gatekeeper/validate.go

Can you do this in stages:

1. Move the function as is to this repo.
2. Make the changes

Otherwise, it's not clear what's being changed. Let me know if this is too much work for some reason.

mengqiy:
> Move the function as is to this repo.
Make the changes

It has already done in this way.
The last commit with name `Migrate policy-controller-validate to use kyaml` contains the real changes.

functions/go/policy-controller-validate/validate.go
```diff
@@ -151,7 +131,7 @@ func parseResults(results []*opatypes.Result) error {
 			return fmt.Errorf("could not cast to unstructured: %+v", r.Resource)
 		}
 		name := u.GetName()
-		path, found := u.GetAnnotations()[constants.SourcePathAnnotation]
+		path, found := u.GetAnnotations()[SourcePathAnnotation]
```
frankfarzan, line 134: our library provides utilities for this. Each functions shouldn't need to define this annotation constant and perform this low-level operation.

functions/go/policy-controller-validate/README.md
```diff
@@ -0,0 +1,126 @@
...
+### Overview
+
+<!--mdtogo:Short-->
+
+Validate the KRM resources using the policy controller.
+
+<!--mdtogo-->
+
+
+config-root/.../*-constraint.yaml and *-template.yaml define Policy Controller constraints and templates which all configs in config-root/ must pass.
```
frankfarzan, line 12: This is not well-formatted/phrased and doesn't belong in the overview section.

functions/go/policy-controller-validate/README.md
```diff
@@ -0,0 +1,126 @@
...
+To learn more about how to write constraint templates and constraints, see:
+https://cloud.google.com/anthos-config-management/docs/how-to/write-a-constraint-template
+and
+https://cloud.google.com/anthos-config-management/docs/how-to/creating-constraints.
+<!--mdtogo-->
+
+### Examples
+
+<!--mdtogo:Examples-->
+We have a constraint template which defines a rule that a config map can't have
```
frankfarzan, line 40: proper noun: config map -> `ConfigMap`

functions/go/policy-controller-validate/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+<!--mdtogo:Long-->
+You can use the policy controller to validate KRM resources. To learn more about
+the policy controller, see: https://cloud.google.com/anthos-config-management/docs/concepts/policy-controller.
+
+The function takes 3 types of resources from the input resource list:
+
+- constraint templates
+- constraints
+- KRM resources to be audited against
```
frankfarzan, line 24: I'm not sure if 'audit' is the established terminology here (Refer to Gatekeeper OSS docs).

Common phrasing: Constraint policies is enforced on KRM resources.

functions/go/policy-controller-validate/README.md
```diff
@@ -0,0 +1,126 @@
...
+  name: super-secret
+  namespace: default
+data:
+  private_key: sensitive data goes here
+```
+
+To run the function to audit the resources:
+
+```shell
+kpt fn run gcr.io/kpt-fn/policy-controller-validate:unstable .
```
frankfarzan, line 110: Why do we need examples in the usage doc, given there's example packages with better UX. In this case, it's up to the user to create the YAML files (which we don't explain), so there's room for confusion.

frankfarzan:
@maxsmythe FYI

functions/go/policy-controller-validate/validate.go
```diff
@@ -0,0 +1,150 @@
...
+		if !ok {
+			return fmt.Errorf("could not cast to unstructured: %+v", r.Resource)
+		}
+		name := u.GetName()
+		path, found := u.GetAnnotations()[SourcePathAnnotation]
+		if !found {
+			path = "?"
+		}
+		constraintName := r.Constraint.GetName()
+		msgs = append(msgs, fmt.Sprintf("%s\n\nname: %q\npath: %s\nviolatedConstraint: %s", r.Msg, name, path, constraintName))
```
frankfarzan, line 139: FunctionResults schema has fields for identifying the object and fields in an object. We should populate those instead of encoding it in the string message.

examples/validators/policy-controller-validate/invalid-configmap/.expected/results.yaml
```diff
@@ -0,0 +1,12 @@
+name: policy-controller-validate
+items:
+- message: |+
+    Found 1 violations:
+
+    [1] The following banned keys are being used in the config map: {"private_key"}
```
frankfarzan, line 6: The only non-structured content that needs to be part of the message are the following:

- `The following banned keys are being used in the config map: {"private_key"}`
- violatedConstraint

Other metadata should be conveyed using other fields here:
- name
- path

Number of violations is an emergent property of the function results.

functions/go/policy-controller-validate/validate.go
mengqiy: I updated it to use the SourcePathAnnotation definition from kyaml.
[`Result.Resource` is a `unstructured.Unstructured`](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/197/files/245c2066d786a31434de2978df02bf2906eafdd0..c636365a5b120b6b2d5cbb6e3d32b7283700a403#diff-a80d7be973fa160be84208745ee7f48d395becec2919f109f89d8fe3d14a8fe5R130). If we want to use the library from kyaml, we need to convert `unstructured.Unstructured` to `RNode` which requires one more round of marshal and unmarshal. It will take more time and I don't think it will make the code more readable.

examples/validators/policy-controller-validate/invalid-configmap/README.md
```diff
@@ -0,0 +1,73 @@
+# policy-controller-validate: invalid configmap
```
frankfarzan, line 1: Current OSS image is called `gatekeeper-validate`, there's a non-OSS image called `policy-controller-validate` as well, but it's not this.

https://googlecontainertools.github.io/kpt/guides/consumer/function/catalog/validators/

examples/validators/policy-controller-validate/invalid-configmap/.expected/results.yaml
```diff
@@ -0,0 +1,14 @@
+items:
+- message: |-
+    The following banned keys are being used in the config map: {"private_key"}
```
frankfarzan, line 3: @droot @Shell32-Natsu We need to think about what the kpt CLI output looks like for functions emitting structured results. It wouldn't be sufficient to just display the message string as it does not contain enough information unless the user looks at the FunctionResults.

frankfarzan: Related: How's the non-structured output being tested?

droot: > @droot @Shell32-Natsu We need to think about what the kpt CLI output looks like for functions emitting structured results. It wouldn't be sufficient to just display the message string as it does not contain enough information unless the user looks at the FunctionResults.

Yes, this has come up while writing the spec for FunctionResult and we didn't have a very good answer. I will schedule sometime tomorrow for us to go over this.

mengqiy:
I understand we may want to review the spec for `FunctionResult`. That may eventually requires us to change our doc for examples.
But is there something blocking this PR from merging?

examples/validators/gatekeeper-validate/invalid-configmap/README.md
```diff
@@ -0,0 +1,73 @@
+# gatekeeper-validate: invalid configmap
+
+## Overview
+
+This example demonstrates how to validate config maps using a constraint.
+
+There are 3 resources: a ConstraintTemplate, a K8sBannedConfigMapKeysV1 and a
+ConfigMap.
+The constraint disallows `private_key` to be used as a key in the config map.
```
maxsmythe, line 9: nit: should read like "The constraint disallows using `private_key` as a key in the config map."

functions/go/gatekeeper-validate/README.md
```diff
@@ -0,0 +1,41 @@
...
+and
+https://cloud.google.com/anthos-config-management/docs/how-to/creating-constraints.
+
+<!--mdtogo-->
+
+### Examples
+
+<!--mdtogo:Examples-->
+
+TODO: link to the examples
```
maxsmythe, line 39: Dangling TODO?

functions/go/gatekeeper-validate/README.md
```diff
@@ -0,0 +1,41 @@
...
+<!--mdtogo-->
+
+### Synopsis
+
+<!--mdtogo:Long-->
+
+You can use the policy controller to validate KRM resources. To learn more about
+the policy controller, see: https://cloud.google.com/anthos-config-management/docs/concepts/policy-controller.
+
+The function ensures the constraint policies are enforced on KRM resources.
```
maxsmythe, line 18: nit: "The function ensures the constraint policies are enforced on KRM resources." -> "The function evaluates constraint policies against KRM resources"

functions/go/gatekeeper-validate/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/gatekeeper-validate
+
+go 1.15
+
+require (
+	github.com/open-policy-agent/frameworks/constraint v0.0.0-20210317225149-4f80ac172ddf
+	github.com/open-policy-agent/gatekeeper v3.0.4-beta.2+incompatible
```
maxsmythe, line 7: We should use the most recent version of Gatekeeper that makes sense.

v3.0.4 is probably missing a few matchers, such as the scope matcher.

https://github.com/open-policy-agent/gatekeeper/releases/tag/v3.3.0

is probably a good starting point.

functions/go/gatekeeper-validate/validate.go
```diff
@@ -0,0 +1,171 @@
...
+
+	for _, r := range results {
+		u, ok := r.Resource.(*unstructured.Unstructured)
+		if !ok {
+			return fmt.Errorf("could not cast to unstructured: %+v", r.Resource)
+		}
+
+		item := framework.Item{
+			Message:  fmt.Sprintf("%s\nviolatedConstraint: %s", r.Msg, r.Constraint.GetName()),
+			Severity: framework.Error,
```
maxsmythe, line 137: Gatekeeper has a few different possible violation levels for a constraint now:

- `deny`
- `warn`
- `dryrun`

I'm not sure how you'll want to interpret these, (e.g. for the admission webhook,  `warn` allows the request but sends a warning ), but that could be useful metadata for users.

maxsmythe (changes requested):

functions/go/gatekeeper-validate/go.mod
mengqiy: I got the following error when trying to get it directly.
```
$ go get github.com/open-policy-agent/gatekeeper@v3.3.0
go get github.com/open-policy-agent/gatekeeper@v3.3.0: github.com/open-policy-agent/gatekeeper@v3.3.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v3
```

It seems it is because it has a semver major version of 2 or higher and is not a Go module yet.
I then got it using the hash ofv3.3.0 directly. It works, but it doesn't look so pretty in go.mod.

functions/go/gatekeeper-validate/validate.go
mengqiy: Updated.
PTAL

mengqiy:
@maxsmythe @frankfarzan PTAL

functions/go/gatekeeper-validate/go.mod
maxsmythe: Yeah, unfortunately there was a decision early on that it should be "v3" because there were 2 previous ways to integrate OPA with Kubernetes. I wish I had fought harder against that.

Gomod requires a specific directory structure for major versions >2, where the code must live in a subfolder `/vN` where `N` is the major version number. You'll need to use the hash.

At some point, I'd like to take the `target` code out of Gatekeeper and put it in its own repo, which should fix this.

examples/validators/gatekeeper-validate/invalid-configmap/resources.yaml
```diff
@@ -32,6 +32,7 @@ kind: K8sBannedConfigMapKeysV1
 metadata:
   name: no-secrets-in-configmap
 spec:
+  enforcementAction: deny
```
maxsmythe, line 35: note: this is superfluous (default is "deny"), but fine.

examples/validators/gatekeeper-validate/warning-only/README.md
```diff
@@ -0,0 +1,67 @@
...
+- what constraint does it violate
+- where does the resource live and its index in the file
+
+To pass validation, let's replace the key `private_key` in the ConfigMap in
+`resources.yaml` with something else e.g. `public_key`.
+Rerun the command. It will no longer have the warning.
+
+## Function Reference
+
+TODO: add the link
```
maxsmythe, line 67: Dangling TODO?

functions/go/gatekeeper-validate/main.go
```diff
@@ -54,11 +54,15 @@ func main() {
 			objects = append(objects, obj)
 		}
 
-		err := Validate(objects)
-		if result, ok := err.(*framework.Result); ok {
+		result, err := Validate(objects)
+		switch {
+		case result != nil && err != nil:
```
maxsmythe, line 59: Why are we using the result if the err != nil? This can sometimes be meaningful (see [Multiple return values](https://golang.org/doc/effective_go#multiple-returns) in Effective Go), but that doesn't appear to be the case here.

Usually you can use a pattern like this:

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

If you need to handle the result no matter what, I still would avoid the case switches, building up whatever return results gradually.

There also seems to be some inconsistency here as to what is getting returned when, so it's hard for me to say what the best approach would be.

examples/validators/gatekeeper-validate/invalid-configmap/.expected/results.yaml
```diff
@@ -0,0 +1,14 @@
+items:
```
frankfarzan, line 1: Can you add a TODO to test expected CLI output? (My previous comment)

This is important part of the UX.

examples/validators/gatekeeper-validate/invalid-configmap/README.md
```diff
@@ -0,0 +1,73 @@
+# gatekeeper-validate: invalid configmap
+
+## Overview
+
+This example demonstrates how to validate config maps using a constraint.
```
frankfarzan, line 5: commented before: `ConfigMap` not `config map`.

examples/validators/gatekeeper-validate/invalid-configmap/README.md
```diff
@@ -0,0 +1,73 @@
...
+- a detailed error message
+- what resource violates the constraints
+- what constraint does it violate
+- where does the resource live and its index in the file
+
+To pass validation, let's replace the key `private_key` in the ConfigMap in
+`resources.yaml` with something else e.g. `public_key`.
+Rerun the command. It will succeed (no output).
+
+## Function Reference
```
frankfarzan, line 71: What's function reference?

functions/go/gatekeeper-validate/README.md
```diff
@@ -0,0 +1,43 @@
+# gatekeeper-validate
+
+### Overview
+
+<!--mdtogo:Short-->
+
+Validate the KRM resources using the policy controller.
```
frankfarzan, line 7: "policy controller" is the admission controller and reserved for proprietary Google product. This is the OSS Gatekeeper constraint framework (Hence the previous discussion around name of the image). @maxsmythe Please check for correct usage.

functions/go/gatekeeper-validate/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+<!--mdtogo-->
+
+### Synopsis
+
+<!--mdtogo:Long-->
+
+You can use the policy controller to validate KRM resources. To learn more about
+the policy controller, see: https://cloud.google.com/anthos-config-management/docs/concepts/policy-controller.
```
frankfarzan, line 16: Ditto, this is not the OSS project

functions/go/gatekeeper-validate/main.go
```diff
@@ -0,0 +1,87 @@
...
+			resourceList.Result = result
+			return err
+		case result != nil && err == nil:
+			resourceList.Result = result
+			return nil
+		case result == nil && err != nil:
+			resourceList.Result = &framework.Result{
+				Name: "gatekeeper-validate",
```
frankfarzan, line 67: @Shell32-Natsu Why does the function need to populate this?

mengqiy: https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml@v0.10.16/fn/framework#Result
It says it should be `Name is the name of the function creating the result`.

frankfarzan: The orchestrator should set this automatically (it has all the information).

functions/go/gatekeeper-validate/README.md
maxsmythe: I'd say "using Gatekeeper constraints"

functions/go/gatekeeper-validate/README.md
maxsmythe: https://open-policy-agent.github.io/gatekeeper/website/docs/howto

examples/validators/gatekeeper-validate/invalid-configmap/.expected/results.yaml
mengqiy: Adding it to the `.expected` directory messes up with the test.
I filed https://github.com/GoogleContainerTools/kpt/issues/1689 and added in m2 milestone.

examples/validators/gatekeeper-validate/invalid-configmap/resources.yaml
mengqiy: Removed it.

examples/validators/gatekeeper-validate/invalid-configmap/README.md
mengqiy: Updated the name a bit.
It meant to be the reference doc for the function.

functions/go/gatekeeper-validate/main.go
mengqiy: Not setting `Name` any more.

mengqiy:
@maxsmythe PTAL
GH doesn't allow me to re-request review in their web UI :/

functions/go/gatekeeper-validate/main.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+			objects = append(objects, obj)
+		}
+
+		err := Validate(objects)
+		if err == nil {
+			return nil
+		}
+
+		if result, ok := err.(*framework.Result); ok {
```
maxsmythe, line 62: What happens if the typecast fails, which seems likely to be the case if this code from parseResults is hit?

```
		if !ok {
			return fmt.Errorf("could not cast to unstructured: %+v", r.Resource)
		}
```

Instead of returning an error which can then potentially be cast, maybe always return a list of results?

The filtering function seems like a good idea. I'm not sure what the significance of the return value is here (does it fail differently?)

maxsmythe: Looking at the code, it does seem functionally correct, but avoiding the typecast seems both simpler and less brittle.

mengqiy:
@maxsmythe Thanks for your review and PTAL

frankfarzan (approved):

PR #200, opened 2021-04-01
Run helm-inflator as current user

johnbelamaric:
Since `kpt` functions run as `nobody`, the commands as listed result in `Error: validation: chart.metadata is required`, unless your directory is world-readable.


Shell32-Natsu (approved):
lgtm

PR #202, opened 2021-04-01
Add metadata.yaml files and remove dir convention for examples

mengqiy:
Ensure the `emails` field has at least 2 people.
The description field is manually copied from the usage doc in either README.md or string in main.go
It should eventually be populated by a tool extract short description from usage doc.

If in release branches, sourceURL and examplePackageURLs should the release branch names.

We should have a CI test to ensure the URLs in the metadata file is up-to-date: https://github.com/GoogleContainerTools/kpt/issues/1703

fixes: https://github.com/GoogleContainerTools/kpt/issues/1619
fixes: https://github.com/GoogleContainerTools/kpt/issues/1598


mengqiy:
cc: @etefera @phanimarupaka

functions/go/apply-setters/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+image: gcr.io/kpt-fn/apply-setter
+description: Apply setter values on resource fields. Setters serve as parameters for template-free setting of field values.
+categories:
+  - mutator
+imageTags:
+  - unstable
+docPath: README.md
+sourceURL: https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/go/apply-setter
```
frankfarzan, line 8: What happens to these URLs when a function is versioned?

functions/go/apply-setters/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+image: gcr.io/kpt-fn/apply-setter
+description: Apply setter values on resource fields. Setters serve as parameters for template-free setting of field values.
```
frankfarzan, line 2: Are you saying this is automatically populated from usage doc? Where's the logic for that?

functions/go/search-replace/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+image: gcr.io/kpt-fn/search-replace
```
frankfarzan, line 1: Given this, I'm assuming you no longer will maintain a list of curated and contrib functions in the Makefile?

functions/go/search-replace/metadata.yaml
```diff
@@ -0,0 +1,14 @@
...
+description: Search and optionally replace fields across all resources.
+categories:
+  - mutator
+imageTags:
+  - unstable
+docPath: README.md
```
frankfarzan, line 7: is there a need for docPath? Either default it to README.md or remove it?

functions/go/apply-setters/metadata.yaml
mengqiy: Re. `sourceURL`, the current plan is only show the link for each function in the catalog landing page. That means we won't show links for different versions.
Users can use this link to go to the mater and then select the desired release tag or branch.

I have shared a design doc for doc site with you offline.

I'm a little hesitant about if we should maintain `imageTags`. It can be very easily out-of-date. An alternative is to drop it in the metadata.file and gather it from git tags when building the site.

functions/go/search-replace/metadata.yaml
mengqiy: Yes, we can achieve it with some changes to the scripts and Makefile

functions/go/search-replace/metadata.yaml
mengqiy: I guess we can default it to `README.md` if unspecified.

frankfarzan (approved):

PR #233, opened 2021-04-19
Make kubeval hermetic

mengqiy:
fixes: https://github.com/GoogleContainerTools/kpt/issues/1621

### DIFF
--- a/examples/kubeval/simple/README.md
+++ b/examples/kubeval/simple/README.md
@@ -1,19 +1,18 @@
-# Kubeval
+# kubeval: simple example
 
 The `kubeval` KRM config function validates Kubernetes resources using kubeval.
 Learn more on the [kubeval website].
 
-This example invokes the kubeval function against Kubernetes v1.18.0.
+This example invokes the kubeval function against the builtin Kubernetes
+v1.19.8 schema.
 
 ## Function invocation
 
 Get this example and try it out by running the following commands:
 
-<!-- TODO: no --network. See: https://github.com/GoogleContainerTools/kpt/issues/1621 -->
-
 ```sh
 kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/kubeval .
-kpt fn run kubeval --network
+kpt fn run kubeval
 ```
 
 ## Expected Results
@@ -21,11 +20,11 @@ kpt fn run kubeval --network
 This should give the following output:
 
 ```sh
-[ERROR] Invalid type. Expected: [integer,null], given: string in object 'v1/ReplicationController//bob' in file example-config.yaml in field spec.replicas
+[ERROR] Invalid type. Expected: [integer,null], given: string in object 'v1/ReplicationController//bob' in file resources.yaml in field spec.replicas
 error: exit status 1
 ```
 
-In the `example-config.yaml` file, replace the value of `spec.replicas`
+In the `resources.yaml` file, replace the value of `spec.replicas`
 with an integer to pass validation and rerun the command. This will return
 success (no output).
 
--- a/examples/kubeval/simple/README.md
+++ b/examples/kubeval/simple/README.md
@@ -20,12 +20,15 @@ kpt fn run kubeval
 This should give the following output:
 
 ```sh
+[ERROR] Additional property templates is not allowed in object 'v1/ReplicationController//bob' in file resources.yaml in field templates
 [ERROR] Invalid type. Expected: [integer,null], given: string in object 'v1/ReplicationController//bob' in file resources.yaml in field spec.replicas
 error: exit status 1
 ```
 
-In the `resources.yaml` file, replace the value of `spec.replicas`
-with an integer to pass validation and rerun the command. This will return
-success (no output).
+There are validation error in the `resources.yaml` file, to fix them:
+- replace the value of `spec.replicas` with an integer
+- change `templates` to `template`
+
+Rerun the command, and it will return success (no output).
 
 [kubeval website]: https://www.kubeval.com/
--- a/functions/ts/kubeval/build/kubeval.Dockerfile
+++ b/functions/ts/kubeval/build/kubeval.Dockerfile
@@ -29,7 +29,7 @@ FROM node:14.15-alpine3.12
 
 RUN apk add --update --no-cache python3 py3-pip && ln -sf python3 /usr/bin/python
 RUN pip install pyyaml jsonref click
-COPY forked/github.com/instrumenta/openapi2jsonschema/*.py /openapi2jsonschema/
+COPY third_party/github.com/instrumenta/openapi2jsonschema/openapi2jsonschema/*.py /openapi2jsonschema/
 RUN chmod +x /openapi2jsonschema/command.py && ln -s /openapi2jsonschema/command.py /usr/bin/openapi2jsonschema
 
 # Run as non-root user as a best-practices:
@@ -42,6 +42,6 @@ COPY --from=builder /home/node/app /home/node/app
 COPY --from=builder /usr/local/bin /usr/local/bin
 COPY openapi.json /home/node/
 
-ENV PATH /usr/local/bin:$PATH
+ENV LOG_TO_STDERR=true
 
 ENTRYPOINT ["node", "/home/node/app/dist/kubeval_run.js"]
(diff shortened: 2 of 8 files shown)


functions/ts/kubeval/build/kubeval.Dockerfile
```diff
@@ -1,6 +1,15 @@
+FROM python:3.9-alpine3.12 as parser
+
+RUN pip install --no-cache-dir openapi2jsonschema
+
+COPY openapi.json /home/
+COPY scripts/preprocess.py /home/
+
+RUN python /home/preprocess.py /home/openapi.json | openapi2jsonschema --kubernetes --expanded --stand-alone -o /home/master-standalone file:///dev/stdin
```
mengqiy, line 8: `preprocess.py` is a workaround for https://github.com/instrumenta/openapi2jsonschema/issues/49.
This step is really time-consuming. It takes 50+ seconds on my mac.

After we use openapi doc injected by kpt, we will need to do this conversion at runtime. I have some concerns about the UX, since it may be executing for 50s without any output.

Shell32-Natsu: Maybe you can try other tools. There are some listed in https://openapi.tools/

examples/kubeval/simple/README.md
```diff
@@ -1,31 +1,30 @@
-# Kubeval
+# kubeval: simple example
```
frankfarzan, line 1: Can you also uncomment the function in `nginx` package and test it works in the quickstart docs:

https://kpt.dev/book/01-getting-started/02-quickstart

Things to test:
- It passes with no change to package
- It fails if user introduced bad changes.

examples/kubeval/simple/README.md
```diff
@@ -1,31 +1,30 @@
...
+kpt fn run kubeval
 ```
 
 ## Expected Results
 
 This should give the following output:
 
 ```sh
-[ERROR] Invalid type. Expected: [integer,null], given: string in object 'v1/ReplicationController//bob' in file example-config.yaml in field spec.replicas
+[ERROR] Invalid type. Expected: [integer,null], given: string in object 'v1/ReplicationController//bob' in file resources.yaml in field spec.replicas
```
frankfarzan, line 23: @Shell32-Natsu @mengqiy Related to the discussion about CLI output and structured results.

functions/ts/kubeval/src/kubeval.ts
```diff
@@ -158,12 +132,6 @@ kubeval.usage = `
 Validates configuration using kubeval.
 
 Configured using a ConfigMap with the following keys, all of which are optional:
-${SCHEMA_LOCATION}: The base URL used to download schemas.  If not specified,
```
frankfarzan, line 161: This change is backwards incompatible. Can we retain these flags but not use them in default declarative flow? In imperative flow, if the user really wants to mount additional schema and use these additional flags, then that's OK. Folks from Spotify have added these, so we should avoid breaking that flow if possible.

mengqiy: The eventual workflow for fn kubeval is like:
1. openapi doc is injected by kpt at some location. (we will bake a openapi doc in the image in the near term)
2. We convert the openapi to json schema and write it to `SCHEMA_LOCATION`.
3. kubeval validates resources against the json schema in `SCHEMA_LOCATION`

However, kpt fn execute functions as `nobody`, if `SCHEMA_LOCATION` is not `/tmp`, we cannot write the json schema files in step 2.

mengqiy: Thinking a bit more about it, it seems what we can do is:
If users specify `SCHEMA_LOCATION`, we always use that.
If `SCHEMA_LOCATION` is not specified, we will use `/tmp` as the default and use the baked-in openapi (or the one injected by kpt in the future).

functions/ts/kubeval/build/kubeval.Dockerfile
mengqiy: instrumenta/openapi2jsonschema is purposely built for kubeval and handle some special cases in k8s. I have spent some time trying to use some of tools in the list. They are need non-trivial amount of work to work with kubeval and k8s.
I'm going to fork `instrumenta/openapi2jsonschema` and fix the issue (instrumenta/openapi2jsonschema#49). And also improve it to write only the files that are needed which can reduce the time for conversion.

mengqiy:
The 1st commit is purely forking instrumenta/openapi2jsonschema code.
The 2nd commit is real changes. It's very different from the code in the previous review round.

I reordered and squashed some commits. Sorry about the force push.

mengqiy:
I'm going to work on the usage doc README.md and verify that https://kpt.dev/book/01-getting-started/02-quickstart is working fine.

functions/ts/kubeval/src/kubeval.ts
frankfarzan: SGTM

functions/ts/kubeval/forked/github.com/instrumenta/openapi2jsonschema/command.py
```diff
@@ -0,0 +1,242 @@
+#!/usr/bin/env python
```
frankfarzan, line 1: Can you explain why we need to run this utility? Can't kubeval already work with the OpenAPI document served by a cluster?

mengqiy: > Can't kubeval already work with the OpenAPI document served by a cluster?

Unfortunately no.
`kubeval` by default fetches json schemas from kubernetesjsonschema.dev which is a static site with its content in https://github.com/instrumenta/kubernetes-json-schema. [Here](https://github.com/instrumenta/kubernetes-json-schema/blob/master/build.sh#L178) is the script to generate new contents. The script is using `openapi2jsonschema`  to do the conversion. https://github.com/instrumenta/openapi2jsonschema is not very active any more, so I fork it and fix a bug and make some improvement to make the conversion faster.

functions/ts/kubeval/forked/github.com/instrumenta/openapi2jsonschema/command.py
```diff
@@ -76,7 +75,7 @@ def default(output, schema, prefix, stand_alone, expanded, kubernetes, strict, a
     info("Parsing schema")
     # Note that JSON is valid YAML, so we can use the YAML parser whether
     # the schema is stored in JSON or YAML
-    data = yaml.load(response.read(), Loader=yaml.SafeLoader)
+    data = json.loads(response.read())
```
mengqiy, line 78: This change makes parsing much faster. It only takes seconds to parse.

examples/kubeval/simple/README.md
mengqiy: kubeval works fine in the quickstart.

functions/ts/kubeval/forked/github.com/instrumenta/openapi2jsonschema/command.py
```diff
@@ -0,0 +1,241 @@
+#!/usr/bin/env python
```
frankfarzan, line 1: we should have third-party code in a `thirdparty` directory. You may want to separate committing the pristine copy of the thirdparty in a separate PR and include information such as git repo/commit.

mengqiy: Created https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/240

mengqiy:
I have to force push to drop an earlier commit which forks the openapi2jsonschema package.
The last commit (05f027d) is the new change since last review.

I have tried to install the our fork using `python -m pip install -e <local project path> `. It turned out it requires much more work than I expected. I will stop trying that unless we really need to do it.

mengqiy:
`functions/ts/kubeval/openapi.json` currently is not pure OSS. It comes from GKE which contains some GCP CRDs (e.g. backendConfig)

examples/kubeval/simple/README.md
frankfarzan: Just make sure these are updated once the output UI changes land.

frankfarzan (approved):
Thanks, this is much improved.

examples/kubeval/simple/README.md
mengqiy: Added a comment in https://github.com/GoogleContainerTools/kpt/issues/1548#issuecomment-825950570 to remind us.

PR #241, opened 2021-04-23
Updating SOPS to 3.7.1

aodinokov:
Sops 3.7.1 introduces a limited AGE support[1]

Adding AGE support to KRM-function
Disabling gpg importing key debug by default (can be overridden),
because it was too verbose and not really useful (SOPS reported the issue anyway in case the key wasn't imported).

[1]
https://github.com/mozilla/sops#encrypting-using-age


examples/contrib/sops/README.md
```diff
@@ -24,23 +26,47 @@ or if your keys are already in `gpg`, it's possibe to run:
...
 Get this example and try it out by running the following commands:
 
 ```sh
 # download sops kpt-function example
 kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/contrib/sops .
-# copy example key from sops project
-curl -fsSL -o sops/key.asc https://raw.githubusercontent.com/mozilla/sops/master/pgp/sops_functional_tests_key.asc
-# run the function
-kpt fn run --env SOPS_IMPORT_PGP="$(cat sops/key.asc)" sops/local-configs
+cd kpt-functions-catalog/examples/contrib/sops
```
mengqiy, line 51: After you run `kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/contrib/sops .`, we will have directory `sops` locally.
So I guess we should run `cd sops` here.

examples/contrib/sops/README.md
```diff
@@ -24,23 +26,47 @@ or if your keys are already in `gpg`, it's possibe to run:
...
+kpt fn run --env SOPS_IMPORT_AGE="$(cat age_keys.txt)" age
 ```
 
 ## Expected result
 
 Verify the updated configuration:
 
 ```sh
-kpt cfg cat sops/local-configs
+kpt cfg cat gpg
```
mengqiy, line 69: Our current model for examples is that each example should have one package and one README.md.
I guess it probably makes more sense to split it into 2 examples.

examples/contrib/sops/README.md
```diff
@@ -24,23 +26,47 @@ or if your keys are already in `gpg`, it's possibe to run:
...
+For AGE case this function introduces and additional `SOPS_IMPORT_AGE` ENV var
+that must contain the [SOPS age keys.txt file](https://github.com/mozilla/sops/blob/master/age/keys.txt).
+This file is needed for decryption. For encryption it's possible to use
+`SOPS_AGE_RECIPIENTS` ENV variable or parameter `age`.
+
+```sh
+kpt fn run --env SOPS_IMPORT_AGE="$(cat <file with age keys>.txt)" <folder>
+```
+
+of if SOPS is already working , it's possible to run:
```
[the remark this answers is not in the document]
aodinokov: Thank you @mengqiy .. this is also misprint s/of/or

Will correct that

examples/contrib/sops/README.md
aodinokov: @mengqiy , thank you for reviewing that!
I've split them, and now there is a following structure

```
examples/contrib/sops/
├── age
│   ├── function.yaml
│   ├── README.md
│   ├── to-decrypt.yaml
│   └── to-encrypt.yaml
├── gpg
│   ├── function.yaml
│   ├── README.md
│   ├── to-decrypt.yaml
│   └── to-encrypt.yaml
└── README.md
```

But now I'm reading your comment again and thinking about a bit different approach:

```
examples/contrib/
├── sops-age
│   ├── function.yaml
│   ├── README.md
│   ├── to-decrypt.yaml
│   └── to-encrypt.yaml
├── sops-gpg
│   ├── function.yaml
│   ├── README.md
│   ├── to-decrypt.yaml
│   └── to-encrypt.yaml
```

The only concern with that - no place to store some common info about the module. And I think - duplicating it into both README's isn't a good idea as well.

One more option is to combine 2 examples into 1, but this may be too complex.

Please let me know what do you think about the current option. I'm ready to modify it based on you suggestions  :)

mengqiy (approved):

aodinokov:
@mengqiy , thank you very much for reviewing and merging that!

Could you please when you have time create the next release for sops function? Since the previous was 0.1.0 and the new functionality like support of AGE was added, but all parameters still compatible with 0.1.0, I think maybe the tag 0.2.0 will be appropriate.

Cheers,
Alexey

mengqiy:
Sound good. Will do.

PR #246, opened 2021-04-26
add check to verify docs

mengqiy:
Part of https://github.com/GoogleContainerTools/kpt/issues/1669


mengqiy:
@frankfarzan @Shell32-Natsu PTAL

Makefile
```diff
@@ -40,6 +40,9 @@ check-licenses:
 	cd functions/go && $(MAKE) check-licenses
 	cd functions/contrib/ts && $(MAKE) check-licenses
 
+check-metadata:
+	scripts/check-metadata.py
```
frankfarzan, line 44: "metadata" by itself is too generic, quality what metadata this refers in the target name and scriptname

.github/workflows/ci.yaml
```diff
@@ -67,8 +67,24 @@ jobs:
         cd functions/go
         make check-licenses
 
+  verify-ci:
+    runs-on: ubuntu-latest
+    env:
+      GOPATH: /home/runner/work/kpt-functions-catalog/functions/go
```
frankfarzan, line 73: why do you need GOPATH if it's running python?

I don't feel strongly about this and don't need to take action now: but we should standardizing on one scripting language: currently, we have shell, Go, python, node. It causes and infra complexity.

mengqiy:
Rebased on top of #247.
It now also checks the `examples/` directory.

.github/workflows/ci.yaml
mengqiy: The script now runs `mdrip` which is written in go.

mengqiy: I want to get rid of shell in our scripts, since they will become less readable over time.

I can't recall if we have any node scripts.

mengqiy:
@frankfarzan PTAL

frankfarzan (approved):

PR #251, opened 2021-04-28
Rename functions

mengqiy:
gatekeeper-validate -> enforce-gatekeeper
set-annotation -> set-annotations
set-label -> set-labels
helm-inflator -> inflate-helm-chart
istioctl-analyze -> analyze-istio

Mostly mechanical changes.
Run `go mod tidy` to trim `go.sum` files which results in some deletion.


frankfarzan (approved):
Going just by the PR description, I'll let others look at actual changes.

PR #254, opened 2021-04-28
use GITHUB_BASE_REF env var

mengqiy:
GITHUB_REF is not working as I expected. I use GITHUB_BASE_REF now.
Ref: https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables


mengqiy:
The CI is passing in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/253 after cherry-picking the commit in this PR.

frankfarzan (approved):

PR #269, opened 2021-05-04
Setup firebase hosting

mortent:
Firebase hosting for site


.github/workflows/firebase-hosting-merge.yml
```diff
@@ -0,0 +1,25 @@
+# This file was auto-generated by the Firebase CLI
+# https://github.com/firebase/firebase-tools
+
+name: Deploy to Firebase Hosting on merge
+'on':
+  push:
+    branches:
+      - master
```
mengqiy, line 8: We should not only trigger the CD on master branch, but also on other release branches.

We can address this in a followup PR

mengqiy (approved):

.github/workflows/firebase-hosting-merge.yml
etefera: Addressed in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/270

PR #270, opened 2021-05-04
Update firebase hosting deployment.

etefera:
Addresses https://github.com/GoogleContainerTools/kpt/issues/1489


.github/workflows/firebase-hosting-merge.yml
```diff
@@ -15,11 +12,12 @@ jobs:
           fetch-depth: 0
       - name: Generate Site
         run: make site-generate
-      - uses: FirebaseExtended/action-hosting-deploy@v0
+      - uses: FirebaseExtended/action-hosting-deploy@276388dd6c2cde23455b30293105cc866c22282d # v0
```
mengqiy, line 15: Why using hash instead of v0?

etefera: Rationale at https://github.com/GoogleContainerTools/kpt/pull/1515#discussion_r586901305

.github/workflows/firebase-hosting-merge.yml
```diff
@@ -1,11 +1,8 @@
-# This file was auto-generated by the Firebase CLI
+# This file was initially auto-generated by the Firebase CLI
 # https://github.com/firebase/firebase-tools
 
 name: Deploy to Firebase Hosting on merge
-'on':
-  push:
-    branches:
-      - master
+on: push
```
mengqiy, line 5: Given that the site only needs the information from the released branches, this can be optimized to avoid re-deploy when PRs merge to master.

mengqiy:
I want to confirm something. If any step fails in the generation and deployment, the site continue to serve the previous good site, right?

etefera:
> I want to confirm something. If any step fails in the generation and deployment, the site continue to serve the previous good site, right?

That's correct, the previous version of the site stays live if the deployment (or site generation) fails

.github/workflows/firebase-hosting-merge.yml
mengqiy: FirebaseExtended is not authored by the team in Google?

.github/workflows/firebase-hosting-merge.yml
etefera: If there are changes to the site code or the generator, which lives in the main branch, we would want the deployment to pick it up.

.github/workflows/firebase-hosting-merge.yml
etefera: According to https://github.com/FirebaseExtended, it's "not officially staffed by Google"

mengqiy (approved):

.github/workflows/firebase-hosting-merge.yml
mengqiy, line 5: Run this workflow on every PR cause some noise. e.g. https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/288 The CI is actually fine except this workflow is failing. 
Ideally, we should filter by branches like:
```
  push:
    branches:
      - master
      - [-\w]*\/v\d*\.\d*  # I assume GHA supports filtering branches by regex.
```

mengqiy: @etefera

etefera: Addressed in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/292

mengqiy: Thanks for the quick response!

PR #301, opened 2021-05-12
Use image paths as names for landing page.

etefera:
image

### DIFF
--- a/scripts/generate_catalog/generate_catalog.go
+++ b/scripts/generate_catalog/generate_catalog.go
@@ -73,6 +73,7 @@ func main() {
 
 type function struct {
 	FunctionName      string
+	ImagePath         string
 	VersionToExamples map[string]map[string]example
 	LatestVersion     string
 	Path              string
@@ -274,7 +275,8 @@ func parseMetadata(f function, md metadata, version string, versionDest string) 
 	f.Path = versionDest
 	f.Description = md.Description
 	sort.Strings(md.Tags)
-	f.Tags = strings.Join(md.Tags, ",")
+	f.Tags = strings.Join(md.Tags, ", ")
+	f.ImagePath = md.Image
 
 	return f
 }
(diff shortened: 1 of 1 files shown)


mengqiy (approved):
Thanks!

runewake2:
Recommend also including the landing pages for the functions. Does this do that?

etefera:
> Recommend also including the landing pages for the functions. Does this do that?
> 
> Ex. there are no image paths on the children pages: https://catalog.kpt.dev/starlark/v0.1/

No, that's handled at the `README.md` level for each function.

frankfarzan:
I think this reduces the readability as all functions now have common prefix (`gcr.io/kpt-fn`). Human eye mostly notices the first few characters of words. I suggest, showing the container image in a different context, for example in the usage docs.

mengqiy:
What about adding a new column called `image` in the landing page?

frankfarzan:
Is there a particular reason, it can't be in the function use doc somewhere here: https://screenshot.googleplex.com/5iSDN7PE3BfBwUD

mengqiy:
I think it is possible.
@etefera Can you please confirm? And also how much work is it?

etefera:
We can automatically stick it somewhere there, but I'm not sure why that would be preferred in this case over having the images written into the `README.md`s by the titles.

mengqiy:
> the images written into the README.mds by the titles.

You mean something like:
```
# apply-setters
image: `gcr.io/kpt-fn/apply-setters`

### Overview
...
```

I don't have strong opinion here. It should be OK as long as it looks reasonable good.

etefera:
> > the images written into the README.mds by the titles.
> 
> @etefera You mean something like:
> 
> ```
> # apply-setters
> image: `gcr.io/kpt-fn/apply-setters`
> 
> ### Overview
> ...
> ```
> 
> I don't have strong opinion here. It should be OK as long as it looks reasonable good.

Yeah, something like that. I think it's probably better to handle at the function `README` level.

PR #305, opened 2021-05-13
make file scheme more visible in kubeval help text

mengqiy:


functions/ts/kubeval/README.md
```diff
@@ -33,8 +33,8 @@ kind: ConfigMap
 metadata:
   name: my-func-config
 data:
-  schema_location: "https://kubernetesjsonschema.dev"
-  additional_schema_locations: "https://example.com,file:///abs/path/to/your/schema/directory"
+  schema_location: "file:///abs/path/to/your/schema/directory"
```
frankfarzan, line 36: so schema_location cannot be a URL?

frankfarzan: This doesn't match line 14 above

mengqiy: > - `schema_location`: The base URL used to fetch the json schemas. The default
  is empty. This feature only works with imperative runs, since declarative runs
  allow neither network access nor volume mount.
> - `additional_schema_locations`: List of secondary base URLs used to fetch the
  json schemas.  These URLs will be used if the URL specified by
  `schema_location` did not have the required schema.  The default is empty.
  This feature only works with imperative runs.

It looks like I was confused by URL and URI. They are not interchangeable.
It supports `file` scheme. So I guess we should replace "URL" with "URI".

frankfarzan (approved):

PR #316, opened 2021-05-18
Document semver, backward compatibility and best practices

mengqiy:
Fixes: https://github.com/GoogleContainerTools/kpt/issues/1901

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -22,26 +22,26 @@ Note: We do NOT support the `latest` tag, since we cannot provide any
 compatibility guarantee for it, and the pipeline won’t produce deterministic
 results.
 
-## Interfaces
+## User-facing Surfaces
 
 All functions in the catalog comply with the [function spec]. To learn more
 about the functions concept,
 see [here](http://kpt.dev/book/02-concepts/03-functions).
 
-There are 2 interfaces for a function:
+There are 2 user-facing surfaces for a function:
 
 - The `functionConfig`
 - The function itself
 
-### functionConfig as an interface
+### functionConfig Surface
 
 A `functionConfig` can be either a core resource (e.g. `ConfigMap`) or a custom
 resource. If the `functionConfig` is a CRD, it can be versioned independently as
 a normal CRD.
 
-### Function as an interface
+### Function Surface
 
-#### What are NOT part of the function interface
+#### What are NOT part of the function surface
 
 - The formatting of serialization for output items and results. e.g. yaml
   indentation and order of fields in a map.
@@ -49,7 +49,7 @@ a normal CRD.
 - The order of result items in the results.
 - The content of the unstructured messages in the results.
 
-#### What are part of the function interface
+#### What are part of the function surface
 
 - The supported fields in the ConfigMap as the `functionConfig`.
 - The supported versions of the CRD `functionConfigs`.
@@ -63,7 +63,7 @@ a normal CRD.
 
 We define a breaking change as: For any given input (including `input items`,
 `functionConfig` and OpenAPI), the function produces a different output
-(including output items, results) that are part of the interface.
+(including output items, results) that are part of the user-facing surface.
 
 ### Backwards Compatibility
 
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -31,7 +31,7 @@ see [here](http://kpt.dev/book/02-concepts/03-functions).
 There are 2 user-facing surfaces for a function:
 
 - The `functionConfig`
-- The function itself
+- The function behavior
 
 ### functionConfig Surface
 
@@ -39,9 +39,9 @@ A `functionConfig` can be either a core resource (e.g. `ConfigMap`) or a custom
 resource. If the `functionConfig` is a CRD, it can be versioned independently as
 a normal CRD.
 
-### Function Surface
+### Function Behavior Surface
 
-#### What are NOT part of the function surface
+#### What are NOT part of the function behavior surface
 
 - The formatting of serialization for output items and results. e.g. yaml
   indentation and order of fields in a map.
@@ -49,15 +49,26 @@ a normal CRD.
 - The order of result items in the results.
 - The content of the unstructured messages in the results.
 
-#### What are part of the function surface
+#### What are part of the function behavior surface
 
-- The supported fields in the ConfigMap as the `functionConfig`.
-- The supported versions of the CRD `functionConfigs`.
+- The supported `functionConfig`:
+    - If the function supports `ConfigMap` as `functionConfig`, the supported
+      fields in the `ConfigMap`.
+    - If the function supports a custom resource as `functionConfig`, the
+      supported versions of the custom resource.
 - How the function behave given the input items and `functionConfig`:
-    - The reminder aspects of the output items that are not mentioned in the
+    - The remaining aspects of the output items that are not mentioned in the
       previous section.
-    - The reminder aspects of the results that are not mentioned in the previous
-      section.
+    - The remaining aspects of the results that are not mentioned in the
+      previous section.
+
+For example, if the `kubeval` function stops supporting
+the `ignore_missing_schemas` option in the `ConfigMap`, it will be a breaking
+change.
+
+Another example, if the `set-namespace` function stops supporting custom
+resource of apiVersion `fn.kpt.dev/v1alpha1` and kind `SetNamespace`, it will be
+a breaking change.
 
 ### Breaking Changes
 
@@ -67,7 +78,7 @@ We define a breaking change as: For any given input (including `input items`,
 
 ### Backwards Compatibility
 
-For versions after v1.0.0, we will:
+For post v1.0.0 versions, we will:
 
 - Bump major version: There are breaking changes.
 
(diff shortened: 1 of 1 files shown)


mengqiy:
If you think this doc should live in different directory or repo, please let me know.

VERSIONING.md
```diff
@@ -0,0 +1,93 @@
...
+Note: We do NOT support the `latest` tag, since we cannot provide any
+compatibility guarantee for it, and the pipeline won’t produce deterministic
+results.
+
+## API
+
+All functions in the catalog comply with the [function spec]. A function can be
+represented as the following:
+
+![function representation](https://kpt.dev/static/images/func.svg)
```
frankfarzan, line 28: Just link to kpt.dev
e.g. http://kpt.dev/book/02-concepts/03-functions

VERSIONING.md
```diff
@@ -0,0 +1,93 @@
...
+
+There are 2 API surfaces for a function:
+- The CRD function config API
```
frankfarzan, line 31: `functionConfig`

VERSIONING.md
```diff
@@ -0,0 +1,93 @@
...
+- The behavior of a function as API
```
frankfarzan, line 32: I'm not a fan of the use of API in this doc.

VERSIONING.md
frankfarzan, line 31: Some of these functions take other KRM resources such as `ConfigMap`

VERSIONING.md
```diff
@@ -0,0 +1,93 @@
+# Versioning
+
+## SemVer and shorter SemVer
+
+We use [semantic versioning] for all of our images. We also support a shorter
+version of semantic versioning. E.g. `v<major> and v<major>.<minor>`.
```
frankfarzan, line 6: worth mentioning that fully specified X.Y.Z image is immutable and will never be changed.

VERSIONING.md
```diff
@@ -0,0 +1,94 @@
+# Versioning
+
+## SemVer and shorter SemVer
+
+We use [semantic versioning] for all of our images. The images with fully
+specified version (e.g. `vX.Y.Z`) are immutable and will never be changed. We
+also support a shorter version of semantic versioning. E.g.
+`v<major> and v<major>.<minor>`.
```
frankfarzan, line 8: "and " shouldn't be verbatim

VERSIONING.md
```diff
@@ -0,0 +1,94 @@
...
+
+The shorter version of semantic versioning is going to be "moving tags":
```
frankfarzan, line 10: I've usually heard "floating tags". Doesn't need quotation.

VERSIONING.md
```diff
@@ -0,0 +1,94 @@
...
+
+- `v<major>.<minor>` always points to `v<major>.<minor>.X` where X is the latest
+  (largest) patch version number. E.g. assume `v1.2` points `v1.2.0` initially.
+  After `v1.2.1` is released, `v1.2` now points to `v1.2.1`.
+  
+- `v<major>` always points to `v<major>.X` where X is the latest (largest) minor
```
frankfarzan, line 16: Does the kpt migration doc point out the changes to the image pull behavior?

frankfarzan: @phanimarupaka

VERSIONING.md
```diff
@@ -0,0 +1,94 @@
...
+results.
+
+## Interfaces
+
+All functions in the catalog comply with the [function spec]. To learn more
+about the functions concept, see [here](http://kpt.dev/book/02-concepts/03-functions).
+
+There are 2 interfaces for a function:
+- The `functionConfig`
+- The function itself
```
frankfarzan, line 32: How's "the function" an interface. Rephrase to be more precise.

VERSIONING.md
```diff
@@ -0,0 +1,94 @@
...
+
+### functionConfig as an interface
+
+A `functionConfig` can be either a CRD or other KRM resource (e.g. `ConfigMap`).
```
frankfarzan, line 36: core or CRD

VERSIONING.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Breaking Changes
+
+We define a breaking change as: For any given input (including `input items`,
+`functionConfig` and OpenAPI), the function produces a different output
+(including output items, results) that are part of the interface.
+
+### Backwards Compatibility
+
+If there are breaking changes, we bump the major version.
```
frankfarzan, line 68: Order these (using bullet points):
- Bumping major version
- Bumping minor version
- Bumping path version

VERSIONING.md
mengqiy: Now we avoiding using "API" and "interface" here. I can't come up with a good word here. Can you please suggest the wording here?

mengqiy: @frankfarzan @droot Any suggested wording here?

mengqiy:
@frankfarzan @droot PTAL

VERSIONING.md
```diff
@@ -0,0 +1,108 @@
...
+
+- The formatting of serialization for output items and results. e.g. yaml
+  indentation and order of fields in a map.
+- The order of resources in the output items.
+- The order of result items in the results.
+- The content of the unstructured messages in the results.
+
+#### What are part of the function surface
+
+- The supported fields in the ConfigMap as the `functionConfig`.
```
frankfarzan, line 54: this assumes all functions take ConfigMap which is not the case. I would generalize this with the next point

VERSIONING.md
```diff
@@ -0,0 +1,108 @@
...
+- Bumping patch version: In all other cases, including backward-compatible
+  features, bug fixes and security fixes.
+
+Users won’t observe breaking changes if they are using the shorter semantic
+versions (e.g. `v1.2` and `v1`) and they can automatically get the latest secure
+patch version for free.
+
+## Best Practices
+
+- It is recommended to use shorter semantic versions (e.g. `v1.2` and `v1`) in
```
frankfarzan, line 93: I think we may want to talk about the benefits of each scheme:

Fully specified:
- Immutable infrastructure
- You are in control of when to update the functions

Floating
- Less maintaince
- Automatically pick up security and other fixes

Then let the reader decide for themselves.

VERSIONING.md
```diff
@@ -0,0 +1,108 @@
+# Versioning
```
frankfarzan, line 1: As a follow up, it's good to link to this from https://catalog.kpt.dev for discoverability.

frankfarzan: Actually, the site only shows vX.Y. What's the plan for showing other tags? 

https://screenshot.googleplex.com/amjaGmi5yvJEvCn

Here's is a good place to to have a tool-tip to this doc in context.

VERSIONING.md
droot: We can use `Function behavior` term if it makes more sense.

VERSIONING.md
```diff
@@ -0,0 +1,108 @@
...
+- The content of the unstructured messages in the results.
+
+#### What are part of the function surface
+
+- The supported fields in the ConfigMap as the `functionConfig`.
+- The supported versions of the CRD `functionConfigs`.
+- How the function behave given the input items and `functionConfig`:
+    - The reminder aspects of the output items that are not mentioned in the
+      previous section.
+    - The reminder aspects of the results that are not mentioned in the previous
```
droot, line 59: s/reminder/remaining ?

VERSIONING.md
```diff
@@ -0,0 +1,108 @@
...
+      section.
```
droot, line 60: We can share an example to help user grok this better.

`For example, if set-namespace function stops supporting particular kinds of resources from a particular version, that will be a breaking change in the behavior of the function`

VERSIONING.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Breaking Changes
+
+We define a breaking change as: For any given input (including `input items`,
+`functionConfig` and OpenAPI), the function produces a different output
+(including output items, results) that are part of the user-facing surface.
+
+### Backwards Compatibility
+
+For versions after v1.0.0, we will:
```
droot, line 70: I think adding a note that describe the semver semantics for pre 1.0 versions will be useful because all of our function's versions are currently pre-1.0.

VERSIONING.md
```diff
@@ -0,0 +1,108 @@
...
+For versions before v1.0.0, the major version is always `0` and we will:
+
+- Bumping minor version: There are breaking changes.
+
+- Bumping patch version: In all other cases, including backward-compatible
+  features, bug fixes and security fixes.
+
+Users won’t observe breaking changes if they are using the shorter semantic
+versions (e.g. `v1.2` and `v1`) and they can automatically get the latest secure
+patch version for free.
```
droot, line 89: `for free` part is not relevant.

VERSIONING.md
droot: This is the most important bit for the function consumers, so listing what Frank listed above is a good starting point.

droot (changes requested):

VERSIONING.md
frankfarzan: To make this more explicit, I think we should remove "recommend to use shorter versions". It's a trade-offs between different types of risk:
- Risk of not having security patches picked up automatically
- Risk of being broken because we either accidentally pushed a breaking change or the consumer is dependent on a feature which is not part of the contract

mengqiy:
@droot PTAL

droot (approved):

PR #332, opened 2021-05-27
set imagePullPolicy to ifNotPresent in e2e tests

mengqiy:
This is to ensure we are testing the newly built images


mengqiy:
No. We are function authors. In our CI in the catalog repo, we first build our new function images, we want the e2e test to be run against the local images (because they contains the changes we want to test) instead of using the images from the remote registry which don't have the new changes.

mengqiy:
Option `always` is for function consumers.

frankfarzan:
> No. We are function authors. In our CI in the catalog repo, we first build our new function images, we want the e2e test to be run against the local images (because they contains the changes we want to test) instead of using the images from the remote registry which don't have the new changes.

In that case, don't you want `never` to use the images built locally?

frankfarzan (approved):

PR #342, opened 2021-06-01
update usage doc for gatekeeper and starlark

mengqiy:
Also split constraint template and constraint from the resources to be validated.


mengqiy:
Rebased on top of the Gatekeeper fn renaming PR.

The last commit contains the new changes since last review.

tests/gatekeeper/valid/policy.yaml
```diff
@@ -32,7 +32,6 @@ kind: K8sBannedConfigMapKeysV1
 metadata:
   name: no-secrets-in-configmap
 spec:
-  enforcementAction: dryrun
```
frankfarzan: why is htis removed here and added above?

tests/gatekeeper/valid/to-validate.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
frankfarzan, line 1: why not just name this config-map.md?

tests/gatekeeper/valid/policy.yaml
mengqiy: This is not true. GitHub diff appears to be not smart enough.
What happened is:
- tests/gatekeeper/valid/resources.yaml => tests/gatekeeper/valid/policy.yaml
- tests/gatekeeper/dryrun/resources.yaml => tests/gatekeeper/dryrun/policy.yaml

But GH diff thought it's:
- tests/gatekeeper/valid/resources.yaml => tests/gatekeeper/dryrun/policy.yaml
- tests/gatekeeper/dryrun/resources.yaml => tests/gatekeeper/valid/policy.yaml

frankfarzan (approved):

functions/go/starlark/README.md
```diff
@@ -4,15 +4,20 @@
...
 
 <!--mdtogo-->
 
-### Synopsis
+### FunctionConfig
 
 <!--mdtogo:Long-->
 
-You can specify your Starlark script inline under field source like this:
+[Starlark] is a dialect of Python. It is commonly used as a configuration language. It is an untyped dynamic language with
```
frankfarzan, line 15: This is part of the overview, not Function Config.

In general, this usage doc is very spartan and needs to be expanded.

PR #369, opened 2021-06-09
improve usage doc for gatekeeper and kubeval

mengqiy:
This PR improve usage doc for our validator functions.


functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,87 @@
...
+[Gatekeeper] provides an extensible policy enforcement engine which allows users
+to validate the KRM resources against the Gatekeeper policies.
+
+There are 2 kinds of resources needed to define a policy:
+- [Constraint Template]: Define the schema and logic of a policy. The policy
+  logic in a Constraint Template must be written in the [Rego] language.
+- [Constraint]: Signal the Gatekeeper the corresponding constraints need
+  to be enforced. Every Constraint must be backed by a Constraint Template.
+
+To make it do something meaningful, you should to define at least one
```
frankfarzan, line 19: this is describing a corner case, not sure why it's in the overview. I suggest removing it.

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,87 @@
...
+[Constraint Template] and at least one [Constraint]. If you don't provide any
+[Constraint Template] or [Constraint], it's a NOOP for the function.
+You can learn more about how to use the `Gatekeeper` project [here][howto]. The
```
frankfarzan, line 22: different paragraph

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,87 @@
 
 <!--mdtogo:Short-->
 
-Validate the KRM resources using [Gatekeeper] policies.
+Validate the KRM resources using [Gatekeeper] policies. The `gatekeeper`
+function wraps the `Gatekeeper` policy enforcement engine.
+
+[Gatekeeper] provides an extensible policy enforcement engine which allows users
```
frankfarzan, line 10: "Gatekeeper" by itself is overloaded. Do you mean the project?

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,87 @@
 
 <!--mdtogo:Short-->
 
-Validate the KRM resources using [Gatekeeper] policies.
+Validate the KRM resources using [Gatekeeper] policies. The `gatekeeper`
+function wraps the `Gatekeeper` policy enforcement engine.
```
frankfarzan, line 8: Second second is extraneous.

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,87 @@
...
 <!--mdtogo-->
 
+<!--mdtogo:Long-->
+
 ### FunctionConfig
 
-<!--mdtogo:Long-->
+The constraint templates and the constraints must be provided using
+`input items` along with other KRM resources to be validated. Nothing need to be
+provided in `input functionConfig`.
```
frankfarzan, line 33: Not sure what `input functionConfig` is.

Showing the CRD in the section called `FunctionConfig` is weird as they's no FunctionConfig. 

The biggest issue is that there are holes in story here, what the reader is looking for from this doc:

- What's this function about?
- What problem does it solve?
- Why should I use it?
- How should I use it?

This PR doesn't actually expand on any of this.

functions/go/gatekeeper/generated/docs.go
```diff
@@ -3,17 +3,73 @@
 // Code generated by "mdtogo"; DO NOT EDIT.
 package generated
 
-var GatekeeperShort = `Validate the KRM resources using [Gatekeeper] policies.`
+var GatekeeperShort = `Validate the KRM resources using [Gatekeeper] policies. The `gatekeeper`
```
mengqiy, line 6: @phanimarupaka Any idea "`" is not escaped? I'm expecting
```
`Validate the KRM resources using [Gatekeeper] policies. The` + "`" + `gatekeeper` + "`" + `...`
```

functions/go/gatekeeper/README.md
droot: I am looking at the function-catalog docs (especially functions that wraps other tools) and realizing that there are two types of audience for such docs:

- Users who don't know about the tools being wrapped. This audience will actually start with a higher level problem such as "I want to ensure some form of extra validation (constraints, policies....) for my configs", so they need something like I would call a `solution page`.
- Users who know what the wrapped tool does. They also know how `kpt` functions work etc. So all they need is a quick reference docs and some examples. Solution page will also refer to these reference page.

So question I am trying to answer in my head is: what does this page mean ? is it `reference page` or `solution page`. It seems it is trying to do both the things and that can get confusing for the author of the docs as well as audience of the docs.

mengqiy:
I have to rebase to pick the fix for the CI.
The last 2 commits contain the new changes.

functions/go/gatekeeper/README.md
mengqiy: This is a higher level discussion then.
I think we can iterate on the docs.
The revision in this PR is better than what we currently have. I guess we can merge this PR soon-ish and then figure out if we want to have reference page and solution page. And how.

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,93 @@
 
 <!--mdtogo:Short-->
 
-Validate the KRM resources using [Gatekeeper] policies.
+The `gatekeeper` function wraps the [`Gatekeeper`] policy enforcement engine to
+evaluate [Gatekeeper] constraints to validate the KRM resources.
```
droot, line 8: Another wording (heavily inspired by the OPA blogpost https://kubernetes.io/blog/2019/08/06/opa-gatekeeper-policy-and-governance-for-kubernetes/) itself:

The `gatekeeper` function enforces policies on the package resources. You can define policies for governance and legal requirements or to enforce best practices and organizational conventions.
 
For example, you can enforce policies like:
- `ConfigMap` must not contain fields with `private_key` name
- All pods must have resource limits
- All namespaces must have a label that lists a point-of-contact

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,93 @@
...
 
 <!--mdtogo-->
 
-### FunctionConfig
+It provides a very flexible and powerful way to validate KRM resources against
+custom policies.
```
droot, line 13: This line doesn't help much.

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,93 @@
...
+
+You can learn more about how to use the `Gatekeeper` project [here][howto]. The
+concept document about the `OPA Constraint Framework` can be found
+[here][concept].
```
droot, line 17: An alternate wording for second sentence:

The policies are expressed using `OPA Constraint Framework`, you can read more about it [here](...).

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,93 @@
...
 
 <!--mdtogo:Long-->
 
-[Gatekeeper] allows users to validate the KRM resources against the Gatekeeper
-policies.
+### Authoring Policies
+
+The `gatekeeper` function follows the [executable configuration] pattern. There
+are 2 kinds of resources needed to define a policy and they need to be provided
+using `input items` along with other KRM resources to be validated.
```
droot, line 25: `input items` feels like low level detail to me. Can we just say: "and they need to be part of the input resources" ?

functions/go/gatekeeper/README.md
```diff
@@ -4,29 +4,93 @@
...
+
+- [Constraint Template]: Define the schema and logic of a policy. The policy
+  logic in a Constraint Template must be written in the [Rego] language.
+- [Constraint]: Signal the Gatekeeper the corresponding constraints need to be
+  enforced. Every Constraint must be backed by a Constraint Template.
+
+The constraint templates and the constraints must be provided using
+`input items` along with other KRM resources to be validated. Nothing need to be
```
droot, line 33: I am wondering if we should just say that: "The constraint templates and the constraints resources should be in the package containing the KRM resources."

I think we can remove the sentence about `input functionConfig`.

functions/go/gatekeeper/README.md
droot: Agree. I think we can have a separate `tutorial.md` for each function and have two links on the function homepage (shown below). I am treating this page more as a reference page for now.

[tutorial](..)|[reference](...)

droot (changes requested):

mengqiy:
@droot PTAL

droot (approved):

PR #374, opened 2021-06-11
use kpt from main branch

mengqiy:


frankfarzan (approved):

droot (approved):

mengqiy:
Change it to install from `main` explicitly. Otherwise it still installs the pre-v1 kpt.

frankfarzan:
> Change it to install from `main` explicitly. Otherwise it still installs the pre-v1 kpt.

This shouldn't be the case. Are you sure?

frankfarzan:
> > Change it to install from `main` explicitly. Otherwise it still installs the pre-v1 kpt.
> 
> This shouldn't be the case. Are you sure?

You're right. We should change this as these instructions are incorrect now:

https://kpt.dev/installation/?id=source

mengqiy:
>You're right. We should change this as these instructions are incorrect now:

I saw you have already updated it.

I guess if we rename `next` to `master` instead of `main`, it should work and we don't need to specify `@main`.

PR #386, opened 2021-06-16
Make catalog generation script more generic.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/2250
Avoids assuming things about the directories between "`examples`"  and `exampleName`


mengqiy:
IIUC the code in this PR can handle both the old example layout (`examples/<fn-name>/<example-name>`) and the new example layout (`examples/<fn-name>-<example-name>`), since it now uses `examplePackageURLs` from `metadata.yaml`.

mengqiy (approved):

PR #395, opened 2021-06-17
Add Examples to bottom of every function page.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/2268
image
image


site/index.html
```diff
@@ -128,7 +134,43 @@
...
+                  exampleNames.forEach((ex, _) => {
+                    if (currentEx != ex) {
+                      exampleItems += `<li><a href="${examples[ex].LocalExamplePath}/">
+                      ${ex}</a></li>`;
+                    }
+                  });
+                  let examplesInlineListSection = exampleItems
+                    ? `
+                  <h3 id="examples">${
+                    currentEx ? "Other Examples" : "Examples"
```
frankfarzan, line 167: "More Examples" might be better

droot:
Not scoped to this PR, but I realized, we need to come up with better labels for these examples. For ex: `suffix` --> `Add suffix in resource names` something more descriptive. @mengqiy

site/index.html
etefera: Done

mengqiy:
> we need to come up with better labels for these examples. For ex: suffix --> Add suffix in resource names something more descriptive.

There are 2 things here:
- The name of the example that will show up in the site (we want it to be descriptive)
- Package path that can be fetched using `kpt pkg get` (we want it to be short)

Eyob and I had some discussion recently about changing the `metadata.yaml` file. Currently we have `examples` which is a string list. We can change it to be a map from example name on site to the example URL. Then we can make name descriptive while the URL short.

mengqiy (approved):

PR #396, opened 2021-06-17
rename functionConfig for set-namespace and improve usage doc

mengqiy:


examples/set-namespace/advanced/README.md
```diff
@@ -41,7 +41,7 @@ configuration.
 
 Suppose we have a CRD with group `dev.example.com`, version `v1` and kind
 `MyResource`. We want the namespace to be set in field
```
droot, line 43: The above line `we embed the CRD in the kptfile` in this example...` seems inaccurate.

examples/set-namespace/advanced/README.md
```diff
@@ -41,7 +41,7 @@ configuration.
...
-`.spec.configmapRef.namespace` as well. We specify it in field `fieldSpecs`.
+`.spec.configmapRef.namespace` as well. We specify it in field `additionalNamespaceFields`.
```
droot, line 45: We can introduce `additionalNamespaceFields` with:

`If you have a resources that uses `namespace` in fields other than `metadata.namespace`, you can use `additionalNamespaceFields` to update those namespace fields. WDYT ?

droot (approved):
A minor change in the language. Ship it after fixing that.

PR #431, opened 2021-06-23
only install mdtogo for golang

mengqiy:


scripts/function-release.sh
```diff
@@ -39,7 +39,9 @@ fn_ver=$(parse_git_tag version)
 
 if [ -d "${scripts_dir}/../functions/${fn_lang}/${fn_name}" ]; then
   cd "${scripts_dir}/../functions/${fn_lang}"
-  make install-mdtogo
+  if [ "${fn_lang}" == "go" ]; then
+    make install-mdtogo
```
frankfarzan, line 43: This is weird, what's the call chain here:

GH -> make -> bash -> make ?

mengqiy: > GH workflow -> make -> bash -> make ?

That's not the call chain. This script is only used by `louhi` to build and push images with semver. 
The actual call chain is `bash -> make`.

> GH workflow

Presubmit CI and site deployment

> Makefile

A bunch of target that build and test all functions or a specific function

> Bash script

All other stuffs.

droot (approved):

PR #435, opened 2021-06-25
update apk in docker

mengqiy:


droot (approved):

droot:
we should do this for all the functions ?

mengqiy:
> we should do this for all the functions ?

Only alpine with node images are affected. Golang images (pure alpine) are not affected.

PR #445, opened 2021-06-29
updte doc and add one example

mengqiy:
- Update usage docs
- Update function short description
- Add one example for gatekeeper

Ref: https://github.com/GoogleContainerTools/kpt/issues/2343 and https://github.com/GoogleContainerTools/kpt/issues/1890


examples/gatekeeper-no-root/README.md
```diff
@@ -0,0 +1,112 @@
+# gatekeeper: No Root
+
+### Overview
+
+This example demonstrates how to declaratively run the [gatekeeper]
```
droot, line 5: This is a very useful example, we should having a brief description of the scenario here. Something along the lines of:

This example demonstrates how to run gatekeeper function declaratively to enforce the policy `Containers must not run as root` on resources.

examples/gatekeeper-no-root/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Fetch the example package
+
+Get the example package by running the following commands:
+
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/gatekeeper-no-root
+```
+
+There are 3 resources: a `ConstraintTemplate`, a `NoRoot` and a `Deployment`.
```
droot, line 16: I think using `DisallowRoot` wording instead of `NoRoot` might reflect the usage better, WDYT ?

functions/ts/kubeval/README.md
```diff
@@ -1,10 +1,24 @@
...
 
-### Overview
+## Overview
 
 The `kubeval` function wraps the [`kubeval`] binary to validate KRM resources
 against their [json schemas].
 
+This function is often used in the following scenarios:
+
+- Validating KRM resources as part of the local development workflow.
```
droot, line 10: I wonder if we have to use `KRM` term ? I think calling `resources` should be enough, right ?

functions/go/starlark/README.md
```diff
@@ -1,18 +1,25 @@
...
 
-### Overview
+## Overview
 
 <!--mdtogo:Short-->
 
 The `starlark` function contains a Starlark interpreter to run a Starlark script
 to mutate or validate resources.
 
+The starlark script is so flexible to do many things such as:
```
droot, line 10: I think we can simply say: `The starlark script can be used to:`

functions/go/starlark/README.md
```diff
@@ -1,18 +1,25 @@
...
+
+- Set namespace for all KRM resources.
```
droot, line 12: Let's avoid use-cases that we have well defined functions for. May be list one that we recently ran into:

`Add an annotation on the basis of a condition`

droot (changes requested):

droot (approved):

PR #457, opened 2021-06-30
Support ConfigMap for starlark and additional parameters

mengqiy:


mengqiy:
cc: @yhrn You may want to take a look at the example.

functions/go/starlark/go.mod
```diff
@@ -4,5 +4,8 @@ go 1.16
 
 require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
-	sigs.k8s.io/kustomize/kyaml v0.10.21
+	k8s.io/api v0.21.0
+	k8s.io/apimachinery v0.21.0
+	sigs.k8s.io/kustomize/kyaml v0.11.1-0.20210630195303-1eb3c1a07570
```
mengqiy, line 9: Using kyaml from HEAD to pick up https://github.com/kubernetes-sigs/kustomize/pull/4023.
We should change to a released version when it is released.

yhrn:
> cc: @yhrn You may want to take a look at the example.

LGTM :+1:

examples/search-replace-create-setters/.expected/exec.sh
```diff
@@ -1,5 +1,5 @@
 #! /bin/bash
 
 # shellcheck disable=SC2016
-kpt fn eval --image gcr.io/kpt-fn/search-replace:unstable -- \
+kpt fn eval --image gcr.io/kpt-fn/search-replace:unstable --image-pull-policy never -- \
```
droot, line 4: I believe you needed this during the development only, right ?

functions/go/starlark/go.mod
droot: Which functionality in current PR is effected if the referred change in kyaml is not picked up ?

droot (commented):
Change is looking good, have a few questions.

functions/go/starlark/go.mod
mengqiy: `metadata.name` will be lost after decoding.
And `metadata.name` will be used as `Name` in the `starlark.Filter`: https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/457/files#diff-45dd66dfd0fba255d8026ad222e5a8b78be5bb608fc4d6d7da177aea735a7b0eR120 And `Name` is a required field there.

Any concerns?

examples/search-replace-create-setters/.expected/exec.sh
mengqiy: Not only during the development, but also in the e2e tests. When in e2e tests, there may be changes in a newly built container, we want to ensure the local image instead of the remote.

Let me try to convert the `exec.sh` to use the `args` which is supported by the test harness. I need to ensure `source="$(cat set-replicas.star)"` still works.

mengqiy: Tried to use `args` supported by the test harness.
But it doesn't work due to some issue regarding with `"`.
I'd suggest to merge it to unblock Spotify and address this test harness issue later.

droot (approved):

bzub:
https://github.com/GoogleContainerTools/kpt/issues/1889#issuecomment-868753586

PR #466, opened 2021-07-08
Upgrade gatekeeper fn to use OPA Gatekeeper v3.5.1

halvards:
Fixes https://github.com/GoogleContainerTools/kpt/issues/2371


droot (approved):
Thanks for the pull request. Looks good to me. @mengqiy Can you confirm if all deps in order and if yes, pl. merge it.

mengqiy (approved):
Thanks for the fix

halvards:
Thanks for the quick turn-around @mengqiy and @droot!

When will a new image and tag be published with this change? I need it to update [this tutorial](https://cloud.google.com/architecture/policy-compliant-resources) from kpt v0 to v1.

PR #472, opened 2021-07-13
stop special processing StarlarkRun

mengqiy:
fixes: https://github.com/GoogleContainerTools/kpt/issues/2389


tests/starlark/eval-fn-config/Kptfile
```diff
@@ -0,0 +1,8 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: example
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/starlark:unstable
```
droot, line 7: We should also have a test case where the functionConfig is not referenced in the pipeline (that's the scenario use ran into) because then it it's part of meta resources.

droot (changes requested):

tests/starlark/eval-fn-config/Kptfile
mengqiy: Done

droot (approved):

PR #473, opened 2021-07-13
Fix starlark function to not drop id annotation

mengqiy:
Fixes: https://github.com/GoogleContainerTools/kpt/issues/2382


functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/starlark.go
```diff
@@ -0,0 +1,288 @@
...
+		return errors.Wrap(err)
+	}
+
+	_, err = writer.Write([]byte(s))
+	return err
+}
+
+// SimpleFilter transforms a set of resources through the provided starlark
+// program. It doesn't touch the id annotation. It doesn't copy comments.
+type SimpleFilter struct {
```
droot, line 204: I wonder if we can add a test case ensuring annotations are preserved.

droot (changes requested):

mengqiy:
Rebased on #474 and added a test.

droot (approved):

PR #474, opened 2021-07-13
import kyaml starlark code

mengqiy:
Preparation for https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/473

### DIFF
--- a/functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/context.go
+++ b/functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/context.go
@@ -1,0 +1,79 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package starlark
+
+import (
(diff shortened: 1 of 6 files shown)


functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/context.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2019 The Kubernetes Authors.
```
droot, line 1: it will be useful to bookkeep which version we are importing in a README.md in `third_party/README.md` and it will be better to import a tagged version if possible.

droot (changes requested):

functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/context.go
mengqiy: Done

droot (approved):

PR #500, opened 2021-07-28
Add input, output and json flag to gatekeeper

mengqiy:
The old gatekeeper function provides the following flags in https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/b7e346f0c18f972edf7ca02272734fbc68fca905/go/pkg/framework/runners/runners.go#L31-L44:

- `--input` (string): default is stdin.
- `--output` (string): default is stdout.
- `--json` (bool): default is not use json format

We used to mirror the old OSS gatekeeper function to the `gcr.io/config-management-release/policy-controller-validate` GCR in https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/b7e346f0c18f972edf7ca02272734fbc68fca905/scripts/mirrorImages.sh#L9.
And `gcr.io/config-management-release/policy-controller-validate` is a component of the policy controller which is a Google product. We saw some users in [b/193160625](b/193160625#comment9) depend on these flags when using `docker run` directly.
Before we mirror our new gatekeeper function, we should ensure it behave the same as before so that it doesn't accidentally break policy controller users.

This PR introduce these flags and make it behave the same as the old gatekeeper function.
The change is backward compatible. It doesn't affect how the function works with the function orchestrator.

This PR partially address https://github.com/GoogleContainerTools/kpt/issues/2406


functions/go/gatekeeper/main.go
```diff
@@ -76,12 +92,96 @@ func (gkp *GatekeeperProcessor) Process(resourceList *framework.ResourceList) er
...
+	return gkp.outputBuf.Write(p)
+}
+
+func (gkp *GatekeeperProcessor) addFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&gkp.input, "input", "i", stdin,
+		`path to the input file`)
+	cmd.Flags().StringVarP(&gkp.output, "output", "o", stdout,
+		`path to the output file`)
+	cmd.Flags().BoolVar(&gkp.useJSON, "json", false,
+		`input and output is JSON instead of YAML`)
```
droot, line 150: How do these flags work with the resourceList input provided by the function orchestrator.

Can you pl. describe the change in the PR description.

mengqiy: Updated. PTAL.

droot: Since we are adding these flags, Can we add a test (script based) to cover these flag's usage ?

mengqiy: Added a few tests to exercise the flags

mengqiy:
@droot PTAL

droot (approved):
Thanks for adding the tests, @mengqiy . Looks good!

PR #515, opened 2021-08-05
migrate some examples for the sops function

mengqiy:


mengqiy:
cc: @aodinokov Is there a way to make the result idempotent?

mengqiy:
> Is there a way to make the result idempotent?

@aodinokov Any ideas?

aodinokov:
>Is there a way to make the result idempotent?
@mengqiy sorry, I missed that question.. need to setup notifications.

I think if you want to create a diff and compare (If understood correctly the intent for [.expected/diff.patch](examples/contrib/sops/encrypt-with-age/.expected/diff.patch) - it won't work.
That's because of how SOPS works under-the-hood:
AFAIK it [generates a random internal key](https://github.com/mozilla/sops#encryption-protocol), encrypts all data with that, and after that it encrypts that key with different methods, e.g. with each pgp public key provided, each age recepient, each key from Vault and etc... that allows SOPS to decrypt the document having at least one (in fact it's possible [to require at least N](https://github.com/mozilla/sops#key-groups), but I didn't do that) cred: pgp private key, age recepient private key, etc...

All that works fine, but since there is a random key in the middle of the process - your encrypted data won't be same each time.
I wasn't able to find a way to provide external master-key. That means that most likely we won't be able to make it idempotent/diff based for encryption. We can use diff for decryption though.

mengqiy:
@aodinokov Thanks for your explanation and suggestion! I will try that.

mengqiy:
@aodinokov Thanks for your explanation and suggestion! I will try that.

PR #533, opened 2021-08-12
Make SOPS work with keys from parameters and function input

aodinokov:
Function input parameters can be taken from
ConfigMap and Secret matching jsonpath expression.

This allows to cover different user scenarios, e.g. 
1. public keys for encryption can be stored as a separate ConfigMap or Secret. and can be part of repo.
2. encryption and decryption fn configurations may reference the same or different configurations.
3. examples and documentation has been updated for kpt v1.


examples/contrib/sops/age-with-key-secret/README.md
```diff
@@ -0,0 +1,31 @@
+# sops: AGE example
```
mengqiy, line 1: I was trying to migrate this function use kpt v1, but there are something unclear to me.
Do you have any insights on https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/515#issuecomment-893832596?

aodinokov: @mengqiy , I've answered. BTW, I also was thinking about migrating examples to kpt v1... as I mentioned in my PR description - I tested that with kpt v1 :) 
Should I try to rework my PR on top of 515 to make it work? I guess you have plans to get rid of kpt 0.x right?

mengqiy: > Should I try to rework my PR on top of 515 to make it work?

It'd be great if you have bandwidth to do it :) 

> we probably can encrypt/decrypt and check the result after.

What you suggested in #515 makes sense. We can do it here.

> I guess you have plans to get rid of kpt 0.x right?

Yes, most functions have been migrated to use kpt v1 for testing. If you can have the SOPS function have examples (e2e tests) working with kpt v1, it'd be great!

aodinokov: @mengqiy 

>we probably can encrypt/decrypt and check the result after.

[here](https://github.com/aodinokov/kpt-functions-catalog/blob/allParams/examples/contrib/sops/workflow-with-age/.expected/exec.sh) is that approach I tried to implement:

From the beginning everything is encrypted. When we decrypt - we can get idempotent diff. After that I encrypt again and see that diff has changed: but it's hard to predict what to expect, so I assume here that if diff has changed - kpt stored the data and it was successfully encrypted. And after that I decrypt again and e2e tool should validate that diff is correct. That diff will be the same as after the first decryption, that it is an additional control to that assumption I had that files are in correct state after the encryption step.

I'll need to rework Readme:
1. need to mention that we put private key into age_keys.txt only for testing purposes. in real life for AGE it will be necessary to do $(cat $XDG_CONFIG_HOME/sops/age/keys.txt), if sops is installed and working, because [sops takes files from this location](https://github.com/mozilla/sops/tree/develop#encrypting-using-age).
2. need to mention that file [encrypt-keys](https://github.com/aodinokov/kpt-functions-catalog/blob/allParams/examples/contrib/sops/workflow-with-age/encrypt-keys.yaml) must be kept in repo - it's used for encryption back and contains info about all people/keys who can decrypt data.
3. need to add note that after decryption it's possible to continue work with kpt like. Sops encryption decryption is supposed to be used only on the TOP level module and there is no sense to use it for submodules.

I'm also thinking about making a similar example for pgp in addition to age.

Also it's possible to use different encryption keys (pgp/age/vault/etc) altogether, and that info will be set in encrypt-keys.yaml when use encrypts data. For decryption - person provides all needed data via ENV variables (SOPS_IMPORT_AGE, SOPS_IMPORT_PGP, VAULT_ADDR/VAULT_TOKEN and etc).. So I'm thinking if I should add example at least for AGE+PGP scenario so users understand the idea: there will be 2 types of keys in encrypt-keys.yaml, let's say user1 has pgp-key1 and user2 age-key1. There will be decryption for user 1 and user 2. and encryption - the same for both users.

I guess the old examples can be deleted after that.

aodinokov: I've created 2 kpt v1 - based examples/tests (should cover https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/515), removed the old ones.

The last test that fails - fails not because of this commit - I've proven that here: https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/542.

examples/contrib/sops/workflow-with-age/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Function invocation
+
+Invoke the decryption by running the following command:
+
+```shell
+$ kpt fn eval \
+        --fn-config workflow-with-age/decrypt.yaml \
+        --env SOPS_IMPORT_AGE="$(cat workflow-with-age/age_keys.txt)" \
+        --image-pull-policy ifNotPresent \
```
mengqiy, line 30: Please drop this line in the user-facing doc.
We can leave it to use the default value for this flag.

examples/contrib/sops/workflow-with-age/README.md
```diff
@@ -0,0 +1,84 @@
...
+Note: since all resouces are decrypted at that point it's possible to do all standard operations like render, apply &etc.
+See [Kptfile](Kptfile) to check what will be done on render invocation.
+
+Invoke the encryption by running the following command:
+
+```shell
+$ kpt fn eval \
+        --fn-config workflow-with-age/encrypt.yaml \
+        --include-meta-resources \
+        --image-pull-policy ifNotPresent \
```
mengqiy, line 54: Same here.

examples/contrib/sops/workflow-with-pgp-and-age/.expected/exec.sh
```diff
@@ -0,0 +1,24 @@
+#! /bin/bash
+
+# decrypt
+kpt fn eval --fn-config decrypt.yaml --env SOPS_IMPORT_PGP="$(cat pgp_keys.txt)" \
+	--include-meta-resources \
+	--image-pull-policy ifNotPresent --image gcr.io/kpt-fn-contrib/sops:unstable
```
mengqiy, line 6: Please use `--image-pull-policy never`.

examples/contrib/sops/workflow-with-pgp-and-age/.expected/exec.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# make sure that docs don't contain ENC entries
+grep 'ENC\[AES256_GCM' ./*.yaml && exit 1
+
+# NOTE: at this point can modify setter config, do kpt render, apply, do 3way merge of submodules & etc
+
+# encrypt
+kpt fn eval --fn-config encrypt.yaml \
+	--include-meta-resources \
+	--image-pull-policy ifNotPresent --image gcr.io/kpt-fn-contrib/sops:unstable
```
mengqiy, line 16: Same here.

mengqiy, line 16: Same here.

examples/contrib/sops/workflow-with-pgp-and-age/.expected/exec.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# make sure that docs contain ENC entries
+grep 'ENC\[AES256_GCM' ./*.yaml || exit 1
+
+# decrypt
+kpt fn eval --fn-config decrypt.yaml --env SOPS_IMPORT_AGE="$(cat age_keys.txt)" \
+        --include-meta-resources \
+        --image-pull-policy ifNotPresent --image gcr.io/kpt-fn-contrib/sops:unstable
```
mengqiy, line 24: Same here

examples/contrib/sops/workflow-with-pgp-and-age/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+Invoke the decryption by running one of the following command:
+
+1. decryption with AGE key
+
+```shell
+$ kpt fn eval \
+        --fn-config workflow-with-pgp-and-age/decrypt.yaml \
+        --env SOPS_IMPORT_AGE="$(cat workflow-with-pgp-and-age/age_keys.txt)" \
+        --image-pull-policy ifNotPresent \
```
mengqiy, line 32: Please drop this line.

examples/contrib/sops/workflow-with-pgp-and-age/README.md
```diff
@@ -0,0 +1,100 @@
...
+        workflow-with-pgp-and-age/
+```
+
+2. decryption with PGP key:
+
+```shell
+$ kpt fn eval \
+        --fn-config workflow-with-pgp-and-age/decrypt.yaml \
+        --env SOPS_IMPORT_PGP="$(cat workflow-with-pgp-and-age/pgp_keys.txt)" \
+        --image-pull-policy ifNotPresent \
```
mengqiy, line 44: Please drop this line.

examples/contrib/sops/workflow-with-pgp-and-age/README.md
```diff
@@ -0,0 +1,100 @@
...
+Note: since all resouces are decrypted at that point it's possible to do all standard operations like render, apply &etc.
+See [Kptfile](Kptfile) to check what will be done on render invocation.
+
+Invoke the encryption by running the following command:
+
+```shell
+$ kpt fn eval \
+        --fn-config workflow-with-pgp-and-age/encrypt.yaml \
+        --include-meta-resources \
+        --image-pull-policy ifNotPresent \
```
mengqiy, line 70: Please drop this line.

examples/contrib/sops/workflow-with-age/.expected/exec.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# make sure that docs don't contain ENC entries
+grep 'ENC\[AES256_GCM' ./*.yaml && exit 1
+
+# NOTE: at this point can modify setter config, do kpt render, apply, do 3way merge of submodules & etc
+
+# encrypt
+kpt fn eval --fn-config encrypt.yaml \
+	--include-meta-resources \
+	--image-pull-policy ifNotPresent --image gcr.io/kpt-fn-contrib/sops:unstable
```
mengqiy, line 16: Please use `--image-pull-policy never`.

mengqiy (commented):
A few minor comments. Otherwise, looks good.

aodinokov:
@mengqiy , done :) Thank you for the thorough review - not sure how I forgot to update exec.sh in the second example :)

mengqiy (approved):

PR #535, opened 2021-08-12
fork helm inflation code from kustomize

natasha41575:


functions/go/inflate-helm-chart/metadata.yaml
```diff
@@ -0,0 +1,9 @@
+image: gcr.io/kpt-fn/inflate-helm-chart
```
mengqiy, line 1: You can make verify-docs.py to skip your directory by add `inflate-helm-chart` in https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/8e56c28a78baf04464520e5d6b3e29aa77382a03/scripts/verify-docs.py#L39

With that you don't need to add this file here.

scripts/verify-docs.py
```diff
@@ -36,7 +36,7 @@
 import subprocess
 
 metadata_filename = 'metadata.yaml'
-directories_to_skip = ['_template', 'dist', 'node_modules']
+directories_to_skip = ['_template', 'dist', 'node_modules', 'inflate-helm-chart']
```
mengqiy, line 39: Please don't forget to revert this change back in the next PR.

mengqiy (approved):

PR #537, opened 2021-08-13
fix: avoid roundtriping when passing objects to starlark fn

mengqiy:
fixes: https://github.com/GoogleContainerTools/kpt/issues/2438


mengqiy:
@natasha41575 This fix only changes our fork, I will upstream the fix to kustomize.

phanimarupaka (approved):

PR #539, opened 2021-08-13
remove helm inflator plugin unit tests

natasha41575:


functions/go/inflate-helm-chart/helminflatorplugin_test.go
```diff
@@ -1,623 +0,0 @@
-// Copyright 2019 The Kubernetes Authors.
```
mengqiy, line 1: Why we had k8s license header in files outside of third_party directory?

mengqiy (approved):

functions/go/inflate-helm-chart/helminflatorplugin_test.go
natasha41575: My mistake, I must have copied and pasted something incorrectly. I've audited the other files to check that the license headers are correct, sorry about that

mengqiy: No worries. You can do that in a separate PR.

PR #540, opened 2021-08-14
Merge gob branch with master before collecting functions.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/2446

### DIFF
--- a/.github/workflows/checkSite.yml
+++ b/.github/workflows/checkSite.yml
@@ -27,6 +27,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - name: Merge gob branch
+        run: git --work-tree=. checkout gob .
       - uses: actions/checkout@v2
         with:
           repository: etefera/href-checker
--- a/.github/workflows/checkSite.yml
+++ b/.github/workflows/checkSite.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Merge gob branch
-        run: git --work-tree=. checkout gob .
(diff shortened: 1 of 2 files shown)


.github/workflows/firebase-hosting-merge.yml
```diff
@@ -14,6 +14,8 @@ jobs:
       - uses: actions/checkout@v2
```
mengqiy, line 14: We should add `gob` as a trigger branches, right?

.github/workflows/checkSite.yml
```diff
@@ -27,6 +27,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - name: Merge gob branch
+        run: git --work-tree=. checkout gob .
```
mengqiy, line 31: This command doesn't work. It seems the following works.
```
git worktree add ./gob gob
cp -r gob/* .
```

.github/workflows/firebase-hosting-merge.yml
mengqiy: Add it to checkSite.yml

.github/workflows/checkSite.yml
etefera: This command worked with `upstream/gob` in place of `gob`. Switched it to your command.

.github/workflows/firebase-hosting-merge.yml
etefera: Done. checkSite is triggered on every branch.

mengqiy (approved):

PR #547, opened 2021-08-23
Fix master branch CI

aodinokov:
CI was failing with error `http://localhost:3001/master//README.md [x1] {404}`
That was because scripts/generate_catalog/generate_catalog.go generates site/catalog.json
without RemoteSourcePath for that function and in that case index.html generates
href = localhost:3001/master//README.md for the button 'Edit this page'.
This patch adds sourceURL, so now catalog.json should contain correct data.


aodinokov:
cc: @etefera, @mengqiy, @phanimarupaka

mengqiy (approved):

PR #554, opened 2021-08-27
This adds the generate-services function to the catalog

g-awmalik:


functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Input
+
+This function accepts all KRM resources, but ignores those that are not managed by Config Controller.
+
+The following config options are supported:
+- `namespace` - namespace set on the generated Services (default: `gcp-services`)
+- `disable-on-destroy` - value for the `cnrm.cloud.google.com/disable-on-destroy` annotation on the generated Services (default: `""` - no annotation)
+
+These options can be configured declaratively in the data field of a `ConfigMap` or imperatively following a `kpt fn run` command.
```
droot, line 23: s/kpt fn run/kpt fn eval

functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Overview
+
+<!--mdtogo:Short-->
+
+The generate-services kpt function generates [Config Controller Service resources](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service) required for usage of the other Config Controller resources supplied to the function.
+
+<!--mdtogo-->
+
+This removes the requirement that users explicitly enable services, but more importantly, it makes it easier to author kpt packages, because Service resources no longer need to be embedded in every package and there's no need for a standard naming scheme across package providers.
```
droot, line 11: aah..so this generates resources that enables GCP APIs for the services required by this package. Is this correct ?

functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
...
+<!--mdtogo:Long-->
+
+## Usage
+
+1. Create any Config Controller resource(s), in one or more yaml files, anywhere in a directory.
+2. Create a ConfigMap with the `function` annotation to configure this resource as input to a containerized kpt function.
+    Example:
+    ```
+    annotations:
+      config.kubernetes.io/function: |
```
droot, line 52: kpt v1 doesn't support this syntax. It was supported for kpt alpha.

functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
...
+        container:
+          image: gcr.io/kpt-fn/generate-services:unstable
+    ```
+3. (Optional) Set the `data.namespace` field to specify which namespace Services will be created in.
+4. (Optional) Set the `data.disable-on-destroy` field to specify the value of the `cnrm.cloud.google.com/disable-on-destroy` annotation.
+5. (Optional) Add the `config.kubernetes.io/local-config: "true"` annotation to tell ConfigSync to exclude the ConfigMap resource when applying.
+6. Run [kpt fn run](https://googlecontainertools.github.io/kpt/guides/consumer/function/#declarative-run) on the directory containing the resource yaml files.
```
droot, line 59: This is linking to older version of `kpt` and also referring to `kpt fn run` ? that is not supported by kpt v1.

droot (commented):
Took a quick look and have a few questions.

g-awmalik:
Thanks for review @droot. I'm going to put this back in draft since we may have this functionality covered in another function. If and when we decided to release this, I'll make sure I address your questions/comments.

examples/generate-services-imperative/README.md
```diff
@@ -0,0 +1,27 @@
+# generate-services: Imperative Example
+
+### Overview
+
+This examples shows how to generate [Config Controller Service resources](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service) given a set Config Controller resources in the package by running [`generate-services`] function imperatively.
```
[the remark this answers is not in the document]
g-awmalik: This is instance of a GCP Service e.g. compute.googleapis.com that is required before you can provision a GCP network or a GCE instance for that matter. I'll add another link but the documentation is limited on this unfortunately.

examples/generate-services-imperative/app.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: container.cnrm.cloud.google.com/v1beta1
+kind: ContainerCluster
+metadata:
+  name: example
+  namespace: service-project-id # kpt-set: ${service-project-id}
```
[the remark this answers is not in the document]
g-awmalik: We could but since the example already project annotations/elements in it, it's better to mention the setters instead. Plus, IMO its better to focus on generate-services and adding `set-project-id` in there may create confusion and/or complexity.

functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
+# generate-services
```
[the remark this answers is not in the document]
g-awmalik: True. I'm open to changing it but CC actually specifies is as a `Service` and that's why we tried to keep it consistent. Although k8s has a `Service` resource as well which does make things confusing. IMO, we should keep it consistent with CC but if you feel strongly otherwise, let me know.

functions/go/generate-services/README.md
g-awmalik: That is correct

functions/go/generate-services/README.md
g-awmalik: Acked

functions/go/generate-services/README.md
g-awmalik: Fixed.

functions/go/generate-services/README.md
g-awmalik: Yup, fixed it.

g-awmalik:
@mengqiy - can you please trigger the build workflow for this?

mengqiy:
@g-awmalik Done!

functions/go/Makefile
```diff
@@ -26,6 +26,7 @@ FUNCTIONS := \
 	format \
 	gatekeeper \
 	remove-local-config-resources \
+	generate-services \
```
mengqiy, line 29: > `	generate-services \	`

There is an empty `\t` after `\`. It should be removed. It is causing the CI to fail.

Also please sort the list.

g-awmalik:
@mengqiy - fixed the unit test issue and it's passing all checks now.

functions/go/generate-services/main_test.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		t.Fatal("not a valid KRM resource\n", err)
+	}
+
+	resourceList := MakeResourceListFromNodes(itemProject, itemServiceDiffProject)
+
+	existingServices, err := createExistingServicesMap(resourceList)
+	if err != nil {
+		t.Fatal("unable to create existing services map\n", err)
+	}
```
bharathkkb, line 177: We can use `require.NoError(err)`

functions/go/generate-services/main.go
```diff
@@ -0,0 +1,373 @@
...
+		services = append(services[:modifiedInd], services[modifiedInd+1:]...)
+	}
+
+	return services
+}
+
+func main() {
+	processor := framework.ResourceListProcessorFunc(func(resourceList *framework.ResourceList) error {
+
+		serviceMapPath := os.Getenv(envServiceMapPath)
```
bharathkkb, line 174: splitting this out from main would make this more testable

functions/go/generate-services/main_test.go
```diff
@@ -0,0 +1,193 @@
...
+	projectServices, err := createProjectServicesMap(resourceList, serviceMap, existingServices)
+	if err != nil {
+		t.Fatal("unable to create project services map\n", err)
+	}
+
+	// require 0 services because there is already a Service resource for "projects"
+	require.Equal(t, 0, len(projectServices["projects"]))
+}
+
+func TestExistingServiceNotPresent(t *testing.T) {
```
bharathkkb, line 129: This can be grouped along with TestExistingServicePresent as two subtests since most of the test code is same

functions/go/generate-services/main.go
```diff
@@ -0,0 +1,373 @@
...
+		disableOnDestroy, err := getValueOrDefault(configMap, configFieldDisableOnDestroy, defaultDisableOnDestroy)
+		if err != nil {
+			return err
+		}
+
+		// Map of Services to list of Project IDs that are already resourced. This is required for
+		// making this function idempotent
+		existingServices, err := createExistingServicesMap(resourceList)
+		if err != nil {
+			log.Fatal("unable to create exisitng services map")
```
bharathkkb, line 200: This wont be surfaced to the user. We should return this as the err.

functions/go/generate-services/main.go
```diff
@@ -0,0 +1,373 @@
...
+		}
+
+		disableOnDestroy, err := getValueOrDefault(configMap, configFieldDisableOnDestroy, defaultDisableOnDestroy)
+		if err != nil {
+			return err
+		}
+
+		// Map of Services to list of Project IDs that are already resourced. This is required for
+		// making this function idempotent
+		existingServices, err := createExistingServicesMap(resourceList)
```
bharathkkb, line 198: Would this also take user created services into consideration?

droot:
sorry for not reviewing this PR last week. I plan to take a look at it, but wanted to confirm, does it have any interplay with https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/568 ?

bharathkkb:
> wanted to confirm, does it have any interplay with #568

@droot both deal with generating service resources. Key difference is that generate-services scans resources to generate corresponding APIs while project-services(#568) require users to explicitly specify which services to generate. If a user adds ContainerCluster for project foo, it will generate service for enabling container.googleapis.com in project foo.

Generally the use case for `generate-services` is when all GCP resources are managed by KCC and admins do not want tight control over API enablement. OTOH `project-services`(#568) just allows an admin who is creating a project to succinctly define all the services necessary in a single resource and have tighter control over which services are enabled in a specific project.

functions/go/generate-services/main.go
```diff
@@ -0,0 +1,373 @@
...
+const defaultServiceNamespace = "gcp-services"
+const configFieldDisableOnDestroy = "disable-on-destroy"
+const defaultDisableOnDestroy = "" // empty means no annotation
+const annotationDisableOnDestroy = "cnrm.cloud.google.com/disable-on-destroy"
+const annotationProjectID = "cnrm.cloud.google.com/project-id"
+const defaultServiceMapPath = "service-map.yaml"
+const envServiceMapPath = "GENERATE_SERVICES_MAP"
+const serviceHostNameSuffix = ".googleapis.com"
+const serviceAPIVersion = "serviceusage.cnrm.cloud.google.com/v1beta1"
+const serviceKind = "Service"
```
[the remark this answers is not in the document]
g-awmalik: Acked.

examples/generate-services-imperative/README.md
```diff
@@ -0,0 +1,27 @@
...
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/generate-services-imperative
+```
+
+### Function invocation
+
+Invoke the function by running the following commands:
+
+```shell
+$ kpt fn eval generate-services-imperative --image gcr.io/kpt-fn/generate-services:unstable
```
mengqiy, line 20: I'd suggest you to show users how to use `namespace` and `disable-on-destroy`.
And then in the `Expected result` section, you can mention that these settings have been reflected on the output resources.

functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+Services produced by this function are also de-duplicated, meaning there will only ever be one Service resource enabling that service in that particular project. This means that the `cnrm.cloud.google.com/disable-on-destroy: "true"` annotation is no longer required, and services can now be disabled when they are no longer in use, as long as all the Resources for that project are managed in the same source directory or repository.
+
+## Input
+
+This function accepts all KRM resources, but ignores those that are not managed by Config Controller.
+
+The following config options are supported:
+- `namespace` - namespace set on the generated Services (default: `gcp-services`)
+- `disable-on-destroy` - value for the `cnrm.cloud.google.com/disable-on-destroy` annotation on the generated Services (default: `""` - no annotation)
```
mengqiy, line 21: This field seems to be a boolean. We should cover what are the allowed value.

functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+These options can be configured declaratively in the data field of a `ConfigMap` or imperatively following a `kpt fn eval` command.
+
+## Output
+
+Executing the generate-services function will generate a unique set of Service resources.
+
+Naming pattern:
```
mengqiy, line 29: We should say it's naming pattern for `metadata.name`

functions/go/generate-services/main.go
```diff
@@ -0,0 +1,373 @@
...
+// been resourced to their list of project-ids
+func createExistingServicesMap(resourceList *framework.ResourceList) (map[string][]string, error) {
+	existingServices := make(map[string][]string)
+	for _, item := range resourceList.Items {
+		itemMeta, err := item.GetMeta()
+		if err != nil {
+			return nil, err
+		}
+
+		if itemMeta.Name == "" || itemMeta.APIVersion == "" || itemMeta.Kind == "" {
```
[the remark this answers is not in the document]
g-awmalik: Sure

functions/go/generate-services/main.go
```diff
@@ -0,0 +1,373 @@
...
+			// Skip. Not a valid KRM resource.
+			continue
+		}
+
+		//if item is not a Service, continue
+		if itemMeta.Kind != serviceKind {
```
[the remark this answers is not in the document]
g-awmalik: Acked

functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+The `HOSTNAME_PREFIX` will be the Service API Hostname without the `.googleapis.com` suffix, with periods replaced by hyphens (ex: `compute.googleapis.com` -> `compute`).
+
+File path pattern:
+
+```
+${NAMESPACE}/service_${NAME}.yaml
+```
+
+<!--mdtogo:Long-->
```
mengqiy, line 43: `<!--mdtogo:Long-->` and `<!--mdtogo-->` should cover
```
## Usage
### Input
### Output
```

functions/go/generate-services/README.md
```diff
@@ -0,0 +1,115 @@
...
+This removes the requirement that users explicitly enable services, but more importantly, it makes it easier to author kpt packages, because Service resources no longer need to be embedded in every package and there's no need for a standard naming scheme across package providers.
+
+Services produced by this function are also de-duplicated, meaning there will only ever be one Service resource enabling that service in that particular project. This means that the `cnrm.cloud.google.com/disable-on-destroy: "true"` annotation is no longer required, and services can now be disabled when they are no longer in use, as long as all the Resources for that project are managed in the same source directory or repository.
+
+## Input
+
+This function accepts all KRM resources, but ignores those that are not managed by Config Controller.
+
+The following config options are supported:
+- `namespace` - namespace set on the generated Services (default: `gcp-services`)
```
mengqiy, line 20: We should call it out when the fields are optional

mengqiy:
@bharathkkb Thanks for clarifying the difference between the 2 functions in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/554#issuecomment-916353181
IMO we should convert https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/554#issuecomment-916353181 into the usage doc for both functions. We can have something like `function foo does X. If you want to do Y, please look at function bar`.

mengqiy:
Please include a `.expected/results.yaml` file in the example so that we can exercise the output.

functions/go/generate-services/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+image: gcr.io/kpt-fn/generate-services
+description: Generates Config Controller Service resources required for supplied Config Controller resources to function.
+tags:
+  - mutator
```
mengqiy, line 4: I'd suggest add labels "generator" "GCP"

functions/go/generate-services/README.md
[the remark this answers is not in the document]
mengqiy: After looking at https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/568, I feel this function (#554) should be called `enable-gcp-services` and #568 should be called `generate-gcp-services`. The reason is that #568 generate the Service resources explicitly specified by the users; while #554 automatically generate the Services resources to enable GCP services based the input resources.

@g-awmalik @bharathkkb @droot WDYT?

functions/go/generate-services/service-map.yaml
```diff
@@ -0,0 +1,201 @@
+apiVersion: blueprints.cloud.google.com/v1alpha1
```
mengqiy, line 1: Do we have source of truth for this file somewhere? How can we ensure it's up-to-date?

examples/generate-services-imperative/app.yaml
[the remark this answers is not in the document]
mengqiy: IIUC we don't need to show these setters here. The function looks at the `apiVersion` and `kind` and then determine what Service does it need.
If the setters are not related here, we should drop it. Doing that can reduce the chance that it confuse the users.

examples/generate-services-imperative/.expected/diff.patch
```diff
@@ -0,0 +1,60 @@
...
++  namespace: gcp-services
++  annotations:
++    cnrm.cloud.google.com/project-id: host-project-id
++spec:
++  resourceID: compute.googleapis.com
+diff --git a/gcp-services/service_projects-cloudresourcemanager.yaml b/gcp-services/service_projects-cloudresourcemanager.yaml
+new file mode 100644
+index 0000000..c9c74a3
+--- /dev/null
++++ b/gcp-services/service_projects-cloudresourcemanager.yaml
```
mengqiy, line 20: What is the reason that we put everything generated in a directory?

functions/go/generate-services/main.go
```diff
@@ -0,0 +1,373 @@
...
+		return defaultValue, nil
+	}
+	return value, nil
+}
+
+// getProjectID returns the project-id annotation value, if set and not empty,
+// otherwise the namespace is assumed to be the project-id, because that's what
+// ConfigConnectorContext uses by default.
+func getProjectID(node *yaml.RNode) (string, error) {
+	key, err := node.Pipe(yaml.GetAnnotation(annotationProjectID))
```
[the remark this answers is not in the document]
g-awmalik: Yes we can. I modified the code and made it simpler.

functions/go/generate-services/README.md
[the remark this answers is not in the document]
g-awmalik: I believe @morgante commented on this in [#568](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/568#discussion_r705824117). If combine the two makes sense to everyone, let's do that and simplify.

g-awmalik:
Closing. this based on the [discussion](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/568#discussion_r705820568) to merge this functionality with the `project-services` function.

PR #559, opened 2021-08-31
Update checkSite.yml

mengqiy:


droot (approved):
nit: PR description (and the title) isn't very useful.

PR #566, opened 2021-09-03
document how subjects works with set-namespace

mengqiy:
Fixes: https://github.com/GoogleContainerTools/kpt/issues/2397


functions/go/set-namespace/README.md
```diff
@@ -36,10 +36,26 @@ by default the function will also update the [fields][commonnamespace] that
...
 - If there is a `Namespace` resource, its `metadata.name` field will be updated.
-- If it's a `RoleBinding` or `ClusterRoleBinding` resource, `subject` can
-  reference either a namespaced resource or a cluster-scoped resource. When
-  there are references to namespace in the `subject` fields, they will be
-  updated by the function.
+- If there's a `RoleBinding` or `ClusterRoleBinding` resource, the function will
+  update the namespace in the `ServiceAccount` if and only if the subject
+  element `name` is `default`. In the following example, the `set-namespace`
+  function will update `subjects.namespace` since the
+  corresponding `subjects.name` is `default`.
```
droot, line 43: Why do we have this limitation ?

droot (approved):

functions/go/set-namespace/README.md
mengqiy: This behavior is inherited from kustomize. The users may not always want to change `subjects.namespace` in `RoleBinding` and `ClusterRoleBinding`. With this behavior, the user can control if the they want the namespace to be updated.
IMO when we have the resource selector feature, we will no longer need this feature.

mengqiy: @natasha41575 Do you have any insights about that behavior in kustomize?

PR #568, opened 2021-09-07
feat: Project Service List local CRD function

bharathkkb:
This adds `ProjectServiceList` local CRD used for generating GCP API Service enablement resources


examples/project-service-list-simple/.expected/config.yaml
```diff
@@ -0,0 +1 @@
+testType: render
```
droot, line 1: I think default testype is `render`, so you can omit this file completely.

examples/project-service-list-simple/README.md
```diff
@@ -0,0 +1,33 @@
...
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/project-service-list-simple
+```
+
+### Function invocation
+
+Render the configuration, `project-service-list` has been defined declaratively in the Kptfile and configured via `services.yaml`.
+
+```shell
+$ kpt fn render project-service-list-simple --results-dir /tmp
```
droot, line 20: since we are specifying `results-dir` here, we should show how to examine the results. Or we can remove `results-dir` ?

examples/project-service-list-simple/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+### Expected result
+
+```shell
+[RUNNING] "gcr.io/kpt-fn/project-service-list:unstable"
+[PASS] "gcr.io/kpt-fn/project-service-list:unstable"
+  Results:
+    [INFO] generated service in object "serviceusage.cnrm.cloud.google.com/v1beta1/Service/proj1-service-compute"
+    [INFO] generated service in object "serviceusage.cnrm.cloud.google.com/v1beta1/Service/proj1-service-redis"
```
droot, line 30: Note to self: I realized the default out from `kpt fn render` isn't well suited for generating resources. For example if it omitted `in`, the sentence will read better.

functions/go/project-service-list/README.md
```diff
@@ -0,0 +1,117 @@
+# project-services-list
+
+## Overview
+
+<!--mdtogo:Short-->
+
+The `project-services-list` function generates [project service](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service)
+resources from a list of services to enable [GCP APIs](https://cloud.google.com/apis) within a specified project.
```
droot, line 8: May be I am missing something, but `enable-gcpservice-apis` might be better name for this function. right ? 
I think there is another function (similar to this) also in progress.

functions/go/project-service-list/README.md
```diff
@@ -0,0 +1,117 @@
+# project-services-list
+
+## Overview
+
+<!--mdtogo:Short-->
+
+The `project-services-list` function generates [project service](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service)
```
droot, line 7: I think it will help to describe the overall motivation to help user understand why and when to use this function.

functions/go/project-service-list/README.md
```diff
@@ -0,0 +1,117 @@
...
+apiVersion: blueprints.cloud.google.com/v1alpha1
+kind: ProjectServiceList
+metadata:
+  name: my-project-services
+  annotations:
+    cnrm.cloud.google.com/deletion-policy: false
+spec:
+  services: # list of services to generate
+    - compute.googleapis.com
+  projectID: foo # explicit project ID to be set for the generated services
```
droot, line 52: Will it make sense to make it work better with `set-project-id` function ? for ex. adding the comment hints to make it work seamlessly with `set-project-id` ?

functions/go/project-service-list/README.md
```diff
@@ -0,0 +1,117 @@
...
+kind: Service
+metadata:
+  name: proj1-service-redis
+  annotations:
+    blueprints.cloud.google.com/managed-by-project-services-list: 'proj1-service'
+spec:
+  resourceID: redis.googleapis.com
+  projectRef:
+    external: proj1
+```
```
droot, line 115: A few more questions:

1. Can the function scan resources in the package and generate the resources for enabling the GCP APIs automatically ? (may be it does, it wasn't clear from the documentation).
2. Ordering: so when do `kpt live apply`, it is expected to apply the resource that enable the API for a service first, then the instance for the service ?

droot (commented):
Some questions.

functions/go/project-service-list/README.md
bharathkkb: 1) No but generate services #554 does this. This requires an explicit list, was https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/5432cdca9675ed6c7c2fbab13fc916ece0b3d2dc/functions/go/project-service-list/README.md#usage clear in this or should I add some more verbiage. We can also cross link to generate-services.

2) kpt will apply both together. Resource will fail at first, API will succeed and eventually resource will succeed. We can make this better when we have dependency ordering in kpt.

functions/go/project-service-list/README.md
bharathkkb: Yup, sgtm. I think only change necessary is to add the `project-id` setter. `set-project-id` should then create pr update the value for `project-id` setter when its invoked.

functions/go/project-service-list/README.md
bharathkkb: I tried to align with our Terraform equivalent https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/modules/project_services
/cc @morgante if you have any suggestions

examples/project-service-list-simple/README.md
bharathkkb: I think we can remove `results-dir` here. Most use these results in this case comes from stdout.

examples/project-service-list-simple/.expected/config.yaml
bharathkkb: Per docs here https://github.com/GoogleContainerTools/kpt/blob/main/pkg/test/runner/README.md#how-does-it-work it seemed like default was eval.

mengqiy: I created https://github.com/GoogleContainerTools/kpt/pull/2492 to fix it.

examples/project-service-list-simple/.expected/diff.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/service_proj1-service-compute.yaml b/service_proj1-service-compute.yaml
+new file mode 100644
+index 0000000..b02d5d8
+--- /dev/null
++++ b/service_proj1-service-compute.yaml
+@@ -0,0 +1,10 @@
++apiVersion: serviceusage.cnrm.cloud.google.com/v1beta1
++kind: Service
++metadata:
++  name: proj1-service-compute
```
mengqiy, line 10: #554 allows you to configure the namespace of the service. What is the best practice for using namespace in `Services`?

functions/go/project-service-list/README.md
morgante: I think we could consider combining with https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/554 and calling the combined function `enable-gcp-services`. By default, it would just process the ProjectServices CRD but if you set `generate: true` it could also generate new services.

@g-awmalik What do you think?

g-awmalik: I agree. @bharathkkb and I were on the fence about having another function that is kinda similar but not. This flag would address both requirements and eliminate an additional (almost redundant) function.

If all agree, I can work with @bharathkkb to merge the logic in here.

mengqiy: Merging these 2 sounds reasonable to me, and it can help reduce user confusion.

examples/project-service-list-simple/.expected/diff.patch
bharathkkb: Any annotations/namespace set for ProjectServiceList CR is added to all generated service resources. I have a section [here](https://github.com/bharathkkb/kpt-functions-catalog/tree/project-services/functions/go/project-service-list#:~:text=Sets%20namespace%20if%20any%20defined%20for%20ProjectServiceList%20custom%20resource%20to%20each%20generated%20resource.) on that. Users can choose to set a NS (recommended) or use default.

functions/go/project-service-list/metadata.yaml
```diff
@@ -0,0 +1,11 @@
+image: gcr.io/kpt-fn/project-service-list
+description: Generate and manage project service enablements for a GCP project.
+tags:
+  - generator
+  - GCP
```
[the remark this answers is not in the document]
bharathkkb: done

examples/project-service-list-simple/resources.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: redis.cnrm.cloud.google.com/v1beta1
```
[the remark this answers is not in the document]
bharathkkb: This is a workaround for https://github.com/GoogleContainerTools/kpt/issues/2477

functions/go/project-service-list/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/project-services
+
+go 1.16
+
+require (
+	github.com/stretchr/testify v1.5.1
+	sigs.k8s.io/kustomize/kyaml v0.11.0
```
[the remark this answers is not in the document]
bharathkkb: done

functions/go/project-service-list/README.md
```diff
@@ -0,0 +1,117 @@
...
+  name: my-project-services
+  annotations:
+    cnrm.cloud.google.com/deletion-policy: false
+spec:
+  services: # list of services to generate
+    - compute.googleapis.com
+  projectID: foo # explicit project ID to be set for the generated services
+```
+
+| Field        |  Description | Example
```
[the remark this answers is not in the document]
bharathkkb: Added another column for required/optional. No defaults, if project is not specified we fall back to [KCC defaults](https://cloud.google.com/config-connector/docs/concepts/namespaces-and-projects#how_uses_namespaces). If no services specified, [we throw an error](https://github.com/bharathkkb/kpt-functions-catalog/blob/5432cdca9675ed6c7c2fbab13fc916ece0b3d2dc/functions/go/project-service-list/projectservicelist/project_service_list.go#L65).

examples/project-service-list-simple/README.md
```diff
@@ -0,0 +1,33 @@
+# project-service-list: Simple Example
+
+### Overview
+
+In this example, we will see how to generate [project services](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service) for `compute.googleapis.com` and `redis.googleapis.com`.
```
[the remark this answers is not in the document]
bharathkkb: The [resource name](https://cloud.google.com/service-usage/docs/reference/rest/v1/services) is service. Since its GCP projects, we call it a project service. Would `GCP project services` be more clearer?

functions/go/project-service-list/README.md
bharathkkb: added project-id setter

functions/go/project-service-list/README.md
bharathkkb: added a line to describe why/when to use this

functions/go/project-service-list/README.md
bharathkkb: SGTM, we can the generate option in a followup PR.

examples/project-service-list-simple/README.md
bharathkkb: removed

examples/project-service-list-simple/.expected/config.yaml
bharathkkb: removed testype

functions/go/project-service-list/README.md
g-awmalik: I'll close the [554](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/554) in that case and work with Bharath to to move that logic here.

droot: Great.

examples/project-service-list-simple/README.md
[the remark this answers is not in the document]
mengqiy: IMO `GCP project services` is better.

mengqiy:
From our earlier discussion, we are going to merge the functionality of #554 into this PR.
@bharathkkb Do you plan to make the changes in this PR?

bharathkkb:
@mengqiy We are planning to add that generate functionality in a follow up but I will rename this function to `enable-gcp-services` in this PR.

bharathkkb:
@mengqiy this should be ready for another look

examples/enable-gcp-services-simple/resources.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: redis.cnrm.cloud.google.com/v1beta1
```
mengqiy, line 1: I'd suggest you to drop this file, since it doesn't affect how the function behave.

examples/enable-gcp-services-simple/services.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: blueprints.cloud.google.com/v1alpha1
+kind: ProjectServiceList
+metadata:
+  name: proj1-service
+spec:
+  services:
+    - compute.googleapis.com
+    - redis.googleapis.com
+  projectID: proj1
```
mengqiy, line 9: Please add a newline at the end

tests/enable-gcp-services/idempotent-prune/.expected/results.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: kpt.dev/v1
+kind: FunctionResultList
+metadata:
+  name: fnresults
+exitCode: 0
+items:
+  - image: gcr.io/kpt-fn/enable-gcp-services:unstable
+    exitCode: 0
+    results:
```
mengqiy, line 9: I don't quite follow what is the purpose of this test. Can you please explain? Maybe add a comment some where.

functions/go/enable-gcp-services/README.md
```diff
@@ -0,0 +1,118 @@
...
+has a `blueprints.cloud.google.com/managed-by-enable-gcp-services` annotation. This annotation allows `enable-gcp-services` function to
+track generated resources for the declarative management of the generated resources. Any changes made to the generate resources will be overwritten and should be made to the `ProjectServiceList` CRD instead.
+
+### FunctionConfig
+
+This function only supports a custom resource `functionConfig` of kind `ProjectServiceList`.
+
+#### `ProjectServiceList`
+
+A functionConfig of kind `ProjectServiceList` has the following supported parameters:
```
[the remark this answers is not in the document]
morgante: This should really not be a `functionConfig` at all actually. It should be a normal resource in the project which is transformed.

In general, I think we should reconsider that convention. CRD names should focus on descriptively explaining their purpose, not coupling to a particular function.

functions/go/enable-gcp-services/README.md
```diff
@@ -0,0 +1,118 @@
+# enable-gcp-services
+
+## Overview
+
+<!--mdtogo:Short-->
+
+The `enable-gcp-services` function generates [project service](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service)
```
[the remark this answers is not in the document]
bharathkkb: done

tests/enable-gcp-services/idempotent-prune/.expected/config.yaml
```diff
@@ -0,0 +1 @@
+runCount: 3
```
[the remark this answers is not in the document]
bharathkkb: switched to 2

functions/go/enable-gcp-services/README.md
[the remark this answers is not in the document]
bharathkkb: So `ProjectServiceList` is a list of `Services` (thats what [KCC](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service) calls them). I think adding `kind` is unnecessary as we only generate `Services`. We can switch to `EnableGCPServices` if that is a convention.

examples/enable-gcp-services-simple/resources.yaml
bharathkkb: I added this as a workaround for https://github.com/GoogleContainerTools/kpt/issues/2477. For tests, do we test with kpt built from main, if so I can drop this.

examples/enable-gcp-services-simple/services.yaml
bharathkkb: done

tests/enable-gcp-services/idempotent-prune/.expected/results.yaml
bharathkkb: Added a [comment](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/0549f1360eb7df7ffee25401589b022d12a5420e/tests/enable-gcp-services/idempotent-prune/services.yaml#L6). What I am testing is that a given ProjectServiceList prunes any previous service it generated and regenerates new ones. This allows us to manage all the generated services owned by that  ProjectServiceList (like removing an old service, removing any manual changes to a generated service etc).

functions/go/enable-gcp-services/README.md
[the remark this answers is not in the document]
bharathkkb: Just saw @morgante's comment. 

> This should really not be a functionConfig at all actually

[Existing pattern](https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/go/render-helm-chart#renderhelmchart) seemed to be to use it as a `functionConfig`. An advantage of this is that we don't need to mark this as a local-config for kpt live and defines an explicit package boundary for the `ProjectServiceList` resource rather than discovering all of them.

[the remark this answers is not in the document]
morgante: I'm not sure the existing pattern is really desirable. Functions are something like local controllers—a controller shouldn't only accept a single instance of the CRD. In particular, we don't want the function to only accept a single list of services—it's feasible to want to split the list into two CRDs. You shouldn't have to call the function twice to generate both.

I'm also considering where future changes like `enableGenerate` should go—do you imagine that going on the `ProjectServiceList`?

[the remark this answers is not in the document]
mengqiy: Functions are like client-side controller. The `functionConfig` is like the object that triggers the reconciliation in the controller. In a server-side controller, one root object is processed in one reconciliation. e.g. Deployment foo and bar both changes. The Deployment manager processes foo and bar in 2 separate reconciliations.
Currently, a function only accepts one **explicit `functionConfig`** (passed in using `input resourceList .functionConfig`) per https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md#schema.

There is another pattern that may be useful in this case. We can use **implicit `functionConfig`** (passed in using `input resourceList .items`). You can pass in arbitrary number of objects as "functionConfig".  Existing example is `gatekeeper`. The `ConstraintTemplate` and the actual Constraint are the "functionConfig" for `gatekeeper`. Multiple can be passed in through `input resourceList .items`.

The intent of y original comment is simply about the naming, since `ProjectServiceList` implies it's a list of `ProjectService`. IMO passing a list to `resourceList .functionConfig` is not user-friendly and is abusing the `.functionConfig` field.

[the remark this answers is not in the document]
morgante: > In a server-side controller, one root object is processed in one reconciliation. e.g. Deployment foo and bar both changes.

But you don't need a separate instance of a controller for each resource. So it's inaccurate to claim that you should have a single CR for each function.

> There is another pattern that may be useful in this case. We can use implicit functionConfig (passed in using input resourceList .items). You

Right, that's the pattern we should be use here.

> The intent of y original comment is simply about the naming, since ProjectServiceList implies it's a list of ProjectService.

It is essentially a list of ProjectService. Technically the underlying resource is named [Service](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service) but since that term is overloaded we suggested `ProjectServiceList`.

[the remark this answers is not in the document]
bharathkkb: I'll investigate using the implicit fn config pattern. Regarding `enableGenerate` I was thinking this could be an explicit fn  config since for a given pkg they only need to set `enableGenerate` once. This would also support nested subpkgs since a user can configure `enableGenerate` in subpkg A while defining `ProjectServiceList` for subPkg B.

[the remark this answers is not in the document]
g-awmalik: Defining it once makes sense and less prone to error. However, I think we can accomplish that with the same fn config. E.g. we could have a boolean attribute `spec.autoGenerate` in `ProjectServiceList` that if set to true generates the required services for the project i.e. `spec.ProjectID`.

Although, this brings up another question i.e. does this fn assume that all resources are being actuated in the same project specified under `spec.ProjectID` since I don't see the project IDs specified on the resources themselves?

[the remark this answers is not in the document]
morgante: There should be a single `ProjectServiceList` per project.

examples/enable-gcp-services-simple/resources.yaml
bharathkkb: I have removed this now we are using implicit fn config

examples/project-service-list-simple/README.md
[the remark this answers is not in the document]
bharathkkb: done

functions/go/enable-gcp-services/README.md
```diff
@@ -0,0 +1,120 @@
...
+kind: ProjectServiceList
+metadata:
+  name: my-project-services
+  annotations:
+    cnrm.cloud.google.com/deletion-policy: false
+    config.kubernetes.io/local-config: true
+spec:
+  services: # list of services to generate
+    - compute.googleapis.com
+  projectID: foo # kpt-set: ${project-id}
```
mengqiy, line 53: > kpt-set: ${project-id}

The setter comment is not really related here. Please drop it to reduce the chance of user confusion.

examples/enable-gcp-services-simple/README.md
```diff
@@ -0,0 +1,33 @@
...
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/enable-gcp-services-simple
+```
+
+### Function invocation
+
+Render the configuration, `enable-gcp-services` has been defined declaratively in the Kptfile and configured via `services.yaml`.
+
+```shell
+$ kpt fn render enable-gcp-services-simple
```
mengqiy, line 20: It may be helpful to let users run it with `--results-dir results` and then let the users check the structured results.
WDYT?

functions/go/enable-gcp-services/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN go mod download
+
+COPY . .
+RUN go build -o /usr/local/bin/function ./
+
+#############################################
+
+FROM alpine:3.13
+COPY --from=0 /usr/local/bin/function /usr/local/bin/function
+ENTRYPOINT ["function"]
```
mengqiy, line 15: Please add a newline

tests/enable-gcp-services/idempotent-prune/services.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: blueprints.cloud.google.com/v1alpha1
+kind: ProjectServiceList
+metadata:
+  name: proj1-service
+  annotations:
+    config.kubernetes.io/local-config: "true"
+spec:
+  services: # We expect only generated service resources by proj1-service ProjectServiceList to be pruned and regenarated. All other service resources remain unmodified.
```
mengqiy, line 8: I don't see this test exercises the prune behavior.
IIUC if I initially have a KCC object for `bigquery.googleapis.com` with annotation `blueprints.cloud.google.com/managed-by-enable-gcp-services: 'proj1-service'`, after I run the function, this bigquery KCC resource should disappear, right?

functions/go/enable-gcp-services/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+`enable-gcp-services` function can be used both declaratively and imperatively.
+
+```shell
+kpt fn eval --image gcr.io/kpt-fn/enable-gcp-services:unstable
+```
+
+The `enable-gcp-services` function does the following:
+
+1. Discovers all `ProjectServiceList` custom resources in a given package.
```
mengqiy, line 25: Nested packages are scanned as well. You should mention it.

functions/go/enable-gcp-services/gcpservices/project_service_list.go
```diff
@@ -0,0 +1,230 @@
...
+
+// ProjectServiceListRunner discovers ProjectServiceList CRs to generate Service CRs
+type ProjectServiceListRunner struct {
+	results []Result
+}
+
+// ProjectServiceList is the functionConfig for generating services
+type ProjectServiceList struct {
+	yaml.ResourceMeta `json:",inline" yaml:",inline"`
+	Spec              struct {
```
mengqiy, line 32: Please don't embed a struct in another type.
Please extract it out to its own definition.

functions/go/enable-gcp-services/gcpservices/project_service_list.go
```diff
@@ -0,0 +1,230 @@
...
+		ProjectID string   `json:"projectID,omitempty" yaml:"projectID,omitempty"`
+	} `json:"spec" yaml:"spec"`
+	results []Result
+}
+
+// Result represents an action performed by ProjectServiceList on a resource
+type Result struct {
+	ResourceRef yaml.ResourceIdentifier
+	FilePath    string
+	Action      string
```
mengqiy, line 43: It seems we have a limited number of `Action`s.
We can do something like `type ActionType string` and then define `GeneratedService` and `PruneService` action types.

functions/go/enable-gcp-services/gcpservices/project_service_list.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return k == ps.Name
+}
+
+// pruneGeneratedServices removes resources managed by ProjectServiceList
+func (ps *ProjectServiceList) pruneGeneratedServices(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
+	n := 0
+	for i := range nodes {
+		if !ps.isManaged(nodes[i]) {
+			nodes[n] = nodes[i]
```
mengqiy, line 220: It seems there is an issue with in-place modification: suppose there are some services managed by ProjectServiceList 1 and some other services are managed by ProjectServiceList 2. If we process ProjectServiceList 1 first, the services managed by ProjectServiceList 2 will be lost.

We should have a test to exercise the behavior

functions/go/enable-gcp-services/gcpservices/services.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// Service represents a valid GCP Service
+type Service struct {
+	APIVersion string `yaml:"apiVersion"`
```
mengqiy, line 12: You can use https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml@v0.11.1/yaml#ResourceMeta

functions/go/enable-gcp-services/gcpservices/services.go
```diff
@@ -0,0 +1,55 @@
...
+	Kind       string `yaml:"kind"`
+	Metadata   struct {
+		Name string `yaml:"name"`
+	} `yaml:"metadata"`
+	Spec struct {
```
mengqiy, line 17: Please exact it

functions/go/enable-gcp-services/gcpservices/services.go
```diff
@@ -0,0 +1,55 @@
...
+		ResourceID string `yaml:"resourceID"`
+		ProjectRef struct {
+			External string `yaml:"external,omitempty"`
+		} `yaml:"projectRef,omitempty"`
+	} `yaml:"spec"`
+}
+
+// getServicesList creates a slice of services from service strings
+func getServicesList(name string, inputServices []string, projectID string) ([]Service, error) {
```
mengqiy, line 26: `getServicesList` doesn't seem to be accurate naming here. Maybe we can use `CreateServicesList` or `ConstructServicesList` or something more reasonable.

functions/go/enable-gcp-services/gcpservices/project_service_list.go
```diff
@@ -0,0 +1,230 @@
+package gcpservices
```
mengqiy, line 1: Do we expect others to import this package?
If not, we can keep it flat and we will not need to define `type Result struct` to pass the information around. Instead we can directly operate on `ResultItem`.

bharathkkb: Initially I thought we would reuse for the generate-services fn, but switched to ResultItem since that was dropped.

functions/go/enable-gcp-services/gcpservices/services.go
bharathkkb: done

functions/go/enable-gcp-services/gcpservices/services.go
bharathkkb: done

functions/go/enable-gcp-services/gcpservices/services.go
bharathkkb: done

functions/go/enable-gcp-services/gcpservices/project_service_list.go
bharathkkb: Added a test. For a fn invocation all services managed by any ProjectServiceList is pruned. In your example if `ProjectServiceList 2` CR still exists, the objects are regenerated so after fn execution it should be noop.

functions/go/enable-gcp-services/gcpservices/project_service_list.go
bharathkkb: done

functions/go/enable-gcp-services/gcpservices/project_service_list.go
bharathkkb: done

functions/go/enable-gcp-services/README.md
bharathkkb: added

tests/enable-gcp-services/idempotent-prune/services.yaml
bharathkkb: Thats correct, I added this scenario to the test as well.

functions/go/enable-gcp-services/Dockerfile
bharathkkb: done

examples/enable-gcp-services-simple/README.md
bharathkkb: I dropped this per https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/568#discussion_r705514638
We surface results via stdout and I don't see much value in this simple example to examine structured results.

functions/go/enable-gcp-services/README.md
bharathkkb: Added per https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/568#discussion_r705523525
After re reading the readme, I agree that it might be confusing, so dropped.

functions/go/enable-gcp-services/README.md
```diff
@@ -0,0 +1,120 @@
...
+    * Adds all annotations defined for `ProjectServiceList` CR to each generated resource. This can be used for enabling features like
+[disable-on-destroy](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service#custom_resource_definition_properties) for generated services.
+    * Sets namespace if any defined for `ProjectServiceList` CR to each generated resource.
+    * Sets projectID if any defined for `ProjectServiceList` CR to each generated resource.
+1. Each generated [GCP project service](https://cloud.google.com/config-connector/docs/reference/resource-docs/serviceusage/service) resource
+has a `blueprints.cloud.google.com/managed-by-enable-gcp-services` annotation. This annotation allows `enable-gcp-services` function to
+track generated resources for the declarative management of the generated resources. Any changes made to the generate resources will be overwritten and should be made to the `ProjectServiceList` CR instead.
+
+### `ProjectServiceList`
```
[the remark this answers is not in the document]
bharathkkb: @mengqiy would `EnableGCPServices` be better? /cc @morgante if you have any thoughts.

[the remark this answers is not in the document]
morgante: CRDs should be nouns, not verbs.

Since `List` is a reserved name, how about `ProjectServiceSet`. There seems to be some precedent for using `Set` for non-list collections. Ex. https://docs.projectcalico.org/reference/resources/globalnetworkset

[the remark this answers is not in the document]
bharathkkb: @mengqiy can we move ahead with `ProjectServiceSet`?

tests/enable-gcp-services/idempotent-prune/.expected/config.yaml
```diff
@@ -0,0 +1 @@
+runCount: 2
```
mengqiy, line 1: Default is 2. So we can remove this file.

tests/enable-gcp-services/idempotent-prune/.expected/results.yaml
```diff
@@ -0,0 +1,49 @@
...
+          path: service_proj1-service-bigquery.yaml
+      - message: pruned service
+        severity: info
+        resourceRef:
+          apiVersion: serviceusage.cnrm.cloud.google.com/v1beta1
+          kind: Service
+          name: proj1-service-compute
+        file:
+          path: service_proj1-service-compute.yaml
+      - message: pruned service
```
mengqiy, line 26: We show `pruned service` for `proj1-service-redis` in `service_proj1-service-redis.yaml` and later we show `generated service` for the same resource. These 2 operations combined together is a noop.
IMO this behavior may be a little confusing for users, since we actually only have one real operation here which is pruning `proj1-service-bigquery`, but we show users 5 operations. It will become harder to read when there are more resources. e.g. I have 10 resources unchanged and only one resource get pruned, there will be 21 events in the resultsList, it will be hard for users to figure out what really going on.
Maybe we can have something called `recreated service` to handle this case.
@bharathkkb WDYT?

functions/go/enable-gcp-services/metadata.yaml
```diff
@@ -0,0 +1,12 @@
+image: gcr.io/kpt-fn/enable-gcp-services
+description: Generate and manage project service enablements for a GCP project.
+tags:
+  - generator
+  - GCP
+  - mutator
+sourceURL: https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/go/enable-gcp-services
+examplePackageURLs:
```
mengqiy, line 8: What do you think about moving the pruning e2e test to examples? IMO it'd be useful showing the users how pruning work in an example.

examples/enable-gcp-services-simple/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+### Expected result
+
```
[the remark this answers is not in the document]
bharathkkb: added more info

tests/enable-gcp-services/idempotent-prune/.expected/results.yaml
bharathkkb: great idea, I have implemented this

functions/go/enable-gcp-services/metadata.yaml
bharathkkb: moved to example

tests/enable-gcp-services/idempotent-prune/.expected/config.yaml
bharathkkb: done

mengqiy (commented):
A few minor comments, LGTM otherwise

examples/enable-gcp-services-advanced/README.md
```diff
@@ -0,0 +1,74 @@
+# enable-gcp-services: Advanced Example
```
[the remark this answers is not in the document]
mengqiy: This example actually 2 things:
- pruning
- multiple projects

So please ignore this comment.

functions/go/enable-gcp-services/gcpservices/project_service_set.go
```diff
@@ -0,0 +1,279 @@
...
+	return append(nodes, generatedNodes...), nil
+}
+
+func (ps ProjectServiceSet) validate() error {
+	if len(ps.Spec.Services) < 1 {
+		return fmt.Errorf("at least one service must be specified under spec.services[]")
+	}
+	val, found := ps.Annotations[filters.LocalConfigAnnotation]
+	if !found {
+		return fmt.Errorf("%s annotation must be set", filters.LocalConfigAnnotation)
```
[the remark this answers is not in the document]
bharathkkb: Yup that is simpler, done

functions/go/enable-gcp-services/gcpservices/project_service_set.go
```diff
@@ -0,0 +1,279 @@
...
+		// add any new generated serviceNodes
+		generatedNodes = append(generatedNodes, newServiceNodes...)
+		r.results = append(r.results, psl.getResults()...)
+	}
+	return append(nodes, generatedNodes...), nil
+}
+
+func (ps ProjectServiceSet) validate() error {
+	if len(ps.Spec.Services) < 1 {
+		return fmt.Errorf("at least one service must be specified under spec.services[]")
```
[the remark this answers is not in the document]
bharathkkb: done

functions/go/enable-gcp-services/README.md
```diff
@@ -0,0 +1,120 @@
...
+  services:
+    - compute.googleapis.com
+    - redis.googleapis.com
+  projectID: proj1
+```
+
+Invoke the function:
+
+```shell
+kpt fn eval --image gcr.io/kpt-fn/enable-gcp-services:unstable
```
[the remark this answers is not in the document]
bharathkkb: done

functions/go/enable-gcp-services/README.md
```diff
@@ -0,0 +1,120 @@
...
+<!--mdtogo-->
+
+<!--mdtogo:Long-->
+
+## Usage
+
+`enable-gcp-services` function can be used both declaratively and imperatively.
+
+```shell
+kpt fn eval --image gcr.io/kpt-fn/enable-gcp-services:unstable
```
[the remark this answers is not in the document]
bharathkkb: done

examples/enable-gcp-services-advanced/README.md
```diff
@@ -0,0 +1,74 @@
...
+├── [proj2-services.yaml]  ProjectServiceSet proj2-service
+├── [resources.yaml]  ComputeNetwork computenetwork-sample
+├── [resources.yaml]  RedisInstance redisinstance-sample
+├── [service_proj1-service-bigquery.yaml]  Service proj1-service-bigquery
+├── [service_proj1-service-compute.yaml]  Service proj1-service-compute
+├── [service_proj1-service-redis.yaml]  Service proj1-service-redis
+└── [service_proj2-service-redis.yaml]  Service proj2-service-redis
+```
+
+We can see two ProjectServiceSet resources `proj1-service` and `proj2-service` for managing service enablement in two projects `proj1` and `proj2`. We can also see the services `compute.googleapis.com`, `redis.googleapis.com` managed within ProjectServiceSet `proj1-service` resource and `redis.googleapis.com` within ProjectServiceSet `proj2-service` resource. Notice that `bigquery.googleapis.com` is no longer present in `proj1-service` ProjectServiceSet. We can re render this package to prune that service resource.
```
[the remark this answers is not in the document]
bharathkkb: done

mengqiy (approved):
LGTM, thanks for working on this!

PR #570, opened 2021-09-08
support loading starlib in starlark fn

mengqiy:
Fixes: https://github.com/GoogleContainerTools/kpt/issues/1083 and https://github.com/GoogleContainerTools/kpt/issues/2427


mengqiy:
cc: @yhrn and @karlkfi

droot (approved):
This is exciting. This opens up possibility of introducing our own KRM library that helps with operating on KRM resources.

karlkfi:
Sweet, thanks!

Now we just need the ability to publish Starlark functions as containers or have kpt run them from a URL, then Starlark can be the defacto function language!

PR #572, opened 2021-09-08
enable some non-standard features in starlark

mengqiy:
Enabled set, global reassign, recursion

LoadBindsGlobally has been deprecated, and it is not enabled. Ref: https://pkg.go.dev/go.starlark.net/resolve#pkg-variables

fixes: https://github.com/GoogleContainerTools/kpt/issues/2429


mengqiy:
cc: @karlkfi

droot (approved):

PR #574, opened 2021-09-09
This adds the generate-folders function to the kpt catalog

g-awmalik:
Migrating the generate-folders source from GoB to GitHub.

### DIFF
--- a/functions/ts/generate-folders/metadata.yaml
+++ b/functions/ts/generate-folders/metadata.yaml
@@ -3,6 +3,7 @@ description: Transform the ResourceHierarchy custom resource into Folder custom 
 tags:
   - mutator
   - GCP
+  - generator
 sourceURL: https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/ts/generate-folders
 examplePackageURLs:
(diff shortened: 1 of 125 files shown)


functions/ts/generate-folders/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+image: gcr.io/kpt-fn/generate-folders
+description: Transform the ResourceHierarchy custom resource into Folder custom resources constituting the hierarchy.
+tags:
+  - mutator
```
mengqiy, line 4: Please add a `generator` tag.

mengqiy (approved):
One nit.
LGTM, given this is migrated from GoB.

bharathkkb:
In a follow up we should remove `generate-folders-resourcehierarchy-v1` and `generate-folders-resourcehierarchy-v2` from examples and add those under tests instead. `v3`  is the [recommended](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/3e7b2ebe37651d7a9d03c8b04801f4cdad0efcf3/functions/ts/generate-folders/src/generate_folders.ts#L76) version that should be surfaced in examples.

PR #610, opened 2021-09-30
Merge master to kubeval/v0.1

mengqiy:
#608 is the fix for kubeval


mengqiy:
We have updated the baked-in openapi version from 1.19 to 1.20.
It'd be safer to create a new minor release, in case there are breaking changes in the newer openapi schema.

PR #649, opened 2021-11-18
add support for ConstraintTemplate v1

mengqiy:


mengqiy:
@juliankatz PTAL

functions/go/gatekeeper/validate.go
```diff
@@ -56,15 +58,16 @@ func createClient() (*opaclient.Client, error) {
...
-		}
 		templ := &templates.ConstraintTemplate{}
-		if err := scheme.Convert(ct, templ, nil); err != nil {
-			return nil, err
+		switch obj.(type) {
+		case *v1.ConstraintTemplate, *v1beta1.ConstraintTemplate, *v1alpha1.ConstraintTemplate:
+			if err := scheme.Convert(obj, templ, nil); err != nil {
+				return nil, err
+			}
+			templs = append(templs, templ)
```
julianKatz, line 67: I recommend substituting this implementation for [that of gktest](https://github.com/open-policy-agent/gatekeeper/blob/master/pkg/gktest/read_constraints.go#L112).  It uses the `ToVersionless()` func that's built into each version of ConstraintTemplate (v1alpha1, v1beta1, v1).

The upside of this is that `ToVersionless()` also [takes care of defaulting](https://github.com/open-policy-agent/frameworks/blob/master/constraint/pkg/apis/templates/v1/helpers.go#L11).  Switching CT schemas to be structural (which is required in k8s clusters >= v1.22) required us to do some special defaulting stuff under the hood.

Rather than one day deal with those problems, using `ToVersionless()` will save you the headache :)

Here's the logic you'll need, as well as I can write it in this comment without running it:

```golang
// outside of this function
type versionless interface {
	ToVersionless() (*templates.ConstraintTemplate, error)
}

// ...

// within the function
v, isVersionless := obj.(versionless)
if !isVersionless {
	continue
}

// this returns a &templates.ConstraintTemplate{}
template, err := v.ToVersionless()
if err != nil {
	// this means there's a bug probably, return to user
}
```

mengqiy: Done!
Great suggestion!

functions/go/gatekeeper/go.mod
```diff
@@ -1,12 +1,88 @@
...
-go 1.16
+go 1.17
 
 require (
-	github.com/open-policy-agent/frameworks/constraint v0.0.0-20210522003146-5c034948ac29
-	github.com/open-policy-agent/gatekeeper v0.0.0-20210527161344-e229247f04d1 // This is v3.5.1. It has a semver major version of 2 or higher and is not a Go module yet.
-	github.com/spf13/cobra v1.1.1
-	k8s.io/apimachinery v0.20.2
+	github.com/open-policy-agent/frameworks/constraint v0.0.0-20211025234246-f478d8a555b5
+	github.com/open-policy-agent/gatekeeper v0.0.0-20211115204247-3ba8e93049b5 // This is v3.7.0. It has a semver major version of 2 or higher and is not a Go module yet.
```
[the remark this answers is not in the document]
mengqiy: The import path should be  `github.com/open-policy-agent/gatekeeper/v3` if it's compatible with go modules. Otherwise, we have to use SHA. It's not a big deal.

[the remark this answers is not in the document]
julianKatz: gotcha!  Thanks

tests/gatekeeper/v1beta1-template/policy.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
```
julianKatz, line 1: You might consider an additional test with a `v1` template.  New users will likely use `v1`, but it will be a long time before the `v1beta1` template are gone.

julianKatz (approved):

julianKatz:
I have limited context on writing kpt funcs, but the g8r-related changes LGTM.

tests/gatekeeper/v1beta1-template/policy.yaml
mengqiy: Our examples are tests too. I have changed all of our examples work with v1 template.

droot (approved):

PR #656, opened 2021-11-20
chore: cancel old commit CI for PRs

bharathkkb:
GitHub Actions lets you cancel builds when new commits are pushed, similar to Prow. In most cases you probably want CI results for new commits and don't care about older commits. More details [here](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) if you are interested.

I saw this while iterating on my other PR https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/64 and multiple  CI jobs were running for +30m on old commits.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,6 +12,11 @@ on:
     - cron:  '0 17 * * *'
 
 concurrency:
+  # If a job with a concurrency group is running with cancel-in-progress, that job is cancelled when a new job with same concurrency group is started.
+  # In a workflow github.head_ref exists only for PRs while github.ref exists for the repo branches/tags.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -11,6 +11,10 @@ on:
   schedule:
     - cron:  '0 17 * * *'
 
+concurrency:
+  group: '${{ github.head_ref || github.ref }}-ci'
```
mengqiy, line 15: I don't understand why we are using `${{ github.head_ref || github.ref }}-ci` as the group name?
Can you please add a comment here?

bharathkkb: Added a comment above this line, lmk if that doesn't make sense.

PR #666, opened 2021-12-06
Site examples patch version

sdowell:
Issue: GoogleContainerTools/kpt#2609

This adds a new catalog-v2.json endpoint which expands upon the previous catalog.json schema to also include the latest patch version for each major/minor function release. This is presented as a new endpoint in order to avoid breaking compatibility with existing consumers of the original endpoint, such as versions of kpt already in the wild.

Updates the catalog pages of the website to include the latest patch version in:
- The version dropdown for the function
- A new dynamically rendered latest image section

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -117,9 +117,9 @@
 
               let catalogPath;
               if (isContribFn) {
-                catalogPath = "/contrib/catalog.json";
+                catalogPath = "/contrib/catalog-v2.json";
               } else {
-                catalogPath = "/catalog.json";
+                catalogPath = "/catalog-v2.json";
               }
               window.Docsify.get(catalogPath).then(function (
                 catalogResponse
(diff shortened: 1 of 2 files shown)


sdowell:
## Before clicking the dropdown:
image

## After clicking the dropdown
image

droot:
I find that image URL to be really big and distracting. It also doesn't tell what that is. One possible presentation could be having a 

Image: <box> (with the dimensions shown below code section of github)


image

droot:
One more question: if you change the version in the drop down, do the image URL also get changed ? 

I think we can also probably display it under the function heading:

apply-setters
==========

<b>Image URL: </b> `gcr.io/kpt-fn/apply-setters:v0.1.1 `

site/index.html
```diff
@@ -117,9 +117,9 @@
 
               let catalogPath;
               if (isContribFn) {
-                catalogPath = "/contrib/catalog.json";
+                catalogPath = "/contrib/catalog-v2.json";
               } else {
-                catalogPath = "/catalog.json";
```
mengqiy, line 122: Do we still serve catalog.json?

sdowell: Yes, it's still served at the same path

droot:
> One more question: if you change the version in the drop down, do the image URL also get changed ?
> 
> I think we can also probably display it under the function heading:
> 
> # apply-setters
> Image URL: `gcr.io/kpt-fn/apply-setters:v0.1.1 `

Closing the loop on this. We discussed offline and decided to add this as an incremental change. Scope of this PR is now reduced to just ensuring content in the catalog has latest patch version instead of the minor versions.

droot (commented):
Will defer to @mengqiy to approve it.

mengqiy (approved):

PR #667, opened 2021-12-08
test: Fix e2e tests

sdowell:
These tests were broken due to upstream changes. The stderr
output includes additional output which would be fragile to
capture in a test case, so the assertion is changed to
results.yaml.

The contrib test is being skipped for now pending further
investigation.


mengqiy (approved):

droot:
fixes https://github.com/GoogleContainerTools/kpt/issues/2611

PR #668, opened 2021-12-09
Add script for updating function docs

sdowell:
This script is intended to be used for updating the docs for
a provided release branch to reflect the latest patch version
of that release.

Issues: GoogleContainerTools/kpt#2609

### DIFF
--- a/scripts/update_function_docs/update_function_docs.go
+++ b/scripts/update_function_docs/update_function_docs.go
@@ -73,6 +73,10 @@ func gitCheckout(branch string) error {
 	return err
 }
 
+func gitTag() (string, error) {
+	return runCmd("git", "tag")
+}
+
 func gitAdd() error {
 	_, err := runCmd("git", "add", "-u")
 	return err
@@ -92,7 +96,9 @@ func gitShow() error {
 }
 
 var (
+	// pattern of release branches, e.g. apply-setters/v1.0
 	releaseBranchPattern = regexp.MustCompile(`[-\w]*\/(v\d*\.\d*)`)
+	// pattern of release tags, e.g. functions/go/apply-setters/v1.0.1
 	releaseTagPattern    = regexp.MustCompile(`.*(go|ts)\/[-\w]*\/(v\d*\.\d*\.\d*)`)
 )
 
@@ -119,6 +125,7 @@ func newFunctionRelease(branch string) (*functionRelease, error) {
 		return nil, fmt.Errorf("invalid branch format")
 	}
 	segments := strings.Split(branch, "/")
+	// assume branch format: */<func_name>/<minor_version>
 	fr.MinorVersion = segments[len(segments)-1]
 	fr.FunctionName = segments[len(segments)-2]
 	if err := fr.readLatestPatchVersion(); err != nil {
@@ -135,7 +142,7 @@ func (fr *functionRelease) readLatestPatchVersion() error {
 	if fr.FunctionName == "" || fr.MinorVersion == "" {
 		return fmt.Errorf("missing function name and/or minor version")
 	}
-	tags, err := runCmd("git", "tag")
+	tags, err := gitTag()
 	if err != nil {
 		return err
 	}
@@ -168,15 +175,29 @@ func (fr *functionRelease) readDocPaths() error {
 		return err
 	}
 	repoBase := filepath.Dir(filepath.Dir(filepath.Dir(executablePath)))
-	fr.FunctionPath = filepath.Join(repoBase, "functions", fr.Language, fr.FunctionName)
-	examplesPath := filepath.Join(repoBase, "examples")
-
-	if !dirExists(fr.FunctionPath) {
-		fr.FunctionPath = filepath.Join(repoBase, "contrib", "functions", fr.Language, fr.FunctionName)
-		examplesPath = filepath.Join(repoBase, "contrib", "examples")
-		if !dirExists(fr.FunctionPath) {
-			return fmt.Errorf("function doc paths not found %s", fr.FunctionPath)
+	pathsToTry := []struct{
+		functionPath string
+		examplesPath string
+	}{
+		{
+			functionPath: filepath.Join(repoBase, "functions", fr.Language, fr.FunctionName),
+			examplesPath: filepath.Join(repoBase, "examples"),
+		},
+		{
+			functionPath: filepath.Join(repoBase, "contrib", "functions", fr.Language, fr.FunctionName),
+			examplesPath: filepath.Join(repoBase, "contrib", "examples"),
+		},
+	}
+	var examplesPath string
+	for _, pathToTry := range pathsToTry {
+		if dirExists(pathToTry.functionPath) {
+			fr.FunctionPath = pathToTry.functionPath
+			examplesPath = pathToTry.examplesPath
+			break
 		}
+	}
+	if fr.FunctionPath == "" {
+		return fmt.Errorf("function doc paths not found from %+v", pathsToTry)
 	}
 	if err = fr.parseMetadata(examplesPath); err != nil {
 		return err
--- a/scripts/update_function_docs/update_function_docs.go
+++ b/scripts/update_function_docs/update_function_docs.go
@@ -100,6 +100,8 @@ var (
 	releaseBranchPattern = regexp.MustCompile(`[-\w]*\/(v\d*\.\d*)`)
 	// pattern of release tags, e.g. functions/go/apply-setters/v1.0.1
 	releaseTagPattern    = regexp.MustCompile(`.*(go|ts)\/[-\w]*\/(v\d*\.\d*\.\d*)`)
+	// pattern for version tags, e.g. unstable, v0.1.1, v0.1
+	versionGroup         = `unstable|v\d*\.\d*\.\d*|v\d*\.\d*`
 )
 
 func dirExists(path string) bool {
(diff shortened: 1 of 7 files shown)


scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Usage: update_function_docs <RELEASE_BRANCH>
+//
+// e.g. update_function_docs origin/apply-setters/v0.2
```
mengqiy, line 17: I'd suggest rename this file to `main.go`.
We should update it to something like `go run ./main.go origin/apply-setters/v0.2` which feels more like a script.

scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+func runCmd(name string, arg ...string) (string, error) {
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	cmd := exec.Command(name, arg...)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	fmt.Printf("%s\n", cmd.String())
+	err := cmd.Run()
+	if err != nil {
+		return stdout.String(), fmt.Errorf("%s\n%s", stderr.String(), err)
```
mengqiy, line 53: Why not print both stderr and stdout?
If we choose to do that, we can use https://pkg.go.dev/os/exec#Cmd.CombinedOutput

scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	return stdout.String(), err
+}
+
+func isCleanRepo() bool {
+	_, err := runCmd("git", "diff-index", "--quiet", "HEAD", "--")
```
mengqiy, line 59: Why is it better than running `git status`?

scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func gitShow() error {
+	stdout, err := runCmd("git", "show")
+	fmt.Printf("%v\n", stdout)
+	return err
+}
+
+var (
+	releaseBranchPattern = regexp.MustCompile(`[-\w]*\/(v\d*\.\d*)`)
```
mengqiy, line 95: Please add some comments. e.g. give a few examples of the targets.

scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+	releaseTagPattern    = regexp.MustCompile(`.*(go|ts)\/[-\w]*\/(v\d*\.\d*\.\d*)`)
```
mengqiy, line 96: Same here

scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// newFunctionRelease allocates and initializes a functionRelease
+func newFunctionRelease(branch string) (*functionRelease, error) {
+	fr := &functionRelease{}
+	if !releaseBranchPattern.MatchString(branch) {
+		return nil, fmt.Errorf("invalid branch format")
+	}
+	segments := strings.Split(branch, "/")
+	fr.MinorVersion = segments[len(segments)-1]
```
mengqiy, line 122: Add comments why we pick segment backward.
It can help folks with little context.

scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	return fr, nil
+}
+
+// readLatestPatchVersion of the release from git tags
+func (fr *functionRelease) readLatestPatchVersion() error {
+	if fr.FunctionName == "" || fr.MinorVersion == "" {
+		return fmt.Errorf("missing function name and/or minor version")
+	}
+	tags, err := runCmd("git", "tag")
```
mengqiy, line 138: You have helper fn for all other git commands. Why not add one?

scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	repoBase := filepath.Dir(filepath.Dir(filepath.Dir(executablePath)))
+	fr.FunctionPath = filepath.Join(repoBase, "functions", fr.Language, fr.FunctionName)
+	examplesPath := filepath.Join(repoBase, "examples")
+
+	if !dirExists(fr.FunctionPath) {
+		fr.FunctionPath = filepath.Join(repoBase, "contrib", "functions", fr.Language, fr.FunctionName)
+		examplesPath = filepath.Join(repoBase, "contrib", "examples")
+		if !dirExists(fr.FunctionPath) {
+			return fmt.Errorf("function doc paths not found %s", fr.FunctionPath)
```
mengqiy, line 178: This error message can be a little misleading, since it complains the contrib path only.

scripts/update_function_docs/update_function_docs.go
```diff
@@ -0,0 +1,294 @@
...
+	tagPattern := regexp.MustCompile(
+		fmt.Sprintf(`%s(:|/)%s`, fr.FunctionName, versionGroup))
+	contents = tagPattern.ReplaceAll(contents,
+		[]byte(fmt.Sprintf(`%s${1}%s`, fr.FunctionName, fr.LatestPatchVersion)))
+	// replace url with minor e.g. https://catalog.kpt.dev/apply-setters/v1.0
+	urlPattern := regexp.MustCompile(
+		fmt.Sprintf(`https://catalog\.kpt\.dev/%s/%s`, fr.FunctionName, versionGroup))
+	contents = urlPattern.ReplaceAll(contents,
+		[]byte(fmt.Sprintf(`https://catalog.kpt.dev/%s/%s`, fr.FunctionName, fr.MinorVersion)))
+
```
mengqiy, line 236: What about updating
- from `kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/kubeval-simple` to `kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/kubeval-simple@kubeval/v0.2.1`
- from `kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/kubeval-simple/v0.2.0` to `kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/kubeval-simple@kubeval/v0.2.1`

sdowell: The case I have intentionally accounted for is
- `kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/kubeval-simple@kubeval/v0.2` to `kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/kubeval-simple@kubeval/v0.2.1`

Which is handled by the `(:|/)` capturing group.

I was not aware that there are other cases that need to be handled. Do you have examples for the other cases we need to handle?

scripts/update_function_docs/update_function_docs.go
sdowell: Updated

scripts/update_function_docs/update_function_docs.go
sdowell: Done

scripts/update_function_docs/update_function_docs.go
sdowell: Done

scripts/update_function_docs/update_function_docs.go
sdowell: Done

scripts/update_function_docs/update_function_docs.go
sdowell: Done

scripts/update_function_docs/update_function_docs.go
sdowell: The primary reason is that this is less fragile than the `git status` alternative.

`git status` returns a zero exit code in both scenarios:
- No changes between HEAD and working tree (i.e. clean repo)
- Changes between HEAD and working tree (i.e. dirty repo)

The only way to differentiate between the two using the porcelain `git status` would be to parse stdout/stderr.

However, the plumbing `git diff-index` command offers an alternative approach - as it can be configured to return a nonzero exit code if there are changes between HEAD and the working tree.

See: https://git-scm.com/docs/git-diff-index#Documentation/git-diff-index.txt---quiet

scripts/update_function_docs/update_function_docs.go
sdowell: I was fairly deliberate here with what I chose to print in order to provide useful yet not overly verbose output

The structure here is intended to print stderr in the case of a fatal error (i.e. exitWithError).

The stdout is deliberately printed for some of the commands to show to the user what changes were made (e.g. `gitCommit`, `gitShow`). It becomes a bit too verbose if the stdout is printed for all of the commands. Also, I think it is useful to keep stdout separate from stderr for the purpose of parsing stdout in the code.

scripts/update_function_docs/update_function_docs.go
sdowell: I don't have a strong opinion on this, so this is okay with me. However, I was following the pattern of `generate_catalog` - so I figure we may want to keep it consistent one way or the other.

sdowell: I have done some refactoring of the module and moved the `main` function to `main.go`

scripts/update_function_docs/update_function_docs.go
sdowell: I did a bit of refactoring and added a `replaceKptPackages` method to handle this case

mengqiy (approved):
Thanks for writing this script

PR #672, opened 2021-12-10
Revert "Site examples patch version"

mengqiy:
Reverts GoogleContainerTools/kpt-functions-catalog#666


mengqiy:
Our site version selection part is broken, roll it back.

PR #673, opened 2021-12-10
fix: Use minor version for version dropdown href

sdowell:
A recent change updated the website to use patch versions in the
dropdown. The site URL still uses the minor version, so the href
needs to continue using the minor version.


mengqiy (approved):
LGTM

mengqiy:
This PR redoes #666 which was rolled back.

PR #674, opened 2021-12-10
fix: Correct path in href for contrib fn dropdown

sdowell:
Contrib functions are under the /contrib path, so this corrects
the path used in the href for contrib function version dropdowns.


mengqiy (approved):

mengqiy:
@sdowell This PR needs rebase

PR #680, opened 2021-12-14
Set namespace rolebinding behavior

sdowell:
Issues: GoogleContainerTools/kpt#2397

This pull request adds support to update the namespace for
ServiceAccount resources which are referenced in the subjects
field of RoleBinding and ClusterRoleBinding resources.

Given a ServiceAccount with name sa-name is included in the
ResourceList, any RoleBinding or ClusterRoleBinding subject with
kind ServiceAccount and name sa-name will also have its namespace
set.

For the time being this functionality is being implemented in kpt
rather than the upstream kustomize library. If this behavior is
well received, it may be a future consideration to push this
functionality into kustomize.

### DIFF
--- a/functions/go/set-namespace/README.md
+++ b/functions/go/set-namespace/README.md
@@ -37,19 +37,35 @@ target the namespace. There are a few cases that worth pointing out:
 
 - If there is a `Namespace` resource, its `metadata.name` field will be updated.
 - If there's a `RoleBinding` or `ClusterRoleBinding` resource, the function will
-  update the namespace in the `ServiceAccount` if and only if the subject
-  element `name` is `default`. In the following example, the `set-namespace`
-  function will update `subjects.namespace` since the
-  corresponding `subjects.name` is `default`.
+  update the namespace in the `ServiceAccount` if one of the following are true:
+  1) the subject element `name` is `default`.
+  2) the subject element `name` matches the name of a `ServiceAccount` resource declared in the package.
+  
+In the following example, the `set-namespace` function will update:
+- `subjects[0].namespace` since `subjects[0].name` is `default`.
+- `subjects[1].namespace` since `subjects[1].name` matches a `ServiceAccount`
+  name declared in the package.
 
 ```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: service-account
+  namespace: original-namespace
+---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   ...
 subjects:
   - kind: ServiceAccount
-    name: default # <======== using name default here
+    name: default # <================== name default is used
+    namespace: original-namespace # <== this will be updated
+  - kind: ServiceAccount
+    name: service-account # <========== name matches above ServiceAccount
+    namespace: original-namespace # <== this will be updated
+  - kind: ServiceAccount
+    name: other-service-account # <==== this will NOT be updated
     namespace: original-namespace
 roleRef:
   kind: Role
--- a/functions/go/set-namespace/namespace_transformer.go
+++ b/functions/go/set-namespace/namespace_transformer.go
@@ -9,9 +9,18 @@ import (
 
 	"sigs.k8s.io/kustomize/api/filters/namespace"
 	"sigs.k8s.io/kustomize/api/resmap"
+	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/filtersutil"
+	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
 	"sigs.k8s.io/yaml"
+)
+
+const (
+	subjectsField          = "subjects"
+	serviceAccountKind     = "ServiceAccount"
+	roleBindingKind        = "RoleBinding"
+	clusterRoleBindingKind = "ClusterRoleBinding"
 )
 
 // Change or set the namespace of non-cluster level resources.
@@ -48,6 +57,8 @@ func (p *plugin) Transform(m resmap.ResMap) error {
 	if len(p.Namespace) == 0 {
 		return nil
 	}
+	sal := serviceAccountLookup{}
+	sal.FromResMap(m)
 	for _, r := range m.Resources() {
 		if r.IsNilOrEmpty() {
 			// Don't mutate empty objects?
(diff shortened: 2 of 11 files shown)


functions/go/set-namespace/namespace_transformer.go
```diff
@@ -60,6 +71,67 @@ func (p *plugin) Transform(m resmap.ResMap) error {
...
+		return o.PipeE(
+			kyaml.LookupCreate(kyaml.ScalarNode, "namespace"),
+			kyaml.FieldSetter{Value: v},
+		)
+	})
+	return err
+}
+
+// ServiceAccountLookup provides an API for tracking ServiceAccount resources
+type ServiceAccountLookup struct {
```
droot, line 120: This struct doesn't need to be exported.

tests/set-namespace/rolebindings/Kptfile
```diff
@@ -0,0 +1,9 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: example
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/set-namespace:unstable
```
droot, line 7: it will be good to have a test case that involves `ensure-name-substring` function in the pipeline and ensure this change works well with that too.

droot (commented):
I think changes seems to be in the right direction. Have minor suggestions.

functions/go/set-namespace/namespace_transformer.go
```diff
@@ -60,6 +71,67 @@ func (p *plugin) Transform(m resmap.ResMap) error {
...
+// namespace for that ServiceAccount where it is referenced in the "subjects"
+// field of a RoleBinding or ClusterRoleBinding
+func (p *plugin) roleBindingHack(r *resource.Resource, sAL ServiceAccountLookup) error {
+	kind := r.GetKind()
+	if kind != roleBindingKind && kind != clusterRoleBindingKind {
+		return nil
+	}
+	subjects, err := r.Pipe(kyaml.Lookup(subjectsField))
+	if err != nil || kyaml.IsMissingOrNull(subjects) {
+		return err
```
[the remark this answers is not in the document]
sdowell: I believe so, and the conditional logic here is taken from the current [roleBindingHack](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/namespace/namespace.go#L109). Therefore with respect to this logic there should not be a deviation in behavior

If all else is valid but the subjects field is missing, then:
- `err` should be `nil`
- `kyaml.IsMissingOrNull(subjects)` will return `false`

Ultimately resulting in this line returning `nil`

functions/go/set-namespace/namespace_transformer.go
```diff
@@ -60,6 +71,67 @@ func (p *plugin) Transform(m resmap.ResMap) error {
...
+
+	err = subjects.VisitElements(func(o *kyaml.RNode) error {
+		subjectKind := o.GetKind()
+		if subjectKind != serviceAccountKind {
+			return nil
+		}
+		var nameRN *kyaml.RNode
+		nameRN, err = o.Pipe(kyaml.Lookup("name"))
+		if err != nil || kyaml.IsMissingOrNull(nameRN) {
+			return err
```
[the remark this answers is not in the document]
sdowell: This case is handled the same as the `subjects` field above, where if there is no error but the field is missing/null then it will return `nil`. I did not think it would be preferred to swallow the error here, but WDYT?

functions/go/set-namespace/namespace_transformer.go
sdowell: I was aiming to present this API in a way that could be easily migrated over to the shared library (kustomize), assuming this behavior is well received. In which case, assuming the API remains unchanged, would mean the struct would move to the shared library and be exported there.

However, it's a fair point that it does not need to be exported in this iteration - which is probably sufficient justification to not export it. I shall un-export it 🙂

tests/set-namespace/rolebindings/Kptfile
sdowell: Do you think that should be included in this test or as a separate test?

droot: Having a separate test case which explicitly tests the interplay of both the functions will be good.

droot (commented):
Thanks @sdowell Will defer to @mengqiy for final approval (I think he is back later this week).

tests/set-namespace/rolebindings/Kptfile
sdowell: Done!

functions/go/set-namespace/README.md
```diff
@@ -51,6 +61,12 @@ subjects:
   - kind: ServiceAccount
     name: default # <======== using name default here
     namespace: original-namespace
+  - kind: ServiceAccount
+    name: service-account # <======== name matches above ServiceAccount
+    namespace: original-namespace
+  - kind: ServiceAccount
+    name: other-service-account # <======== will NOT be updated
+    namespace: original-namespace
```
mengqiy, line 69: nit:
```yaml
subjects:
  - kind: ServiceAccount
    name: default # <=============== name default is used
    namespace: original-ns # <====== this will be updated
  - kind: ServiceAccount
    name: service-account # <======= name matches above ServiceAccount
    namespace: original-ns # <====== this will be updated
  - kind: ServiceAccount
    name: other-service-account
    namespace: original-ns # <====== this will NOT be updated
```

functions/go/set-namespace/namespace_transformer.go
```diff
@@ -60,6 +71,67 @@ func (p *plugin) Transform(m resmap.ResMap) error {
...
 	}
 	return nil
 }
+
+// roleBindingHack extends the behavior of the upstream kustomize filter
+// This extension adds the following behavior:
+// Given a ServiceAccount is present in the ResourceList, additionally set the
+// namespace for that ServiceAccount where it is referenced in the "subjects"
+// field of a RoleBinding or ClusterRoleBinding
+func (p *plugin) roleBindingHack(r *resource.Resource, sAL serviceAccountLookup) error {
```
mengqiy, line 86: > sAL serviceAccountLookup

Just use `sal` should be fine.

tests/set-namespace/ensure-name-substring/Kptfile
```diff
@@ -0,0 +1,12 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: example
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/ensure-name-substring:unstable
```
mengqiy, line 7: What are we trying to do here?
Are we trying to ensure they both update the `ServiceAccount` under `subjects`?

sdowell: This was in response to [this discussion](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/680#discussion_r769086646) with @droot, i.e.
>it will be good to have a test case that involves ensure-name-substring function in the pipeline and ensure this change works well with that too.

My Interpretation of the idea is as you say, to ensure they both update the `ServiceAccount` under `subjects`

functions/go/set-namespace/README.md
sdowell: Good suggestion, done!

functions/go/set-namespace/namespace_transformer.go
sdowell: Done!

tests/set-namespace/ensure-name-substring/Kptfile
mengqiy: Got it

mengqiy (approved):

PR #716, opened 2021-12-23
Set image function

sdowell:
This function is intended to be used for setting the image field
from a specified name to a specified newName:tag or newName@digest
for all resources. It uses the common defaults for image fields,
but additional image fields can be passed in via the
additionalImageFields parameter of the functionConfig.

This has been implemented using the new SDK, and is currently pinned to
a feature branch as the new SDK is not yet merged.

Issues: GoogleContainerTools/kpt#2577


functions/go/set-image/image_transformer_test.go
```diff
@@ -0,0 +1,271 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/kustomize/kyaml/fn/sdk"
+)
+
+// helper function to convert ResourceList items to yaml
+func itemsToYaml(items []*sdk.KubeObject) string {
```
sdowell, line 12: @mengqiy the new SDK exposes a `ToYaml` method for the `ResourceList` type, which includes the `FunctionConfig` in addition to the `Items`. For the purpose of testing expected output I figured it would be more simple/concise to exclude the `FunctionConfig` from the yaml and only include the `Items` (as is done in the unit tests for many of the other functions).

I was thinking perhaps we could push this logic up to the SDK by changing the type of `Items` from `[]*KubeObject` to a new type (e.g. `KubeObjects`) which could include its own `ToYaml` method. WDYT?

functions/go/set-image/image_transformer_test.go
```diff
@@ -0,0 +1,271 @@
...
+	var itemYamls []string
+	for _, item := range items {
+		itemYamls = append(itemYamls, item.String())
+	}
+	result := strings.Join(itemYamls, "---\n")
+	return result
+}
+
+func runImageTransformerE(input, config string) (string, error) {
```
sdowell, line 21: @mengqiy Let me know if there is a better way to handle this unit test scaffolding in the new SDK.

Given this is along the right track, I was thinking we might consider implementing a similar layer in the SDK to streamline the experience of writing tests and reduce code duplication.

functions/go/set-image/README.md
```diff
@@ -0,0 +1,120 @@
...
+the path `spec/manifest/images[]/image` in `MyKind` resource, we use the
+following `functionConfig`:
+
+```yaml
+apiVersion: fn.kpt.dev/v1alpha1
+kind: SetImage
+metadata:
+  name: my-func-config
+  annotations:
+    config.kubernetes.io/local-config: "true"
```
droot, line 103: This annotation should not be needed, so better to omit it.

examples/set-image-advanced/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+```yaml
+# fn-config.yaml
+apiVersion: fn.kpt.dev/v1alpha1
+kind: SetImage
+metadata:
+  name: my-func-config
+  annotations:
+    config.kubernetes.io/local-config: "true"
```
droot, line 28: This annotation should not be needed.

droot (commented):
Some drive by comments. This function addresses a common need for k8s applications. Overall looking good, will leave @mengqiy to approve this.

functions/go/set-image/README.md
sdowell: Done

examples/set-image-advanced/README.md
sdowell: Done

functions/go/set-image/image_transformer_test.go
mengqiy: Have you looked at http://104.197.156.154/pkg/sigs.k8s.io/kustomize/kyaml/fn/sdk/testutil/?
That package currently have 3 methods and they assume that golden tests will be used. Maybe we should add another helper that take inline input.

sdowell: Yeah I did take a look at the `testutil` package

I felt that creating a separate file/directory for each test case would become a bit cumbersome in this instance, so I opted to create this function which accepts string input instead. I was thinking as you say that we could add something like this as another helper which accepts inline (string) input. WDYT?

functions/go/set-image/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Usage
+
+This function can be used with any KRM function orchestrators (e.g. kpt).
+
+For all resources which specify an image of a given `name`, the `set-image`
+function sets that image field to the specified `newName:newTag` or
+`newName@digest`.
+
+By default the function updates the standard [fields][commonimage] used to
```
mengqiy, line 26: Should we update the images in Kptfile?

functions/go/set-image/README.md
```diff
@@ -0,0 +1,118 @@
...
+kind: ConfigMap
+metadata:
+  name: my-config
+data:
+  name: nginx
+  newName: nginx
+  digest: sha256:3cbbd11b65aab276c8578c039d0c21d0ffb7a496e09c0f632bac1a1b2c115256
+```
+
+To use a `SetImage` custom resource as the `functionConfig`, the desired
```
mengqiy, line 73: Per https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/_template/README.md#functionconfig, We should describe the fields more precisely.

functions/go/set-image/README.md
```diff
@@ -0,0 +1,118 @@
...
+  omitted.
+- `kind`: Select the resources by resource kind. Will select all kinds if
+  omitted.
+- `path`: Specify the path to the field that the value needs to be updated. This
+  field is required.
+- `create`: If it's set to true, the field specified will be created if it
+  doesn't exist. Otherwise, the function will only update the existing field.
+
+To set image `nginx` to `bitnami/nginx:1.21.4` for all built-in resources and
+the path `spec/manifest/images[]/image` in `MyKind` resource, we use the
```
mengqiy, line 94: We should mention the syntax for the path. I guess not everyone knows what `images[]` means

functions/go/set-image/go.mod
```diff
@@ -0,0 +1,12 @@
...
+go 1.16
+
+require (
+	github.com/stretchr/testify v1.7.0
+	sigs.k8s.io/kustomize/api v0.10.1
+	sigs.k8s.io/kustomize/kyaml v0.13.0
+
+)
+
+replace sigs.k8s.io/kustomize/kyaml v0.13.0 => github.com/mengqiy/kustomize/kyaml v0.15.0
```
mengqiy, line 12: I have move the fork. You can change it to https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/export-terraform/go.mod#L6

functions/go/set-image/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+There are 2 kinds of `functionConfig` supported by this function:
+
+- `ConfigMap`
+- A custom resource of kind `SetImage`
+
+To use a `ConfigMap` as the `functionConfig`, the desired image spec must be
+specified using the following fields:
+- `data.name`
+- `data.newName`
```
mengqiy, line 41: If I don't want to update the image name, do I still need to specify this field?

sdowell: In the current implementation, yes you still need to supply a `newName`. However to leave the name unchanged you can specify the same value for both `name` and `newName`

functions/go/set-image/README.md
sdowell: I am not sure whether or not that would be a desirable default behavior

The user could of course always supply this as an option in `additionalImageFields`, although that would be less intuitive.

I could also see an argument for a separate function specifically for the use case of updating Kptfiles. I could see this as a function which automatically discovers the latest patch version for images specified in Kptfiles and update them dynamically based on what was fetched

mengqiy: > I could also see an argument for a separate function specifically for the use case of updating Kptfiles. I could see this as an imperative function which automatically discovers the latest patch version for images specified in Kptfiles and update them dynamically based on what was fetched

No doubt of that. But IMO they are not mutually exclusive. The set-image function can be a supplement to help users to achieve some one-off operations that can't be done by the update-to-latest-image-in-kptfile function. People may argue that one-off operations can be done by manual in-place edit.

If we support updating images in Kptfile in this PR, user will need to specify `eval --include-meta-resources` to make it work, since eval doesn't include Kptfile by default and render doesn't support it.

I don't feel very strong about it and I'd like to hear others' opinions. @droot @phanimarupaka Thoughts?

functions/go/set-image/README.md
mengqiy: In the current design, `newName` seems to be a required field, but IMO it doesn't have to be.
I guess we can make it optional: if it's specified, we update the image. If not, we keep the image name unchanged.

sdowell: This is a good point, and has me rethinking the input validation that I've put in place here.. I put the input validation in hopes of providing a more intuitive UX, however upon reconsideration it seems the default behavior of the kyaml filter may be desirable here. 

For reference, here is the default behavior of the kyaml filter:
`name`: the original name used to identify which images to replace. If omitted the operation becomes a noop
`newName`: the new name to set for images matching `name`. If omitted the name will be unchanged
`newTag`: the new tag to set for images matching `name`. If omitted the tag will be unchanged
`digest`: the new digest to set for images matching `name`. if omitted the digest will be unchanged. If provided, the `digest` will be used rather than `newTag`

I think we could largely just rely on the default behavior here, but I'd like to suggest a few exceptions which we can validate ourselves:
- Make `name` a required field, since the function becomes a noop without a provided `name`
- Allow `newName`, `newTag`, and `digest` to be optional. But return a validation error if none are provided since the function will become a noop
- Return a validation error if both a `newTag` and a `digest` are provided. If a `newTag` and `digest` are provided then the kyaml filter will silently ignore the `newTag` and use `digest`, which seems like a potentially confusing function UX.

WDYT?

mengqiy: The default behavior of the kyaml filter + the validation logic you suggested SGTM

functions/go/Makefile
```diff
@@ -19,6 +19,7 @@ GOBIN := $(shell go env GOPATH)/bin
 
 # Edit this list to contain all go functions
 FUNCTIONS := \
+	set-image \
```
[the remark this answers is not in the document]
sdowell: Thanks for pointing this out - I did not realize this on my first pass through. Done!

functions/go/set-image/go.mod
sdowell: Done

functions/go/set-image/README.md
sdowell: Done

functions/go/set-image/README.md
sdowell: I've updated some of the descriptions here - let me know if you find them to be more precise now. Note that I based this doc off of similar docs from the other functions, e.g.
https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/set-annotations/README.md
https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/set-namespace/README.md
etc.

functions/go/set-image/README.md
sdowell: Done

functions/go/set-image/README.md
droot: I think extending the set-image logic to update images in `Kptfile` sounds like a good enhancement. But we can address this as a follow-up.

functions/go/set-image/README.md
```diff
@@ -0,0 +1,124 @@
...
+### FunctionConfig
+
+There are 2 kinds of `functionConfig` supported by this function:
+
+- `ConfigMap`
+- A custom resource of kind `SetImage`
+
+To use a `ConfigMap` as the `functionConfig`, the desired image spec must be
+specified using the following fields:
+- `data.name`: Original name of image to replace. This field is required.
```
mengqiy, line 40: We should point out that the image name should not contain tag or digest.

functions/go/set-image/image_transformer.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Config initializes SetImage from a functionConfig sdk.KubeObject
+func (si *SetImage) Config(functionConfig *sdk.KubeObject) error {
+	si.Image = types.Image{}
+	si.AdditionalImageFields = nil
+	switch {
+	case validGVK(functionConfig, "v1", "ConfigMap"):
+		functionConfig.GetOrDie(&si.Image, "data")
+	case validGVK(functionConfig, fnConfigAPIVersion, fnConfigKind):
+		functionConfig.AsOrDie(si)
```
mengqiy, line 66: I'd suggest not suggest using *OrDie in our catalog. Reasons are we can't guarantee that it will always succeed and we should be more conservative. And we should provide a little more context in the error e.g. "unable to get field data: %w" and "unable to convert the functionConfig to ...: %w".

functions/go/set-image/README.md
sdowell: Done

functions/go/set-image/image_transformer.go
sdowell: Good suggestion - done

mengqiy (approved):

functions/go/set-image/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-image
+
+go 1.16
```
mengqiy, line 3: Let's use go 1.17

functions/go/set-image/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.16-alpine3.13
```
mengqiy, line 1: Let's use go 1.17

functions/go/set-image/go.mod
sdowell: Done

functions/go/set-image/Dockerfile
sdowell: Done

PR #719, opened 2021-12-29
feat: set-namespace supports depends-on annotation

sdowell:
The depends-on annotation can reference the namespace of other
resources, and this change makes sure that the namespace portion of such
annotations is properly handled by the set-namespace function.

Issues: GoogleContainerTools/kpt#2579


functions/go/set-namespace/namespace_transformer.go
```diff
@@ -74,6 +83,42 @@ func (p *plugin) Transform(m resmap.ResMap) error {
...
+	namespaceIdx := 2
+	segments[namespaceIdx] = p.Namespace
+	dependsOn = strings.Join(segments, "/")
+	return dependsOn, true
+}
+
+// dependsOnHack updates the namespace for the depends-on annotation if the
+// annotation is for a namespaced resource. The expected syntax for a namespaced
+// resource is <group>/namespaces/<namespace>/<kind>/<name>
+func (p *plugin) dependsOnHack(r *resource.Resource) error {
```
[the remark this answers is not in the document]
sdowell: Sure thing 😁 I did not intend to add a negative connotation - I encountered a few similar-looking edge case handlers which were also called "hacks", but this one very well may be a different pattern

mortent:
This change updates the namespace for all depends-on annotations without verifying whether the resource being referenced is also being updated. Are there any situations where the namespace of only a subset of resources are changed, and therefore we should verify before updating the reference?

sdowell:
> This change updates the namespace for all depends-on annotations without verifying whether the resource being referenced is also being updated. Are there any situations where the namespace of only a subset of resources are changed, and therefore we should verify before updating the reference?

@mortent This is a good suggestion. My understanding is that (from the perspective of the function) the only way we can be sure that a given resource will have its namespace updated is if that resource is also in the same kpt package. A solution I might suggest is that we check that the resource being referenced is in also included in the input ResourceList. To do this we can look up the resource by `group`/`kind`/`name`. WDYT?

Conceptually this approach would be similar to this solution for updating the namespace for service accounts referenced in role bindings: https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/680

sdowell:
@mortent I've pushed a commit which implements the behavior described above

functions/go/set-namespace/namespace_transformer.go
```diff
@@ -71,19 +90,63 @@ func (p *plugin) Transform(m resmap.ResMap) error {
...
+	return nil
+}
+
 // roleBindingHack extends the behavior of the upstream kustomize filter
 // This extension adds the following behavior:
 // Given a ServiceAccount is present in the ResourceList, additionally set the
 // namespace for that ServiceAccount where it is referenced in the "subjects"
 // field of a RoleBinding or ClusterRoleBinding
-func (p *plugin) roleBindingHack(r *resource.Resource, sal serviceAccountLookup) error {
+func (p *plugin) roleBindingHack(r *resource.Resource) error {
```
mengqiy, line 149: As Sunil suggested earlier, let's avoid using `hack` in the method name.

mengqiy (commented):
Code LGTM.
Please update the usage doc and add an example for this use case.

functions/go/set-namespace/namespace_transformer.go
sdowell: Sounds good, I can update this

However, just as a reference this method is intended as an extension of the behavior of the [upstream roleBindingHack](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/namespace/namespace.go#L101)

mengqiy:
As I mentioned in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/719#pullrequestreview-855921220, we should mention this behavior in function's usage doc.

sdowell:
> As I mentioned in [#719 (review)](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/719#pullrequestreview-855921220), we should mention this behavior in function's usage doc.

@mengqiy My mistake I saw the example part but missed the usage doc part of that comment. I'll update the usage doc now

functions/go/set-namespace/README.md
```diff
@@ -50,23 +53,27 @@ In the following example, the `set-namespace` function will update:
...
     name: default # <================== name default is used
-    namespace: original-namespace # <== this will be updated
+    namespace: ns # <================== this will be updated
   - kind: ServiceAccount
-    name: service-account # <========== name matches above ServiceAccount
-    namespace: original-namespace # <== this will be updated
+    name: sa # <======================= name matches above ServiceAccount
+    namespace: ns # <================== this will be updated
   - kind: ServiceAccount
     name: other-service-account # <==== this will NOT be updated
```
mengqiy, line 75: > # <==== this will NOT be updated

This comment actually belongs to the next line, right?

mengqiy (commented):
One minor issue, LGTM otherwise

functions/go/set-namespace/README.md
sdowell: Looks like you are correct here - I will fix this up

mengqiy (approved):
LGTM

PR #720, opened 2021-12-30
feat: ensure-name-substring supports depends-on

sdowell:
The depends-on annotation can reference the name of other resources,
and this change makes sure that the name portion of such annotations
is properly handled by the ensure-name-substring function.

Issues: GoogleContainerTools/kpt#2579


mengqiy (approved):
LGTM

PR #721, opened 2021-12-30
Update usage doc for starlark

mengqiy:
Include how to access env var and openapi


sdowell (approved):
Are these `docs.go` files meant to be updated whenever any of the docs (e.g. `README`s) change?

mengqiy:
Yes.
But it seems https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/.github/workflows/check-doc-sync.yaml is not doing its job. It is supposed to work the same as https://github.com/GoogleContainerTools/kpt/blob/main/.github/workflows/checkDocSync.yml.
If you have some bandwidth, can you please take a look?

PR #722, opened 2022-01-06
feat(docs-fn): support explicit pkg name, handle empty pkg

bharathkkb:
This adds support for specifying an optional pkg name while generating documentation to support blueprint repo dir naming convention https://github.com/GoogleContainerTools/kpt/issues/2592. Also handles cases where a pkg may be empty.


mengqiy:
Do we need to update the usage doc for this? I don't see it got updated.

bharathkkb:
@mengqiy done

mengqiy (approved):

PR #735, opened 2022-01-11
docs: Dynamically render latest image tag in docs

sdowell:
This change displays the latest image tag for the function in the
function doc page. This is intended as a convenience for users
to discover the full tag of the latest function image.


mengqiy (approved):

mengqiy:
Thanks for the quick fix

PR #739, opened 2022-01-13
feat: Rewrite apply-time-mutation function

karlkfi:
- Add support for custom resource object syntax
- Update to use cli-utils as the SDK for apply-time-mutations
- Add more tests
- Shorten package name to 'annotate-apply-time-mutation/pkg'
- Use sigs.k8s.io/yaml for yaml formatting, to avoid a few bugs in
  sigs.k8s.io/kustomize/kyaml/yaml

Updated readme: https://github.com/karlkfi/kpt-functions-catalog/blob/karl-mutation-crd/contrib/functions/go/annotate-apply-time-mutations/README.md

Build container:
```
cd contrib/functions/go
make func-build CURRENT_FUNCTION=annotate-apply-time-mutations TAG=unstable
```

Run e2e test:
```
cd tests/e2etest
go test -v ./... -run TestE2E/../../contrib/examples/annotate-apply-time-mutations-custom-resource
go test -v ./... -run TestE2E/../../contrib/examples/annotate-apply-time-mutations-inline-comment
```


karlkfi:
@thackem Can you take a look, since you wrote the original?

contrib/functions/go/annotate-apply-time-mutations/README.md
```diff
@@ -30,36 +30,137 @@ pipeline:
...
+content from the source field. When specified, the function will copy them into
+the field itself, surrounding a generated token.
 
-1.  Scans the package for `apply-time-mutation` comment markup.
-2.  Appends the equivalent `config.k8s.io/apply-time-mutation` annotation to the same.
+```
+field: [prefix]$ref[suffix] # apply-time-mutation: ...
+```
+
+`version` is also optional. It's generally recommended to avoid specifying the
```
mengqiy, line 72: When `version` is omitted, should I use remove one `/`. IOW, `[group]//namespaces/[source-namespace]/[kind]/[source-name]` or `[group]/namespaces/[source-namespace]/[kind]/[source-name]`.
Please clarify.

contrib/functions/go/annotate-apply-time-mutations/go.mod
```diff
@@ -2,30 +2,61 @@ module github.com/GoogleContainerTools/kpt-functions-catalog/contrib/functions/g
 
 go 1.17
 
-require sigs.k8s.io/kustomize/kyaml v0.11.1
```
mengqiy, line 5: Have you run `go mod tidy`?
In go 1.17 and later, direct and transitive dependencies are in 2 `require` stanzas.
It seems you have mixed them together.

contrib/functions/go/annotate-apply-time-mutations/pkg/types.go
```diff
@@ -0,0 +1,25 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/cli-utils/pkg/object/mutation"
+)
+
+// ApplyTimeMutation is a Kubernetes resource that allows specifying mutations
+// using a seperate KRM object, instead of an annotation string on the target
+// object.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type ApplyTimeMutation struct {
+	// TODO: Use v1.TypeMetaApplyConfiguration from client-go instead?
```
mengqiy, line 13: Are there any difference between them?

contrib/examples/annotate-apply-time-mutations-inline-comment/Kptfile
```diff
@@ -0,0 +1,4 @@
+apiVersion: kpt.dev/v1
```
mengqiy, line 1: This is an empty Kptfile, do we really need it?

Same comment for `annotate-apply-time-mutations-crd/Kptfile`

mengqiy:
I'd suggest someone who is very familiar with the design to review the logic. @thackem may be a candidate.

contrib/examples/annotate-apply-time-mutations-inline-comment/Kptfile
karlkfi: Will the example tests run without it? I just copied what was already there.

contrib/functions/go/annotate-apply-time-mutations/go.mod
karlkfi: Ah. I was confused by that change. It wasnt' clear why there's two blocks.

contrib/functions/go/annotate-apply-time-mutations/pkg/comment_scanner.go
```diff
@@ -0,0 +1,180 @@
...
+
+type ScanResult struct {
+	Path         string
+	Value        interface{}
+	Comment      string
+	Substitution mutation.FieldSubstitution
+}
+
+// Scan searches for mutation markup comments and parses them as substitutions.
+func (cs *CommentScanner) Scan(obj *yaml.RNode, results map[string]ScanResult, fieldPath ...interface{}) error {
```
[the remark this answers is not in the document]
karlkfi: It's used to build a JSONPath expression recursively. The Substitution spec requires JSONPath.

`framework.Field.Path` in `framework.Result.Field` also requires a string, tho it might not actually be JSONPath. It seems to be a concatenation of map keys delimited by periods without any list indexes... That format seems pretty useless, but it's not clear if it's used for anything other than printing to the user.

contrib/functions/go/annotate-apply-time-mutations/pkg/types.go
karlkfi: Looks like the -ApplyConfiguration resources in client-go are purely for server-side apply, making every field a pointer. Since ApplyTimeMutation never gets sent to the server, we don't need server-side apply. So I can remove the TODO.

contrib/functions/go/annotate-apply-time-mutations/pkg/comment_scanner.go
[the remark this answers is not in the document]
karlkfi: Digging deeper, the Field.Path in the framework is exclusively used to generate stringified user output, not for looking up the field in any way. So the format of the Path is probably irrelevant, except for consistency. And since the convention sucks (always  period delimited, even if the field has period or quote or bracket in it... and skips all list indexes...) I think it's probably fine to just set it to the JSONPath value I need for the apply-time-mutation annotation anyway.

But because you asked nicely, I wrote a simple PathBuilder to pass around and generate the JSONPath on-demand instead of the vararg of interface{}. It also incidentally simplifies the errors and logging with path in them.

contrib/examples/annotate-apply-time-mutations-inline-comment/Kptfile
karlkfi: Tests pass without them.

karlkfi:
@thackem is out of office. Requesting review from @mortent and/or @haiyanmeng instead, since they have context on apply-time-mutation.

mengqiy:
@mortent @haiyanmeng PTAL

contrib/functions/go/annotate-apply-time-mutations/pkg/object_scanner_test.go
```diff
@@ -0,0 +1,75 @@
...
+							TargetPath: "$.spec.data",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run("", func(t *testing.T) {
```
mortent, line 66: Seems like we only have one testcase here so it doesn't matter that much, but we might want to add proper names to the subtests.

mortent (approved):
Only a small comment, but looks good to me.

contrib/functions/go/annotate-apply-time-mutations/pkg/object_scanner_test.go
karlkfi: fixed.

mengqiy (approved):

PR #740, opened 2022-01-14
upgrade golang 1.17

mengqiy:
Our release pipeline for `unstable` tag is not working, since one of the function requires go 1.17.
This PR does a bulk update.

### DIFF
--- a/contrib/functions/go/generate-kpt-pkg-docs/Dockerfile
+++ b/contrib/functions/go/generate-kpt-pkg-docs/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.16-alpine3.13
+FROM golang:1.17-alpine3.13
 ENV CGO_ENABLED=0
 WORKDIR /go/src/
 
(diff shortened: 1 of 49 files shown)


contrib/functions/go/generate-kpt-pkg-docs/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.16-alpine3.13
+FROM golang:1.17-alpine3.13
```
sdowell, line 1: What would you think about defining common Dockerfile(s) for all of the functions? It seems like if our intent is to keep the dependencies (e.g. language version, base image, etc) consistent, then it might make this easier to manage/maintain if we reused common Dockerfile(s).

Another option might be to pass in the base image as an argument to the Dockerfile(s), so that we can define the value(s) in one place

I'd say this suggestion is out of scope for this issue, but I was just curious of your thoughts

sdowell (approved):
LGTM with a passing comment for you to consider

PR #744, opened 2022-01-19
Common go dockerfile

sdowell:
This logic gives the option to not define a Dockerfile for a given
function. If the function does not have its own Dockerfile defined, then
it will reuse the common Dockerfile.

Use cases which do not fit the pattern of the common Dockerfile maintain
the option to define their own Dockerfile (e.g. render-helm-chart)


mengqiy:
https://github.com/GoogleContainerTools/kpt-functions-catalog/blame/master/CONTRIBUTING.md#L67-L73 will need to be updated.
Why we don't do it for TS functions?

sdowell:
> https://github.com/GoogleContainerTools/kpt-functions-catalog/blame/master/CONTRIBUTING.md#L67-L73 will need to be updated. Why we don't do it for TS functions?

@mengqiy I looked at adding this for TS as well but felt that I didn't see a strong pattern across the TS Dockerfiles that could be generalized. They each seem to have their own one-off dependencies that they need to install in the image. WDYT?

I'll update that section of the contributing doc - thank you for pointing that out

mengqiy:
We have a template Dockerfile in https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/ts/create-kpt-functions/templates/Dockerfile.mustache. Some functions add their own dependencies on top of it.
`generate-folders` is mostly using the template.
I guess we can the template as the shared the Dockerfile. But one thing we should figure out is the entrypoints are different across the ts functions since the script names is different.

sdowell:
@mengqiy I've updated the PR to include a common dockerfile for TS as well. For now this is only applied to the `generate_folders` function

scripts/docker.sh
```diff
@@ -0,0 +1,75 @@
...
+
+  case "${type}" in
+    contrib) function_dir="${repo_base}/contrib/functions/${lang}/${name}" ;;
+    curated) function_dir="${repo_base}/functions/${lang}/${name}" ;;
+    *) err "unknown function type: ${type}" ;;
+  esac
+
+  case "${lang}" in
+    ts)
+      translated_name=$(echo "${name}" | tr - _)
```
mengqiy, line 43: We won't need to do it if we address https://github.com/GoogleContainerTools/kpt/issues/2533.
What is the status of https://github.com/GoogleContainerTools/kpt/issues/2533?

CONTRIBUTING.md
```diff
@@ -56,7 +56,12 @@ Contributions are required to follow these style guides:
 │     │     └── ts: Home for all typescript-based contrib function source code
 │     └── examples: Home for all contrib function examples.
 ├── scripts
-└── tests: Home for e2e tests
+├── tests: Home for e2e tests
+└── docker
```
mengqiy, line 60: I'd suggest to add them under `build/docker/`. If we have more stuff related to build in the future, we can put them in `build`.

sdowell: Done

scripts/docker.sh
sdowell: I have not started on 2533 yet, although I do agree that resolving 2533 would simplify this logic

mengqiy: Got it

mengqiy (approved):

PR #747, opened 2022-01-25
render-helm-chart function changes

natasha41575:
- Support multiple values files in `RenderHelmChart` functionConfig (https://github.com/kubernetes-sigs/kustomize/issues/4219)
- Make the `repo` field optional to support authentication with private repositories (https://github.com/kubernetes-sigs/kustomize/issues/4335)
- Support other requested options for [`helm template`](https://helm.sh/docs/helm/helm_template/), such as [`api-versions`](https://github.com/kubernetes-sigs/kustomize/issues/3816), `name-template`,  and [`skip-tests`](https://github.com/kubernetes-sigs/kustomize/issues/3831). 

EDIT: Removing `create-namespace` and `skip-crds` because I'd misunderstood what they do; I don't think they will provide value here as IIUC they involve communication with the cluster. 

This is a breaking change, and will require a version bump.


functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -135,12 +133,28 @@ func (p *HelmChartInflationGeneratorPlugin) runHelmCommand(
...
 
-// createNewMergedValuesFile replaces/merges original values file with ValuesInline.
-func (p *HelmChartInflationGeneratorPlugin) createNewMergedValuesFile() (
-	path string, err error) {
+// createNewMergedValuesFiles replaces/merges original values file with ValuesInline.
+func (p *HelmChartInflationGeneratorPlugin) createNewMergedValuesFiles(path string) (
+	string, error) {
+	pValues, err := ioutil.ReadFile(path)
+	if err != nil {
+		if u, err := url.Parse(path); err == nil && (u.Scheme == "http" || u.Scheme == "https") {
```
mengqiy, line 141: If it's a valid URL, but the scheme is not supported, we should return an error.

If it's neither a local path nor a valid URL, we should return an error.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -135,12 +133,28 @@ func (p *HelmChartInflationGeneratorPlugin) runHelmCommand(
...
+			var hc *http.Client
+			hc = &http.Client{}
+			resp, err := hc.Get(path)
```
mengqiy, line 144: No need to create a client here. You can simply do `http.Get(path)`

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -172,39 +183,16 @@ func (p *HelmChartInflationGeneratorPlugin) replaceValuesInline() error {
...
-	b []byte) (string, error) {
+	b []byte, path string) (string, error) {
 	if err := p.establishTmpDir(); err != nil {
 		return "", fmt.Errorf("cannot create tmp dir to write helm values")
 	}
-	path := filepath.Join(p.tmpDir, p.Name+"-kustomize-values.yaml")
-	return path, ioutil.WriteFile(path, b, 0644)
+	// use a hash of the provided path to generate a unique, valid filename
+	hash := md5.Sum([]byte(path))
+	newPath := filepath.Join(p.tmpDir, p.Name+"-"+hex.EncodeToString(hash[:])+"-kustomize-values.yaml")
```
mengqiy, line 194: Why do we have the hash in the middle? I guess use it as the end is better.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -172,39 +183,16 @@ func (p *HelmChartInflationGeneratorPlugin) replaceValuesInline() error {
...
 func (p *HelmChartInflationGeneratorPlugin) writeValuesBytes(
-	b []byte) (string, error) {
+	b []byte, path string) (string, error) {
 	if err := p.establishTmpDir(); err != nil {
 		return "", fmt.Errorf("cannot create tmp dir to write helm values")
 	}
-	path := filepath.Join(p.tmpDir, p.Name+"-kustomize-values.yaml")
-	return path, ioutil.WriteFile(path, b, 0644)
+	// use a hash of the provided path to generate a unique, valid filename
+	hash := md5.Sum([]byte(path))
```
mengqiy, line 193: Is the path an absolute path?

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: done

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: done

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: moved it!

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: Yes, it is an absolute path in the `/tmp` directory.

natasha41575:
cc @yuwenma

examples/render-helm-chart-local/README.md
```diff
@@ -37,6 +37,17 @@ releaseName=test \
 valuesFile=tmp/charts/helloworld-values/values.yaml
```
mengqiy, line 37: Should it start with `/tmp`?

examples/render-helm-chart-local/README.md
```diff
@@ -37,6 +37,17 @@ releaseName=test \
...
 ```
 
+You can optionally skip tests in the templated output with `skipTests`.
+
+```shell
+$ kpt fn eval --image gcr.io/kpt-fn/render-helm-chart:unstable \
+--mount type=bind,src=$(pwd),dst=/tmp/charts -- \
+name=helloworld-chart \
+releaseName=test \
+valuesFile=tmp/charts/helloworld-values/values.yaml \
```
mengqiy, line 47: Same here.
Please audit the examples to ensure we are using `/tmp`

functions/go/render-helm-chart/README.md
```diff
@@ -86,34 +90,44 @@ helmGlobals:
...
-`includeCRDs`  | Specifies if Helm should also generate CustomResourceDefinitions. Defaults to false. | true
-
-The only required field is `name`.
+| Field           |  Description | Example
+| -----------:    |  ----------- | -----------
+`helmGlobals`     | Parameters applied to all Helm charts
+`helmCharts`      | An array of helm chart parameters
+`chartArgs`       | Arguments that describe the chart being rendered. 
+`templateOptions` | A collection of fields that map to flag options of `helm template`. 
+`chartHome`       | A filepath to a directory of charts. The function will look for the chart in this local directory before attempting to pull the chart from a specified repo. Defaults to "tmp/charts". When run in a container, this path MUST have the prefix "tmp/". | tmp/charts
```
mengqiy, line 117: Same comment here. `tmp/` or `/tmp`?

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
mengqiy: It seems this path is originally passed in by the users. What if they happen to provide a relative path? A safer approach is to resolve to absolute path before compute the hash.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -172,39 +187,16 @@ func (p *HelmChartInflationGeneratorPlugin) replaceValuesInline() error {
...
+	newPath := filepath.Join(p.tmpDir, p.Name+"-kustomize-values-"+hex.EncodeToString(hash[:])+".yaml")
```
mengqiy, line 198: Is `hex.EncodeToString(md5.Sum(input))` a common practice in kustomize?

I guess `hex.EncodeToString(hash[:])` can be simplified to `hex.EncodeToString(hash)`, right?

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -135,12 +133,32 @@ func (p *HelmChartInflationGeneratorPlugin) runHelmCommand(
...
+				if err != nil {
+					return "", err
+				}
+				defer resp.Body.Close()
+				pValues, err = ioutil.ReadAll(resp.Body)
+				if err != nil {
+					return "", err
+				}
+			} else { // url scheme is not http or https
+				return "", fmt.Errorf("unsupported url type: %s", path)
```
mengqiy, line 153: To be more accurate, let's use `unsupported URL scheme`.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -135,12 +133,32 @@ func (p *HelmChartInflationGeneratorPlugin) runHelmCommand(
...
+			}
+		} else { // invalid path and invalid URL
+			return "", err
```
mengqiy, line 156: `err` is the error when trying to read it as local file. This error can be misleading when it's an invalid URL.
We can return something like `fmt.Errorf("unable to read %q as URL or file path: %v when reading it as a file path; %v when reading it as a URL", path, err, urlErr)`. What I give you here can be a starting point. You can further polish it. The error message should provide enough context and actionable.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: It is not something we do in kustomize, but the hasher we have in kustomize hashes an entire RNode rather than a string, the code is here: https://github.com/kubernetes-sigs/kustomize/blob/master/api/hasher/hasher.go

Alternatives here could be to use the kustomize's hasher and do something like `hasher.SortArrayAndComputeHash([]string{path})` - but we are not using this anywhere in kustomize so I am unsure how reliable it is. Or try something like the [kubernetes deep hasher object](https://github.com/kubernetes/kubernetes/blob/master/pkg/util/hash/hash_test.go). 

However I chose to use `hex.EncodeToString(md5.Sum(input))` because it seemed like the simplest thing to do here. 

> I guess hex.EncodeToString(hash[:]) can be simplified to hex.EncodeToString(hash), right?

The latter doesn't compile, because `hash` is type `[16]byte` but `hash[:]` is type `[]byte`.

functions/go/render-helm-chart/README.md
natasha41575: For `chartHome`, the default is `tmp/`. IIUC when run in a container, the code is running at the root so it becomes the same thing as `/tmp`. However for `chartHome` it is important to distinguish, because when run locally (i.e. `exec`) it will look for files in `tmp/charts` by default, not `/tmp/charts`. I have updated the containerized `valuesFiles` examples to use `/tmp` to be explicit, but `chartHome` is `tmp` and not `/tmp` intentionally.

examples/render-helm-chart-local/README.md
natasha41575: done

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: done

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: done, thank you for the suggestion

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: Ah, my mistake, I thought you meant the path we are returning here. Fixed, we are now passing in the absolute path into `writeValuesBytes` which is used to generate the hash.

mengqiy (approved):

PR #749, opened 2022-01-26
Imagetag track mutations

sdowell:
Add support to track which images are updated by the set-image function.

Leverages new functionality from the underlying `imagetag` filter from kustomize to support the functionality


examples/set-image-advanced/.expected/results.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kpt.dev/v1
+kind: FunctionResultList
+metadata:
+  name: fnresults
+exitCode: 0
+items:
+  - image: gcr.io/kpt-fn/set-image:unstable
+    exitCode: 0
+    results:
+      - message: set image nginx:1.20.2 to bitnami/nginx:1.21.4 for 2 field(s)
```
mengqiy, line 10: We are showing a summary (# of the fields that has been updated).
Should we actually show each place (GVKNN + fields path) that has been changed?
Also we should set severity to info.

sdowell: The main reason I opted for this UX is that there is a limitation when iterating over a `SequenceNode` (e.g. `spec.containers[]`), where the `FieldPath` of any nested nodes treat the `SequenceNode` as the root.
e.g.
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: app
spec:
  containers:
  - name: store
    image: redis:6.2.6
```
The `FieldPath` of the `RNode` for `image` here would be `image` rather than something you might expect like `spec.containers[].image`. I opted to display this summary to provide a more intuitive UX given this limitation.

This is actually something that impacts `set-annotations` and `set-labels` as well, but we haven't seen this yet since no test cases involve a fieldspec including a `SequenceNode`. However, this behavior quickly arises for `set-image` since a default fieldspec includes a path with `SequenceNode`.

To change this would require some updates to the relevant behavior in kustomize - do you think we should change this behavior to display the full `FieldPath` past a `SequenceNode`?

Here is an example output:
```bash
$ kpt fn render set-image-simple
Package "set-image-simple": 
[RUNNING] "gcr.io/kpt-fn/set-image:unstable"
[PASS] "gcr.io/kpt-fn/set-image:unstable" in 1.5s
  Results:
    [info]: set image nginx:1.20.2 to bitnami/nginx:1.21.4 for 2 fields

Successfully executed 1 function(s) in 1 package(s).
```
Severity already seems to be `info`

mengqiy: The behavior with SequenceNode seems to be a bug. Can you please file an issue in kyaml? If it's a bug, can you check this is a bug that can be fixed within 1 hour? If so, that'd be great if you can fix it. Otherwise, we can keep the current behavior in this PR.

sdowell: Fixed by this PR: https://github.com/kubernetes-sigs/kustomize/pull/4425

I've updated this PR to consume the upstream changes and have changed the UX to include GVK + FieldPath. Let me know what you think

functions/go/set-image/image_transformer.go
```diff
@@ -39,6 +43,53 @@ type SetImage struct {
...
+		return results
+	}
+	for k, v := range si.setImageResults {
+		imageJson, err := json.Marshal(v)
+		if err != nil {
+			return nil
+		}
+		resourceRef := k.ResourceRef
+		results = append(results, &sdk.Result{
+			Message:     fmt.Sprintf("set image: %s", imageJson),
```
mengqiy, line 85: imageJson is something like `{old-val: new-val}`. If we always only have one key-value pair, why do we use it in json format. Can we do something like `set image from %q to %q`?

sdowell: It does not always have one key-value pair, see this unit test:
https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/749/files#diff-b85391f699436541ad10a25bf2aeb38cac0e0a1c0969482bc9d132777137fb51R348

If the FieldPath is to a list element (e.g. `spec.containers.image`), and multiple list elements have changed image fields, then there will be multiple key-value pairs

sdowell: This behavior can of course be changed if this does not seem intuitive, I am open to feedback

mengqiy: If it changes 2 image fields in containers, I'd expect something like:
```
spec.containers[0].image: set image from foo:1.2.3 to bar:4.5.6
spec.containers[1].image: set image from foo:latest to bar:4.5.6
```
WDYT?

sdowell: I think this comes down to an implementation detail of kustomize again, since `FieldPath` in this case is tracked as something like `spec.containers.image`. I could see a case for making this change, although I don't believe there is a straightforward way to include the index of a `SequenceNode` in `FieldPath` without changing the behavior of how `FieldPath` is tracked in kustomize.

We could alternatively do something like this given the current API:
```
spec.containers.image: set image from foo:1.2.3 to bar:4.5.6
spec.containers.image: set image from foo:latest to bar:4.5.6
```

WDYT?

mengqiy: This looks better. We can do this for now.

sdowell: Done

functions/go/set-image/image_transformer.go
```diff
@@ -39,6 +42,57 @@ type SetImage struct {
...
+	FieldPath string
+}
+
+// setImageResult maps a previous image value to a new image value where set-image is applied
+// e.g. "nginx:1.20.2" -> "nginx:1.21.6"
+type setImageResult struct {
+	// CurrentValue is the value before applying the set-image mutation
+	CurrentValue string
+	// ProposedValue is the value that will be set after applying set-image
+	ProposedValue string
```
mengqiy, line 66: This name doesn't feel very accurate. Maybe call it `UpdatedValue`.

functions/go/set-image/go.mod
```diff
@@ -5,8 +5,8 @@ go 1.17
 require (
 	github.com/GoogleContainerTools/kpt-functions-catalog/thirdparty/kyaml/fnsdk v0.0.0-20220111011035-c598c94c9a02
 	github.com/stretchr/testify v1.7.0
-	sigs.k8s.io/kustomize/api v0.10.1
-	sigs.k8s.io/kustomize/kyaml v0.13.1
+	sigs.k8s.io/kustomize/api v0.10.2-0.20220127182200-a5b61016bb40
+	sigs.k8s.io/kustomize/kyaml v0.13.2-0.20220127182200-a5b61016bb40
```
mengqiy, line 9: Kustomize has new releases today. We can use the released versions for both api and kyaml.

mengqiy (commented):
A few minor comments.LGTM otherwise

functions/go/set-image/image_transformer.go
sdowell: I based this on the [analogous type from the SDK](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/thirdparty/kyaml/fnsdk/result.go#L111). Do you feel that this has a different context?

functions/go/set-image/go.mod
sdowell: Done

functions/go/set-image/image_transformer.go
sdowell: SGTM, done

mengqiy (approved):

PR #752, opened 2022-01-31
pinning the version of gatekeeper

mikebz:
When looking through examples I noticed that we forgot to pin the gatekeeper content.  Why do I think we should do it:
1) it shows the best practice
2) it introduces stability for our own selves.

Hopefully this is a quick PR review 👀 .

### DIFF
--- a/examples/gatekeeper-disallow-root-user/.expected/results.yaml
+++ b/examples/gatekeeper-disallow-root-user/.expected/results.yaml
@@ -4,7 +4,7 @@ metadata:
   name: fnresults
 exitCode: 1
 items:
-  - image: gcr.io/kpt-fn/gatekeeper:unstable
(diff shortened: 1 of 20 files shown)


sdowell:
My understanding is that it is intentional to use `unstable` as the tag on the master branch. The version is only set to an explicit tag for release branches, e.g.
https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/gatekeeper/v0.2/examples/gatekeeper-imperative

mikebz:
@sdowell you are correct. I saw this python validation function after I submitted the PR.

```
validate_example_kptfile
```

and it does have an expected tag:
```
    if branch == 'master':
        tag = 'unstable'
```

mikebz:
maybe the right action item here is to make that a part of the expectations of function contributions.

PR #756, opened 2022-02-02
Add `set-project-id` function

yuwenma:
### Update
- Extend to support wildcard match and replace a project from a string with given fieldspec. The wildcard should be configured using the `regexPattern` in each fieldspec in  `costs/custom.go` file. The wildcard should use named groups "prefix", "suffix" to distinguish the reserved strings. The new string will be  ${prefix match}${new project ID}${suffix match}  
- thirdparty forks `kustomization/api`, the only change is to add regexPattern to `fieldspec`  

## Description
- Enumerate a builtin spec field list to tell what`project` really means for different KRM resources. So the resources are endowed with practical value on the opposite of merely yaml syntax (templating). 
- Leverage kustomize plugin `annotation` filter to add/update the annotation with key `cnrm.cloud.google.com/project-id`. These annotations include  all KRM resources `metadata.annotations` and  `spec.metadata.annotations` of major k8s Kind resources (e.g. `Deployment`, `Service`, `ReplicaSet`).   
 
TODO:
- Add examples and guides in README
- Update generated/docs.go
- Add tests.
- Extend the specField list to cover more KCC resources.


mengqiy:
Please let me know when this PR is ready for a review.

yuwenma:
> Please let me know when this PR is ready for a review.

Thank you! This is PoC. 
Also, I may move cnrm functions to other dir so they can be well maintained and not accidentally used by irrelevant cases.

efiacor:
No longer a priority

PR #759, opened 2022-02-04
Add `set-default-name` function

yuwenma:
- Update `metaname.name` if the input resources' GVK are white listed. 
- Support name reference (different from kustomize): If a input resource has a fieldspec RNode (has to be ScalaNode) that is  white listed and have a referrer, this fieldspec RNode value will be updated to be the referrer's `metadata.name`. This does not require the referring resource to have the same name as the referrer resource.


efiacor:
GCP specific. Wont do

PR #760, opened 2022-02-04
Add `set-gcp-location` function

yuwenma:


yuwenma:
Related https://github.com/GoogleContainerTools/kpt/issues/2658

PR #762, opened 2022-02-08
Add `set-domain` function

yuwenma:
related https://github.com/GoogleContainerTools/kpt/issues/2677


efiacor:
GCP specific. Wont do

PR #763, opened 2022-02-08
Add `set-org-id` function

yuwenma:
Related https://github.com/GoogleContainerTools/kpt/issues/1400


efiacor:
GCP specific. Wont do

PR #765, opened 2022-02-11
Replace kpt gatekeeper with gator internals

julianKatz:
With the [creation](https://github.com/open-policy-agent/gatekeeper/pull/1786) of the `gator test` CLI in the Gatekeeper project came a new golang library in gatekeeper's `pkg` directory.  This library exposes functionality that replaces code used in the `kpt gatekeeper` function.

Replacing this code with code owned by the Gatekeeper project provides better separation of concerns between the two projects.  For example, when a change in how un-declared fields in JSONSchemas were ingested as part of the introduction of v1 CRDs was fixed in the Constraint Framework underlying Gatekeeper, it took additional code to take advantage of that fix in kpt-gatekeeper (see #649).  By factoring this Gatekeeper-specific logic out of `kpt-functions-catalog`, a future such fix will require only the updating of a dependency and no other code changes.

This PR also includes some changes that came about from following the [contribution guide](https://github.com/julianKatz/kpt-functions-catalog/blob/master/CONTRIBUTING.md).


julianKatz:
@mengqiy PTAL

functions/go/gatekeeper/validate.go
```diff
@@ -15,18 +15,12 @@
...
 	"strconv"
 
-	opaapis "github.com/open-policy-agent/frameworks/constraint/pkg/apis"
-	"github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1"
-	opaclient "github.com/open-policy-agent/frameworks/constraint/pkg/client"
-	"github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local"
-	"github.com/open-policy-agent/frameworks/constraint/pkg/core/templates"
 	opatypes "github.com/open-policy-agent/frameworks/constraint/pkg/types"
-	"github.com/open-policy-agent/gatekeeper/pkg/target"
+	gatortest "github.com/open-policy-agent/gatekeeper/pkg/gator/test"
```
mengqiy, line 23: Is this a testing package?

functions/go/gatekeeper/validate.go
```diff
@@ -35,98 +29,25 @@ import (
...
+		un, ok := o.(*unstructured.Unstructured)
+		if !ok {
+			return nil, fmt.Errorf("cannot cast runtime.Object of kind %q as unstructured", o.GetObjectKind().GroupVersionKind().Kind)
 		}
+
+		unstrucs = append(unstrucs, un)
 	}
 
-	resps, err := client.Audit(ctx)
+	resps, err := gatortest.Test(unstrucs)
```
mengqiy, line 46: Is this doing something like `client.Audit()`? `Test()` is not a great name IMO.

functions/go/gatekeeper/go.mod
```diff
@@ -3,12 +3,12 @@ module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/gateke
 go 1.17
 
 require (
-	github.com/open-policy-agent/frameworks/constraint v0.0.0-20211025234246-f478d8a555b5
-	github.com/open-policy-agent/gatekeeper v0.0.0-20211115204247-3ba8e93049b5 // This is v3.7.0. It has a semver major version of 2 or higher and is not a Go module yet.
-	github.com/spf13/cobra v1.1.3
-	k8s.io/apimachinery v0.21.2
+	github.com/open-policy-agent/frameworks/constraint v0.0.0-20220121182312-5d06dedcafb4
+	github.com/open-policy-agent/gatekeeper v0.0.0-20220208150435-b36e85531dbe // This is v3.7.0. It has a semver major version of 2 or higher and is not a Go module yet.
```
mengqiy, line 7: I guess the version is no longer v3.7.0. Please update the comment.

mengqiy:
Go unit test is failing https://github.com/GoogleContainerTools/kpt-functions-catalog/runs/5161761800?check_suite_focus=true#step:4:267

functions/go/gatekeeper/validate.go
julianKatz: no, it's the libraries for the `gator test` command: https://open-policy-agent.github.io/gatekeeper/website/docs/next/gator#the-gator-test-subcommand

functions/go/gatekeeper/validate.go
julianKatz: These names were chosen in the Gatekeeper project in an effort to align with other CLI tools that are part of the `open-policy-agent` github org.  i.e. I can't change it without a large political effort there.

functions/go/gatekeeper/go.mod
julianKatz: This is the current `HEAD` of gatekeeper, and doesn't correspond to a SEMVER tag.  What would you like in the comment?  Could we just remove it?

I think this is implied by the `go.mod` format of including the date and commit hash, rather than a SEMVER tag.

functions/go/gatekeeper/validate.go
mengqiy: Thanks for clarifying! Can you please add a comment that provide the context and call out it's not a testing package

functions/go/gatekeeper/go.mod
mengqiy: Let's remove the comment here

functions/go/gatekeeper/validate.go
```diff
@@ -35,98 +29,25 @@ import (
...
-	}
+	unstrucs := []*unstructured.Unstructured{}
 
-	for _, obj := range objects {
-		if _, err = client.AddData(ctx, obj); err != nil {
-			return nil, err
+	for _, o := range objects {
+		un, ok := o.(*unstructured.Unstructured)
+		if !ok {
+			return nil, fmt.Errorf("cannot cast runtime.Object of kind %q as unstructured", o.GetObjectKind().GroupVersionKind().Kind)
```
mengqiy, line 40: `gatortest.Test()` expect `[]*unstructured.Unstructured`. But our code [here](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/765/files#diff-9a709bd97c536723b79e86281fcfadba7f81cf32e59254d0a19131cdd6059379R72-R83) convert some of them to typed objects (e.g. appsv1.Deployment). And IIUC it will cause error here.
If that's the case, we should remove the scheme the related code. We should always use `*unstructured.Unstructured` to deserialize.

mengqiy: I update the line in the comment above. GH doesn't bring us to the right place when using the old URL. Please check the updated link.

mengqiy: Our e2e tests also catch this: https://github.com/GoogleContainerTools/kpt-functions-catalog/runs/5162431817?check_suite_focus=true#step:7:507

julianKatz: My [newest commit](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/765/commits/b980789cc8e0082c324364e04913b370b8932aa0) should make the changes you've suggested.  PTAL and LMK if you'd suggest other changes.

Please run CI as well, as that will reveal if the test is now passing.

mengqiy (approved):
LGTM

mengqiy:
I will merge it after CI passes.

PR #766, opened 2022-02-14
Add `source-gcloud-config` function

yuwenma:
## Description
This function shall run as executable. It calls `gcloud` to get the local default gcloud context and stores the gcloud config in a gcloud-config.yaml file using ConfigMap kind, whose meta name is fixed to `gcloud-config.kpt.dev`. For nested packages, if the top package does not have gcloud-config.yaml file, it creates one. If the file already exists on the top package, it updates the latest gcloud context data. This function will also update sub-package gcloud-config.yaml files only if they already exist. 

Related to 
- https://github.com/GoogleContainerTools/kpt/issues/2795
- https://github.com/GoogleContainerTools/kpt/issues/2590

This function relies on executing `gcloud` on the client side. Thus, we don't add the CI test and the document only covers the imperative function call with `--exec` flag.


functions/go/source-gcloud-config/exec/gcloud.go
```diff
@@ -0,0 +1,107 @@
...
+		"region":    region,
+		"domain":    domain,
+		"orgID":     orgID,
+	}, nil
+
+}
+
+func getGcloudConfig(property string) string {
+	var cmdOut, cmdErr bytes.Buffer
+	cmd := exec.Command("gcloud", "config", "get-value", property)
```
martinmaly, line 73: There may be more performant way to extract the config than via execing gcloud binary, or execing it multiple times.

I suspect the gcloud config file structure is documented and can be directly parsed?

functions/go/source-gcloud-config/kptfile/v1/types.go
```diff
@@ -0,0 +1,328 @@
+// Copyright 2021 Google LLC
```
martinmaly, line 1: can this be obtained by import from kpt repository rather than copying the file?

functions/go/source-gcloud-config/exec/gcloud.go
yuwenma: > There may be more performant way to extract the config than via execing gcloud binary, or execing it multiple times.
> 
> I suspect the gcloud config file structure is documented and can be directly parsed?

I was exactly thinking the same 😄   But really cannot find where those gcloud configs are stored (searched internal docs and google3). Another option is to use `gcloud config list` . I can go that way if it's better.

functions/go/source-gcloud-config/kptfile/v1/types.go
yuwenma: I swear I asked the similar questions before. My understanding is that kpt does not want function developers to vendor kpt repo. FYI, the kptfile schema is in [kpt-function-sdk  repo](https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/pkg/api/kptfile/v1/types.go) and the kpt repo keeps a copy of it.

martinmaly: Thank you for the added context! vendoring from kpt-function-sdk seems ideal.

@loudej fyi re: authoritative copy of types.go relevant to OCI work

functions/go/source-gcloud-config/exec/gcloud.go
martinmaly: ${HOME}/.config/gcloud/... seems to be where they are. There seems to be some indirection happening but with some luck we can figure it out :)

yuwenma: oh, how come I forgot HOME/.config

functions/go/source-gcloud-config/exec/gcloud.go
```diff
@@ -0,0 +1,110 @@
...
+	"strings"
+)
+
+var domainRegex = regexp.MustCompile(`\S+@(\S+)`)
+var PullGcloudConfig = PullLocalGcloudConfig
+
+// PullLocalGcloudConfig executes `gcloud` commands to get default gcloud config value.
+// Potential caveat: The gcloud stdout format is based on Google Cloud SDK 351.0.0. If gcloud SDK changes the output
+// format, it might break this generator function on the client-side. Once that happened, this function shall be modified
+// to be able to parse different stdout format according to the installed SDK version.
```
yuwenma, line 30: @martinmaly Yeah, parsing the config from gcloud file gives a better handling for this caveat. Thank you!

functions/go/source-gcloud-config/kptfile/v1/types.go
yuwenma: done. Actually I changed the config so the Kptfile is not needed.

functions/go/source-gcloud-config/exec/gcloud.go
bharathkkb: `gcloud config list --format json` could also give parsable JSON output

functions/go/source-gcloud-config/exec/gcloud.go
```diff
@@ -0,0 +1,121 @@
...
+		projectID, "--format=get(id)")
+	cmdListAncestors.Stdout = &buf
+	cmdListAncestors.Stderr = &err
+	if e := cmdListAncestors.Run(); e != nil {
+		return "", e
+	}
+	if err.Len() > 0 {
+		return "", fmt.Errorf(err.String())
+	}
+	cmdOrgID := exec.Command("tail", "-1")
```
bharathkkb, line 109: `gcloud projects get-ancestors PROJECT_ID --format=json` should let you filter to match by `type == organization` which maybe better than execing  tail.

bharathkkb (commented):
drive by comment since I thought this would make the logic simpler

functions/go/source-gcloud-config/exec/gcloud.go
yuwenma: Thanks.

functions/go/source-gcloud-config/main.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/source-gcloud-generator/generated"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+)
+
+type Processor struct{}
+
+func (p *Processor) Process(resourceList *framework.ResourceList) error {
+	err := func() error {
+		gen := &GcloudConfigGenerator{}
```
droot, line 29: We want to vendor this function in `kpt` repo as builtin. To be able to invoke it as library, we will have to refactor it so that the core logic can be executed by CLI (like here) and also from the kpt code where it accepts serialized input resource-list and output serialized resource-list.

@martinmaly had an example for that in apply-setters, not sure if that got merged.

functions/go/source-gcloud-config/exec/gcloud.go
yuwenma: @bharathkkb prefer not using gcloud config list here because the output is in ini file format. It's different than the org-id one.

bharathkkb: Generally it is [not recommended](https://cloud.google.com/sdk/docs/scripting-gcloud#:~:text=Don%27t%20depend%20on%20the%20raw%20output%20of%20messages%20printed%20to%20standard%20output) to rely on gcloud stdout. I haven't looked into the ini format so it maybe as stable as json op 😅

functions/go/source-gcloud-config/exec/gcloud.go
yuwenma: Updated.

examples/source-gcloud-config-simple/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+This package contains the `source-gcloud-config` executable.
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/source-gcloud-config-simple
+```
+
+### Run the function
+
+```shell
+$ kpt fn eval --exec ./source-gcloud-config .
```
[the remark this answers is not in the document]
yuwenma: This runs as an exec and works. I don't think we want to change this to an image.
I refactored the code so it can be imported by kpt as a lib as well :)

examples/source-gcloud-config-simple/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+### Expected result
+
+A new file `configmap_gcloud-config.kpt.dev.yaml` should be added which contains a ConfigMap resource
```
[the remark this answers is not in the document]
yuwenma: Kpt (or the kio it relies on) decides the name when writing it to the disk.

functions/go/Makefile
```diff
@@ -37,6 +37,7 @@ FUNCTIONS := \
 	set-labels \
 	set-namespace \
 	set-project-id \
+	source-gcloud-config \
```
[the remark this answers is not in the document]
yuwenma: ah, this is just the directory name. it can be called anything in kpt.

functions/go/source-gcloud-config/main.go
yuwenma: Done.  the lib would be like 
```go
import "github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/source-gcloud-generator/generator"

func () {
       gen  := generator.NewGcloudConfigGenerator()
       newItems, err := gen.Generate(resourceList.Items)
       ...  
}
```

functions/go/Makefile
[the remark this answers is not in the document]
yuwenma: I saw both `generate-gcloud-config` and `source-gcloud-config` are used in your doc. I picked source because I don't want this to be surfaced as a normal generator. Unlike other curated KRM functions, it 1.)does not manipulate the  input resources and 2.) it calls executable `gcloud` which is abnormal. It serves more as a pure (functionConfig) source provider.

examples/source-gcloud-config-simple/README.md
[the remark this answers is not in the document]
yuwenma: Done.

functions/go/source-gcloud-config/gcloudconfig/generator_test.go
```diff
@@ -0,0 +1,54 @@
...
+  namespace: test-namespace
+  orgID: test-orgID
+  projectID: test-projectID
+  region: test-region
+  zone: test-zone
+kind: ConfigMap
+metadata:
+  annotations:
+    config.kubernetes.io/local-config: "true"
+  name: gcloud-config.kpt.dev
```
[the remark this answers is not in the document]
yuwenma: oh, this follows the naming convention here. https://github.com/GoogleContainerTools/kpt/issues/2184#issuecomment-949863947

Basically the variant constructor resource name will be in the form of `<resourceType>.kpt.dev`

functions/go/source-gcloud-config/gcloudconfig/generator.go
```diff
@@ -0,0 +1,73 @@
...
+	var newNodes []*yaml.RNode
+	for _, curNode := range nodes {
+		gcloudGvk := resid.GvkFromNode(gcloudNode)
+		curNodeGvk := resid.GvkFromNode(curNode)
+		if curNode.GetName() == gcloudNode.GetName() && curNodeGvk.Equals(gcloudGvk) {
+			continue
+		}
+		newNodes = append(newNodes, curNode)
+	}
+	newNodes = append(newNodes, gcloudNode)
```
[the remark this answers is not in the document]
yuwenma: Since this is pure gcloud context and does not contain Kptfile or kpt package information, the `gcloud-config` resources will always be the same, and the the gcloudNode is stored on the top-level package. 

Do you mind explaining more about "they will simply be dropped and only one will be added at the top". I was assuming "drop" means the sub-packge gcloud-config.yaml files are removed from disk. But my manual testing shows sub-package gcloud-config files are not passed into the hydration pipeline. I doubt if that's because all gcloud-config shares the same GVKNN so duplicate nodes are filtered out. I haven't dug into the code to prove that. But by running `kpt fn source` only, it seems only the top-packge gcloud-config is listed.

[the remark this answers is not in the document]
droot: >  I was assuming "drop" means the sub-packge gcloud-config.yaml files are removed from disk. But my manual testing shows sub-package gcloud-config files are not passed into the hydration pipeline.

It is because `kpt fn render` doesn't provide `meta resources` as input to functions today. This PR https://github.com/GoogleContainerTools/kpt/pull/2781 is going to change that.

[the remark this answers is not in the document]
yuwenma: This gcloud-config is `ConfigMap` kind, when not used as functionConfig, running `kpt fn source` only gives me the top level gcloud-config and ignore the gcloud-config from sub-package.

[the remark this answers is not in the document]
yuwenma: yeah, the "weird" behavior is because the gcloud is passed in as meta resource

Since all gcloud-config.yaml have the same gcloud context, I assume running this gcloud config function shall update both top package (create if not exists) and  sub pacakges gcloud-config.yaml file. I updated the logic to reflect that. Please see the last commit

droot (approved):

PR #768, opened 2022-03-08
new function: apply-replacements

natasha41575:
/cc @mengqiy 
/cc @yuwenma 
/cc @droot


functions/go/Makefile
```diff
@@ -19,6 +19,7 @@ GOBIN := $(shell go env GOPATH)/bin
 
 # Edit this list to contain all go functions
 FUNCTIONS := \
+    apply-replacements \
```
yuwenma, line 22: super nit: fix indentation

functions/go/apply-replacements/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN go mod download
+
+COPY . .
+RUN go build -o /usr/local/bin/function ./
+
+#############################################
+
+FROM alpine:3.13
+COPY --from=0 /usr/local/bin/function /usr/local/bin/function
+ENTRYPOINT ["function"]
```
yuwenma, line 15: This Dockerfile is not needed because functions under `./functions/go/` should use the same Dockerfile defined in https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/build/docker/go/Dockerfile

functions/go/apply-replacements/README.md
```diff
@@ -0,0 +1,107 @@
+# apply-replacements
+
+### Overview
+
+<!--mdtogo:Short-->
+
+Invoke the [kustomize replacements] feature as a KRM function.
```
yuwenma, line 7: suggest adding a link to the kustomize replacements doc.

functions/go/apply-replacements/README.md
```diff
@@ -0,0 +1,107 @@
...
+```yaml
+apiVersion: kpt.dev/v1alpha1
+kind: Replacements
+metadata:
+  name: replacements-fn-config
+replacements:
+  # your replacements here
+```
+
+The syntax for `replacements` is described in the [kustomize replacements] docs.
```
yuwenma, line 26: suggest giving a simple example here and add the link to kustomize replacement.

functions/go/apply-replacements/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+### FunctionConfig
+
+<!--mdtogo:Long-->
+
+We use a `Replacements` object to configure the `apply-replacements` function. 
+
+```yaml
+apiVersion: kpt.dev/v1alpha1
+kind: Replacements
```
yuwenma, line 19: nit: how about `SetReplacements` or `ApplyReplacements`. curated functionConfig type has the naming convention like `SetNamespace` `SetAnnotations` `SetImage`

functions/go/apply-replacements/main.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"os"
+
+	sdk "github.com/GoogleContainerTools/kpt-functions-catalog/thirdparty/kyaml/fnsdk"
```
yuwenma, line 6: this fnsdk is experimental library, we eventually should use the one from kpt-functions-sdk. Can you add a TODO here and to the  kpt-functions-catalog/thirdparty/kyaml/fnsdk as well so we know this function need to be updated when deprecating kpt-functions-catalog/thirdparty/kyaml/fnsdk?

functions/go/apply-replacements/main.go
```diff
@@ -0,0 +1,22 @@
...
+
+func main() {
+	if err := sdk.AsMain(sdk.ResourceListProcessorFunc(applyReplacements)); err != nil {
+		os.Exit(1)
+	}
+}
+
+func applyReplacements(rl *sdk.ResourceList) error {
+	r := Replacements{}
+	return r.Process(&framework.ResourceList{
```
yuwenma, line 18: Since this KRM function is using the fnsdk library, can you use the [fnsdk.ResourceList](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/thirdparty/kyaml/fnsdk/resourcelist.go#L23)?

functions/go/apply-replacements/replacements.go
```diff
@@ -0,0 +1,56 @@
...
+func (r *Replacements) Process(rl *framework.ResourceList) error {
+	if err := r.Config(sdk.NewFromRNode(rl.FunctionConfig)); err != nil {
+		return err
+	}
+	transformedItems, err := replacement.Filter{Replacements: r.Replacements}.Filter(rl.Items)
+	if err != nil {
+		return err
+	}
+	rl.Items = transformedItems
+	return nil
```
yuwenma, line 55: Suggest recording the errors in rl.Results field. Here's some examples https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/set-image/main.go#L25

yuwenma (changes requested):
Add some comments.

functions/go/apply-replacements/README.md
mengqiy, line 7: `Invoke` implies it invokes kustomize's binary directly. And it's not the case here.

functions/go/apply-replacements/README.md
```diff
@@ -0,0 +1,107 @@
...
+<!--mdtogo-->
+
+### FunctionConfig
+
+<!--mdtogo:Long-->
+
+We use a `Replacements` object to configure the `apply-replacements` function. 
+
+```yaml
+apiVersion: kpt.dev/v1alpha1
```
mengqiy, line 18: Normally we use `fn.kpt.dev/v1alpha1`.

functions/go/apply-replacements/README.md
mengqiy: +1
Maybe `ApplyReplacements`.

functions/go/apply-replacements/Dockerfile
mengqiy: +1

functions/go/apply-replacements/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+The syntax for `replacements` is described in the [kustomize replacements] docs. 
+
+<!--mdtogo-->
+
+### Examples
+
+<!--mdtogo:Examples-->
+
+Suppose you have the following Job:
```
mengqiy, line 34: It seems this example is identical to the one in examples dir.
If that's the case, we can emit the examples section here.

functions/go/apply-replacements/main.go
mengqiy: +1

functions/go/apply-replacements/replacements.go
```diff
@@ -0,0 +1,56 @@
+package main
+
+import (
+	"fmt"
+	"io"
+
+	sdk "github.com/GoogleContainerTools/kpt-functions-catalog/thirdparty/kyaml/fnsdk"
+	"sigs.k8s.io/kustomize/api/filters/replacement"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
```
mengqiy, line 10: When users use `fnsdk` package, they don't need to use `kyaml/fn/framework` anymore.

mengqiy:
I'd suggest making its logic improtable by other go program.
With it, we can compile it as part of porch if we want to do.

functions/go/Makefile
natasha41575: done

functions/go/apply-replacements/Dockerfile
natasha41575: done

functions/go/apply-replacements/README.md
natasha41575: done

functions/go/apply-replacements/main.go
natasha41575: Now using kpt-functions-sdk!

functions/go/apply-replacements/main.go
natasha41575: done!

functions/go/apply-replacements/README.md
natasha41575: fixed!

functions/go/apply-replacements/README.md
natasha41575: done

functions/go/apply-replacements/replacements.go
natasha41575: removed!

functions/go/apply-replacements/README.md
natasha41575: done!

functions/go/apply-replacements/replacements.go
```diff
@@ -0,0 +1,56 @@
...
+	return nil
+}
+
+// Run accepts a reader and writer, and calls framework.Execute to run the function. This exists
+// to enable the function to be invoked as a library.
+func (r *Replacements) Run(reader io.Reader, writer io.Writer) error {
+	rw := &kio.ByteReadWriter{
+		Reader:                reader,
+		Writer:                writer,
+		KeepReaderAnnotations: true,
```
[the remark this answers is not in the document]
natasha41575: Got it. I added this here because I thought (maybe incorrectly) that in standup we said that we wanted to have a `Run` method similar to the [pkg_context builtin function](https://github.com/GoogleContainerTools/kpt/blob/43c9d295300018748339802542ad6704d10832db/internal/builtins/pkg_context.go#L51), to make the logic importable. 

If we want to have the `Run` method that takes in a reader and a writer, but we don't want to use `framework.Execute`, maybe we can add a `framework.Execute` to the fn sdk? However, `Run` is not strictly necessary, so I have removed it, and I made `ApplyReplacements(*fn.ResourceList)` in main.go public. Kpt/porch will be able to call this function if they would like to import the replacements logic. 

@yuwenma @droot FYI.

natasha41575:
> I'd suggest making its logic improtable by other go program.

@mengqiy I made `ApplyReplacements(*fn.ResourceList)` in main.go public. LMK if this fulfills this requirement.

functions/go/apply-replacements/go.mod
```diff
@@ -0,0 +1,34 @@
+module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-image
```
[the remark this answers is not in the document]
natasha41575: done, thank you

natasha41575:
Question for @mengqiy and @yuwenma, right now we have to convert to RNode because the replacements filter (like all kustomize filters) accepts and returns a list of RNodes. Do we want to add some sort of wrapper to the kpt-functions-sdk to the kustomize filters so that users do not need to do this conversion?

functions/go/apply-replacements/replacements.go
natasha41575: These errors already get recorded in the Results field when run by kpt. Kpt has some logic to look at the returned error and store it in Results.

For example, currently running `kpt fn render` on an error package gives:

```
Package "apply-replacements-error": 
[RUNNING] "gcr.io/kpt-fn/apply-replacements:unstable"
[PASS] "gcr.io/kpt-fn/apply-replacements:unstable" in 1.1s
  Results:
    [error]: nothing selected by Pod.[noVer].[noGrp]/my-po.[noNs]:spec
```

If we record the error in rl.Results in the function logic itself, we will duplicate the error:

```
Package "apply-replacements-error": 
[RUNNING] "gcr.io/kpt-fn/apply-replacements:unstable"
[PASS] "gcr.io/kpt-fn/apply-replacements:unstable" in 1.1s
  Results:
    [error]: nothing selected by Pod.[noVer].[noGrp]/my-po.[noNs]:spec
    [error]: nothing selected by Pod.[noVer].[noGrp]/my-po.[noNs]:spec
```

I suggest we just return the error when we get one and let the orchestrator (kpt) handle it, rather than recording it again.

functions/go/apply-replacements/README.md
natasha41575: changed `invoke` to `use`, thank you for the suggestion!

functions/go/apply-replacements/main.go
```diff
@@ -0,0 +1,76 @@
...
+		return err
+	}
+	rl.Items = transformedItems
+	return nil
+}
+
+func (r *Replacements) Transform(items []*fn.KubeObject) ([]*fn.KubeObject, error) {
+	var transformedItems []*fn.KubeObject
+	var nodes []*yaml.RNode
+	for _, obj := range items {
```
natasha41575, line 62: When https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/526/files#r826210047 is done, we can use that instead of doing this conversion

yuwenma: done.

examples/apply-replacements-simple/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-catalog.git/examples/apply-replacements-simple
+```
+
+We use a `ApplyReplacements` object to configure the `apply-replacements` function. 
+
+```yaml
+apiVersion: fn.kpt.dev/v1alpha1
+kind: ApplyReplacements
```
droot, line 20: Should this be singular `ApplyReplacement` ?  It's a config kind that can contain replacements.

functions/go/Makefile
```diff
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#      http://www.apache.org/licenses/LICENSE-2.0
+#\t\t\thttp://www.apache.org/licenses/LICENSE-2.0
```
droot, line 7: some extra tabs (probably from your IDE)

functions/go/apply-replacements/main.go
```diff
@@ -0,0 +1,76 @@
...
+	r.Replacements = []types.Replacement{}
+	if err := functionConfig.As(r); err != nil {
+		return fmt.Errorf("unable to convert functionConfig to %s:\n%w",
+			"replacements", err)
+	}
+	return nil
+}
+
+// Process implements framework.ResourceListProcessor interface.
+func (r *Replacements) Process(rl *fn.ResourceList) error {
```
droot, line 47: @natasha41575 @mengqiy @yuwenma 

Regarding using this function as a library inside porch or kpt:

This is the interface https://github.com/GoogleContainerTools/kpt/blob/main/pkg/fn/eval.go#L27 porch or kpt CLI uses for injecting builtin function. It is basically at the wire-format level so that function-executor can be replaced by the builtin/remote/container-based executors.

So I see two levels:

1. Wireformat level (that takes in reader and writer streams) 
2. ResourceList data structure Level (the one defined by `Process` method here)

If we finalize the 2) as this `Processor` interface, then we can have an adaptor function in `kpt` or `porch` that implements 1) for each function. Here is an example (it uses framework....but we can migrate to use function SDK) --> https://github.com/GoogleContainerTools/kpt/blob/main/internal/builtins/pkg_context.go#L51

functions/go/apply-replacements/replacements.go
yuwenma: Hmmm, I don't think kpt appends error (sdk does though, but only when the Process function returns err), and the apply-replacements error should results in `[FAIL]` not `[PASS]`. Mind sharing the code about the duplicate errors case?

mengqiy (commented):
> Reduce the exposed surface: moving variants to internal.

The `variants` were private so they are not exposed surface. From a user's perspective, there are no surface changes related to `variants`.

functions/go/apply-replacements/main.go
```diff
@@ -0,0 +1,76 @@
...
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+func main() {
+	if err := fn.AsMain(fn.ResourceListProcessorFunc(ApplyReplacements)); err != nil {
+		os.Exit(1)
+	}
+}
+
+func ApplyReplacements(rl *fn.ResourceList) error {
```
mengqiy, line 19: Things under the `main` package can not be exported by other packages. We should move the business logic to a separate package.

`main` package can be as simple as
```
func main() {
	if err := fn.AsMain(fn.ResourceListProcessorFunc(anotherpkg.ApplyReplacements)); err != nil {
		os.Exit(1)
	}
}
```

functions/go/apply-replacements/replacements.go
natasha41575: > Hmmm, I don't think kpt appends error (sdk does though, but only when the Process function returns err)

I see. Yeah, since Process function is returning an error, the error is already getting appended to the Results. 

> Mind sharing the code about the duplicate errors case?

```
// Process implements framework.ResourceListProcessor interface.
func (r *Replacements) Process(rl *fn.ResourceList) error {
	if err := r.Config(rl.FunctionConfig); err != nil {
		return err
	}
	transformedItems, err := r.Transform(rl.Items)
	if err != nil {
		rl.Results = append(rl.Results, &fn.Result{
			Message: err.Error(),
			Severity: fn.Error,
		})
		return err
	}
	rl.Items = transformedItems
	return nil
}
```

Changing `Process` to the above (appending the error to Results) results in the error being duplicated in the Results field because the returned error is appended elsewhere after Process returns the error. 

> the apply-replacements error should results in [FAIL] not [PASS]

I am not sure why this is happening, but it is definitely strange. I will investigate.

functions/go/apply-replacements/main.go
natasha41575: I see. So to confirm, you are suggesting these functions implement `Process`, and kpt/porch can do the translation from Wireformat to ResourceList?

examples/apply-replacements-simple/README.md
natasha41575: Hm, we can. I don't have a strong opinion, but I am slightly in favor of plural `ApplyReplacements` because (1) the function name is apply-replacements which is plural and (2) because this one function can apply multiple replacements. `ApplyReplacement` might imply that the function can only apply one replacement per invocation.  

However I am ok to change it if you feel strongly.

functions/go/apply-replacements/replacements.go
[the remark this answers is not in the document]
natasha41575: Further discussion is in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/768#discussion_r826278649

functions/go/apply-replacements/main.go
natasha41575: > Things under the main package can not be exported by other packages. We should move the business logic to a separate package.

TIL. Thank you, done! Moved business logic to `replacements` package.

natasha41575: Fixed, thanks for the catch!

functions/go/Makefile
natasha41575: Fixed, thanks for the catch!

functions/go/apply-replacements/main.go
droot: Yes. Though, it will be good to have the helpers in fn-sdk.

droot (commented):
Looks good to me. Will leave @mengqiy and @yuwenma to approve it.

mengqiy (approved):
LGTM

mengqiy:
We may want to wait for @yuwenma to take another look

functions/go/apply-replacements/main.go
natasha41575: I can add some helpers to the sdk to do this. FYI @yuwenma

functions/go/apply-replacements/replacements/replacements.go
```diff
@@ -0,0 +1,71 @@
...
+func (r *Replacements) Transform(items []*fn.KubeObject) ([]*fn.KubeObject, error) {
+	var transformedItems []*fn.KubeObject
+	var nodes []*yaml.RNode
+	for _, obj := range items {
+		objRN := obj.ToRNode()
+		nodes = append(nodes, objRN)
+	}
+	transformedNodes, err := replacement.Filter{
+		Replacements:  r.Replacements,
+	}.Filter(nodes)
```
yuwenma, line 63: In lieu of supporting the conversion between RNode to KubeObjects, we shall have replacement.Filter (maybe forked to sdk) to accept KubeObjects intead. WDYT?

functions/go/apply-replacements/main.go
yuwenma: +1 to add the helper in fn-sdk

functions/go/apply-replacements/replacements/replacements.go
mengqiy: I have a similar problem with starlark fn: https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/starlark.go#L51
The `Filter` takes a list of RNode as input. I will need to convert KubeObjects to RNodes.

yuwenma: Good to now, I'll try prioritizing this fork work.

natasha41575: @yuwenma did you have some concerns with @droot's suggestion in standup? If the sdk can convert a yaml string to KubeObject, than we can do RNode->yaml->KubeObject easily. Do you prefer to fork instead?

yuwenma: "convert a yaml string to KubeObject" --> yes, we have `fn.ParseKubeObject()` that converts a byte slice to KubeObject. From the long term perspective considering the code quality and maintenance cost, forking is more clean and reliable than RNode->yaml-> KubeObject.  I would say `fn.ParseKubeObject()` is a temporary workaround for replacement transformer.

natasha41575: Ah, makes sense. Wasn't aware of fn.ParseKubeObject(). I will use that for now and will use the fork when it is ready. Thank you!

functions/go/apply-replacements/replacements.go
natasha41575: Oh I see, I didn't realize we didn't want to return the error. Made the change, thanks!

functions/go/apply-replacements/main.go
natasha41575: Filed https://github.com/GoogleContainerTools/kpt/issues/2897. Will start work on this shortly.

yuwenma (approved):
Sweet! Thanks for the change.

PR #769, opened 2022-03-08
feat: set-namespace without kyaml/rnode dependencies

yuwenma:
This function re-implements the set-namespace function but removes all kyaml dependencies. 
It's not ready to be merged until its dependent functions are merged in  kpt-functions-sdk


functions/go/set-namespace-v2/namespace_transformer.go
```diff
@@ -0,0 +1,151 @@
...
+type NamespaceTransformer struct {
+	Namespace string
+	DependsOnMap map[string]bool
+	Errors []string
+}
+
+func (p *NamespaceTransformer) Config(o *krmfn.KubeObject) error {
+	switch {
+	case o.APIVersion() == "v1" && o.Kind() == "ConfigMap" && o.Name() == "kptfile.kpt.dev":
+		p.Namespace = o.GetStringOrDie("data", "name")
```
mengqiy, line 62: `GetString` returns a potential error while `GetStringOrDie` exits on error. `***OrDie` pattern comes from kustomize, there are many usages in its code base.
If we want to be more conservative, we should use `GetString`. If we want to to more concise (fewer LoC), we can use `GetStringOrDie`.
We need to decide which one should we use here.

functions/go/set-namespace-v2/namespace_transformer.go
```diff
@@ -0,0 +1,151 @@
...
+
+type NamespaceTransformer struct {
+	Namespace string
+	DependsOnMap map[string]bool
+	Errors []string
+}
+
+func (p *NamespaceTransformer) Config(o *krmfn.KubeObject) error {
+	switch {
+	case o.APIVersion() == "v1" && o.Kind() == "ConfigMap" && o.Name() == "kptfile.kpt.dev":
```
bgrant0607, line 61: If the user specifies a function config, it seems like that should take precedence.

functions/go/set-namespace-v2/namespace_transformer.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	return nil
+}
+
+
+func (p *NamespaceTransformer) Transform(objects []*krmfn.KubeObject) {
+	p.SetDependsOnMap(objects)
+	for _, o := range objects {
+		switch {
+		case o.APIVersion() == "v1" && o.Kind() == "Namespace":
```
bgrant0607, line 90: A GVK matching function would look a little cleaner.

case isGVK("v1", "Namespace"):

functions/go/set-namespace-v2/namespace_transformer.go
```diff
@@ -0,0 +1,151 @@
...
+		p.Namespace = o.GetStringOrDie("data", "name")
+		if p.Namespace == "" {
+			return fmt.Errorf("`data.name` should not be empty")
+		}
+	case o.APIVersion() == "v1" && o.Kind() == "ConfigMap":
+		p.Namespace = o.GetStringOrDie("data", "namespace")
+		if p.Namespace == "" {
+			return fmt.Errorf("`data.namespace` should not be empty")
+		}
+	case o.APIVersion() == fnConfigAPIVersion && o.Kind() == legacyFnConfigKind:
```
mengqiy, line 71: We can remove things related to legacy kind

functions/go/set-namespace-v2/namespace_transformer.go
yuwenma: moved this convention to later case.

functions/go/set-namespace-v2/namespace_transformer.go
yuwenma: Sure. Added `IsGVK` to our sdk and updated the caller function

functions/go/set-namespace-v2/namespace_transformer.go
yuwenma: done! Thanks!

functions/go/set-namespace-v2/namespace_transformer.go
yuwenma: Personally prefer not to handle the errors on the caller side (more clean code and let SDK decides how to reveal the message and give suggestions). I added the panic recovery for `***OrDie` pattern to give users some guidance on which resource's which field path causes the problem. https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/531  
For example, I changed my `ClusterRoleBinding` resource `subjects` to be a list of string (it should be a list of new MapVariant), the output is
```bash
[RUNNING] "./set-ns"
[FAIL] "./set-ns" in 700ms
  Stderr:
    "2022/03/15 22:34:07 Resource(apiVersion=v1, kind=ClusterRoleBinding, Name=yuyu2) has unmatched field type: `subjects`"
    ""
  Exit code: 1
```

functions/go/set-namespace-v2/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-namespace-v2
+
+go 1.16
+
+require github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220315215931-2ef5e56080fb // indirect
```
yuwenma, line 5: needs to be updated once https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/531 is merged.

functions/go/Makefile
```diff
@@ -35,7 +35,7 @@ FUNCTIONS := \
 	set-enforcement-action \
 	set-image \
 	set-labels \
-	set-namespace \
+	set-namespace-v2 \
```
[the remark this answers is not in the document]
mengqiy: The old function will be living in the release branch. If any maintenance release is needed, we can still do it.

functions/go/set-namespace-v2/main.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+func main() {
+	if err := fn.AsMain(fn.ResourceListProcessorFunc(run)); err != nil {
+		os.Exit(1)
+	}
+}
+
+func run(rl *fn.ResourceList) error {
```
mengqiy, line 29: Same comment as in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/768#discussion_r826346186
`main` module is not importable.

yuwenma: oh, right! Good catch, thank you!

examples/set-namespace-simple/README.md
```diff
@@ -1,10 +1,9 @@
-# set-namespace: Simple Example
+# set-namespace: Simple example
```
mengqiy, line 1: nit: why change this? It will inconsistent with other examples.

examples/set-namespace-simple/README.md
```diff
@@ -37,11 +36,9 @@ key-value pair:
 Invoke the function by running the following commands:
 
 ```shell
-$ kpt fn render set-namespace-simple
+$ kpt fn render set-namespace
```
mengqiy, line 39: `set-namespace-simple` is the package name you fetched from earlier step.
If you change this, the steps here won't work.

examples/set-namespace-simple/README.md
```diff
@@ -37,11 +36,9 @@ key-value pair:
...
 ```
 
 ### Expected result
 
-Check all resources have `metadata.namespace` set to `example-ns`:
-
-[`set-namespace`]: https://catalog.kpt.dev/set-namespace/v0.1/
```
mengqiy, line 47: If we remove this, the link in line 5 will be broken.
We can use `https://catalog.kpt.dev/set-namespace/v0.3/`, if we plan to release it very soon.

functions/go/set-namespace/README.md
```diff
@@ -4,76 +4,47 @@
...
 
-The `set-namespace` function update or add namespace to all namespaced
-resources. Kubernetes supports multiple virtual clusters backed by the same
-physical cluster through namespaces.
-
-Namespaces are often used in the following scenarios:
-
-- Separate resources between environments (prod, staging and test).
-- Separate resources between different team or users to divide resource quota.
+The `set-namespace` function replace the `namespace` specific resource type in a variety
```
mengqiy, line 7: nit: replaces

functions/go/set-namespace/README.md
```diff
@@ -4,76 +4,47 @@
 
 <!--mdtogo:Short-->
 
-The `set-namespace` function update or add namespace to all namespaced
-resources. Kubernetes supports multiple virtual clusters backed by the same
-physical cluster through namespaces.
-
-Namespaces are often used in the following scenarios:
```
mengqiy, line 11: We should follow the template: https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/_template/README.md#overview

Some of these lines can be kept. If you don't think they should be part of the short description (i.e. between `<!--mdtogo:Short-->` and `<!--mdtogo-->`), we can keep them between `<!--mdtogo-->` and `<!--mdtogo:Long-->`

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -0,0 +1,167 @@
...
+		case o.IsGVK("apiregistration.k8s.io/v1", "APIService"):
+			err := o.Set(p.Namespace, "spec", "service", "namespace")
+			if err != nil {
+				p.Errors = append(p.Errors, err.Error())
+			}
+		case o.GetKind() == "ClusterRoleBinding" || o.GetKind() == "RoleBinding":
+			var subjects []*Subject
+			o.GetOrDie(&subjects, "subjects")
+			for _, s := range subjects {
+				if s.Name == "default" {
```
mengqiy, line 121: As discussed in [this doc](https://docs.google.com/document/d/17rHGvJH7UTo9QaTa9w2kNJZvhkkCSQo4XrU-GWdTGQc/edit#bookmark=id.f4hvd131xsw0), there are 2 scenarios to update the namespace field:
- subject is a ServiceAccount and its name is "default"
- subject is a ServiceAccount and its name is not "default", but there is a ServiceAccount object that has the matching name.

It's a superset of what you currently have.

examples/set-namespace-simple/README.md
yuwenma: weird. GitHub seems not show the latest commit c115f5031c050c77db859820a88f43c25200ba85 which does not change this line and below (ln39) ?

yuwenma: nevermind, it's an issue on my goland.

examples/set-namespace-simple/README.md
yuwenma: fixed.

examples/set-namespace-simple/README.md
yuwenma: Fixed.

functions/go/set-namespace/README.md
yuwenma: oops.

functions/go/set-namespace/README.md
yuwenma: moved between  and

functions/go/set-namespace/transformer/namespace.go
yuwenma: Yeah, do you mind me making the change in the following PR. I have some other logic relevant to this need to add as well. Basically, the new set-namespace only expect a single namespace exists. It raises error if multiple namespace is detected. Thus, here the namespace will be updated if and only if it is a ServiceAccount kind and the existing namespace matches the found one.

mengqiy: I'm OK with doing this logic in a followup PR.

.github/workflows/ci.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
         go-version: 1.17
     - name: Install kpt
       run: |
-        go get github.com/GoogleContainerTools/kpt@main
```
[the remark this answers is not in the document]
mengqiy: The latest kpt CLI can be installed in the CI: https://github.com/GoogleContainerTools/kpt-functions-catalog/runs/5649211193?check_suite_focus=true

.github/workflows/ci.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
...
+        go get github.com/GoogleContainerTools/kpt@v1.0.0-beta.14
```
mengqiy, line 101: We can revert this line now

.github/workflows/ci.yaml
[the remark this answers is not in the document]
yuwenma: Thanks!

mengqiy (approved):

mengqiy:
The e2e CI is failing since we need the change in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/778/files#diff-ffb14a1d95b60197845214f7c645cfb58afc4d1de90eb37587be970ebe7c6f57R8 to make it passing.

mengqiy:
I split it out in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/783.
When #783 merges, this PR can be merged with a rebase.

mengqiy:
@yuwenma you can rebase this PR, it should be passing after that.

PR #771, opened 2022-03-11
feat:adds the delete-annotations function to the catalog

g-awmalik:
Selfishly needed this for myself but realized this also addresses this [issue](https://github.com/GoogleContainerTools/kpt/issues/2524)


g-awmalik:
@mengqiy - do you mind taking a look at this when you get a chance? many thanks!

functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,133 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
```
mengqiy, line 9: @yuwenma Should we ask users to try to use the SDK in the kpt-functions-sdk repo?
The logic in this function is relatively simple, I'd suggest to implement the function with the new SDK. WDYT?

yuwenma: Totally agree we should push users using the new SDK. Also, deleting annotations should be a standalone function (at least not in the curated catalog) because this is an imperative operation and theoretically should only be called by `kpt fn eval`. If it is only called by `kpt fn eval` as one-time change, `git restore` could be an easier and cleaner solution.  @g-awmalik do you mind explaining more about the use cases? If this deleting annotation is called by `kpt fn render`, please make the change on existing `set-annotation` function. But that function uses the old kustomize/kyaml sdk and the majority of the code is encapsulated as a kustomize transformer so the change may not be easy to achieve. 
@g-awmalik anyway, let us know if `git` can fulfill the needs. If not, you can either make the change in set-annotation or file an issue and we will do the implementation for you.

g-awmalik: Calling this fn imperatively is the main use case and I don't think `git restore` achieves it since that may wipe out additional (necessary) changes as well. For me specifically, I needed fn to remove annotations before publishing resources for consumers. E.g. we add annotations to Gatekeeper Constraint resources to process them but those annotations are not useful for the customers using them and would like to take them out before publishing them on GitHub. Like a mentioned above, declarative execution is not required and imperative will work just fine.

Thanks for offering the help but I can't wait until May for this. What are options for getting this out? Can we push this to the registry without making it part of the catalog? Or is modifying `set-annotations` the only feasible option here?

g-awmalik: Also, I might add that deleting annotations using `set-annotations` may go against the guidelines for fn naming i.e. <verb>-<noun> and it may not be obvious to users at first that the fn supports deleting annotations as well.

g-awmalik: @yuwenma - tagging you as a reminder.

yuwenma: Thank you for the context. I took a deeper look at the set-annotation function and the change would be non-trivial. Basically we need to refactor the code completely to use the new SDK and fork/extend the kustomize/api and kustomize/kyaml library to support object reset (create, update are supported but not delete). But I'm afraid we cannot plan this work before May.
As a short term quick solution, you can run the delete-annotation as executable in your local. Have you tried the --exec flag in `kpt fn eval`? The PR description said "needed this for myself ", you don't need to have the function be stored in the `gcr.io/kpt-fn`container registry. If you want to share the image with others, you can build it (Here's the [Dockerfile](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/build/docker/go/Dockerfile)) and upload the image to your own container registry. Would that help?

g-awmalik: Thanks @yuwenma. Is it OK if I publish this under [`contrib`](https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/contrib)

yuwenma: That's a good idea. If so, can you update this PR to use the new sdk as Mengqi suggested? Here's an example PR https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/769

morgante: @yuwenma Can you explain why you think this function should not be added to the main repo? I think it's a critical and common requirement to remove annotations, so I don't really see a need to segregate it in `contrib/`.

Also not sure how this is "imperative-only." It works well in either imperative or declarative modes.

yuwenma: Annotation is a horizontal function and we will extend set-annotation to add, update and delete annotations. So we don't plan to accept another function other than set-annotation in the curated catalog. 

"how this is imperative-only": Awais explains the use case as deleting the annotation as a one-off operation before publishing the kpt pkg. So the delete-annotation itself is not included in kptfile.

morgante: > Annotation is a horizontal function and we will extend set-annotation to add, update and delete annotations. So we don't plan to accept another function other than set-annotation in the curated catalog.

Do you have an idea for the UX to do this? I do agree that we should probably just update the existing function.

> "how this is imperative-only": Awais explains the use case as deleting the annotation as a one-off operation before publishing the kpt pkg.

That is one possible use. It is not definitive. We should not use whether someone is invoking a function imperatively or declaratively as a criterion for inclusion in the catalog.

g-awmalik:
```
Do you have an idea for the UX to do this? I do agree that we should probably just update the existing function.
```
I see two possibilities:
- we make the update to set-annotations as it is now and refactor the whole thing later (in May-ish timeframe)
- we publish delete-annotations to contrib for now and deprecate it when set-annotations is fully refactored

morgante: > we make the update to set-annotations as it is now and refactor the whole thing later (in May-ish timeframe)

This is my preference, since it isn't really blocking to have the delete-annotations function published.

g-awmalik: @yuwenma - please share your thoughts.

yuwenma: Refactoring the whole thing later is not a problem for me, but changing the set-annotation is not a easy task. I would say much harder than writing a delete function. 

- It needs to support not merely the metadata.annotations, but deleting other "annotation" fields like Deployment `spec.template.annotations` etc.  
- fork and change the third-party library kustomize/api fieldspec filter which has been highly encapsulated.
- it requires a new UX design even before refactoring and CaD team should own the UX design. 

Thus, I'd suggest publishing the delete-annotation to contrib which is much faster and easier. @g-awmalik

yuwenma: As for the UX, I've assigned myself on #2524. We should design the UX for annotation, labels and probably other meta fields altogether and build a new user pattern in the sdk. I've seen too many risk conditions and would like to think more careful about it before rushing to the design.

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
```
yuwenma, line 12: Suggest using the new SDK. Here's an example https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/769

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,133 @@
...
+		metadata, err := node.GetMeta()
+		if err != nil {
+			return nil, err
+		}
+
+		if metadata.Name == "" || metadata.Kind == "" {
+			continue
+		}
+
+		keysList := strings.Split(strings.TrimSpace(keys), annotationDelimeter)
```
yuwenma, line 67: Suggest moving this outside the for loop.

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+		for _, annotationKey := range keysList {
+			mutatedNode, err := node.Pipe(yaml.Lookup("metadata", "annotations"), yaml.FieldClearer{
```
yuwenma, line 70: Just curious, how hard do you feel about searching and using functions like `yaml.FieldClearer`?

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,133 @@
...
+	return resultItems, nil
+}
+
+// getAnnotationKeys gets the keys to delete from resources from the functionConfig
+func getAnnotationKeys(fc *yaml.RNode, keys *string) error {
+	if len(fc.GetDataMap()) < 1 {
+		return errors.New("expecting 1 or more annotation keys to delete as part of the ConfigMap")
+	}
+
+	*keys = fc.GetDataMap()[annotationKeysLiteral]
```
yuwenma, line 118: Suggest moving the annotation splitting logic here.

contrib/functions/go/delete-annotations/main.go
g-awmalik: Acked

contrib/functions/go/delete-annotations/main.go
g-awmalik: Ack

contrib/functions/go/delete-annotations/main.go
g-awmalik: Not fond of it. Piping wasn't a good experience. It's much simple now with all (most) CRUD operations available on `KubeObject`

contrib/functions/go/delete-annotations/main.go
g-awmalik: Ack

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,34 @@
+# delete-annotations
+
+## Overview
+
+<!--mdtogo:Short-->
```
mengqiy, line 5: Please ensure you are using these tags correctly. Please take a look at the template.

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Deletes the supplied annotation keys from the resource(s) in a package. It can delete multiple annotations provided as a comma separated string as part of the function config (usage instructions below).
```
mengqiy, line 7: The 2nd sentence is not necessary here. It belongs in the Usage section.

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+<!--mdtogo:Long-->
+
+## Usage
+
```
mengqiy, line 12: The 2nd sentence can be moved here.

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,34 @@
...
+The function will execute as follows:
+
+1. Searches for resources with valid metadata
```
mengqiy, line 15: The steps here are not very useful, since what this function does is simple.

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,34 @@
...
+2. Deletes the annotation keys provided in KptFile or imperatively as config parameters
+
```
mengqiy, line 17: Please follow https://raw.githubusercontent.com/GoogleContainerTools/kpt-functions-catalog/master/functions/go/_template/README.md to ensure you have a `FunctionConfig` section.

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		for _, annotationKey := range annotationKeys {
+			removed, err := o.Remove("metadata", "annotations", annotationKey)
+			if err != nil {
+				return nil, err
+			}
+
+			if removed {
+				itemFilePath := o.GetAnnotations()["internal.config.kubernetes.io/path"]
```
mengqiy, line 60: There is a helper `o.GetPathAnnotation()`. you can use that.

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,112 @@
...
+				if itemFilePath == "" {
+					itemFilePath = o.GetAnnotations()["config.kubernetes.io/path"]
```
mengqiy, line 62: No need to do this any more

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,112 @@
...
+				})
+			}
+		}
+	}
+
+	if len(resultItems) > 0 {
+		infoResultSlice := []*fn.Result{}
+		infoResultSlice = append(infoResultSlice, &fn.Result{
+			Severity: fn.Info,
+			Message:  "The following annotations were deleted from the resources",
```
mengqiy, line 80: This message is not very useful unless it tells how many places it has updated.

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,112 @@
...
+		})
+
+		resultItems = append(infoResultSlice, resultItems...)
+	} else if len(resultItems) == 0 {
+		resultItems = append(resultItems, &fn.Result{
+			Message:  "None of the resources had the provided annotations to delete",
+			Severity: fn.Warning,
```
mengqiy, line 87: Maybe `Info` is OK

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,112 @@
...
+	if annotationKeysString == "" {
+		return nil, fmt.Errorf("%s was not provided as part of the config or paramters to the function", annotationKeysLiteral)
+	}
+
+	annotationKeys := strings.Split(strings.TrimSpace(annotationKeysString), annotationDelimeter)
+	return annotationKeys, nil
+}
+
+// getErrorItem returns the item for an error message
+func getErrorItem(errMsg string) *fn.Result {
```
mengqiy, line 107: The `fn` already provides something similar. Please use that

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+To execute `delete-annotations` declaratively include the function in kpt package pipeline as follows:
+```yaml
+...
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn-contrib/delete-annotations:unstable
+      configMap:
+        annotationKeys: annotation.to.delete,another.annotation.to.delete
```
mengqiy, line 30: Does k8s allow `.` in annotations?
It's more common to use dash `-`. If you agree, please update other places as well.

contrib/functions/go/delete-annotations/README.md
g-awmalik: Acked

contrib/functions/go/delete-annotations/README.md
g-awmalik: Acked

contrib/functions/go/delete-annotations/README.md
g-awmalik: Acked

contrib/functions/go/delete-annotations/README.md
g-awmalik: Acked

contrib/functions/go/delete-annotations/README.md
g-awmalik: Acked

contrib/functions/go/delete-annotations/main.go
g-awmalik: Acked.

contrib/functions/go/delete-annotations/README.md
g-awmalik: k8s allows `.` in annotations. I've verified that.

contrib/functions/go/delete-annotations/main.go
g-awmalik: Acked

contrib/functions/go/delete-annotations/main.go
g-awmalik: This message won't ever appear by itself. It will have a list of resources underneath it where the annotations were deleted and that is constructed a few lines before this and appended to this.

contrib/functions/go/delete-annotations/main.go
g-awmalik: Ack.

contrib/examples/delete-annotations-simple/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Invoke the function by running the following command:
+
+```shell
+$ kpt fn render delete-annotations-simple
+```
+
+### Expected result
+
+One of the two resources in `resources.yaml` should have been mutated with the annotation `annotation.to.delete` removed from `metadata.annotations`. There shouldn't be any changes to the second resource as it doesn't have the supplied annotation.
```
mengqiy, line 25: > `annotation.to.delete`

Please change it to something that doesn't look like a path. ``annotation-to-delete` is a better choice.

contrib/examples/delete-annotations-simple/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```yaml
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: K8sRestrictRoleBindings
```
mengqiy, line 29: I guess we don't need to copy the output here, since what we do here is not complex and static content here may get out-of-date.

contrib/examples/delete-annotations-simple/resources.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: K8sRestrictRoleBindings
```
mengqiy, line 2: I'd suggest using 2 easier-to-understand Kinds here. Maybe some builtin types like ConfigMap or Pod

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Overview
+
+<!--mdtogo:Short-->
+
+Deletes the supplied annotation keys from the resource(s) in a package.
+
+<!--mdtogo-->
+
+This function helps users remove annotations that are not necessary for deployment
```
mengqiy, line 11: > This function helps users remove annotations that are not necessary for deployment across a package.

This is one of the use case.
You may want to write it like:
```
One of the use case is .... 
```

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,51 @@
+# delete-annotations
+
+## Overview
+
+<!--mdtogo:Short-->
+
+Deletes the supplied annotation keys from the resource(s) in a package.
```
mengqiy, line 7: > in a package

IMO not necessary

contrib/functions/go/delete-annotations/README.md
mengqiy: Got it.
I'd still suggest to make use `-` in between which is more common. Otherwise, when you mention it in the doc, it looks like a path.

contrib/functions/go/delete-annotations/README.md
```diff
@@ -0,0 +1,51 @@
...
+        annotationKeys: annotation.to.delete,another.annotation.to.delete
+...
+```
+
+### FunctionConfig
+
+This function takes the annotation key names as part of the function config parameter
+`annotationKeys` where the key names can be provided as comma separated values as follows:
+
+`annotationKeys=annotation.key.1,annotation.key.2`
```
mengqiy, line 44: Styling issue: I'd suggest using `-` here and below.

contrib/functions/go/delete-annotations/go.mod
```diff
@@ -0,0 +1,32 @@
+module github.com/GoogleContainerTools/kpt-functions-catalog/contrib/functions/go/delete-annotations
+
+go 1.17
+
+require (
+	github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220322205615-b3b2aa33ba46
```
mengqiy, line 6: Please ensure you are using the latest.

contrib/functions/go/delete-annotations/main.go
mengqiy: It appears it doesn't have it our golang SDK.
I proposed `GetPathAnnotation` in OSS SDK.
@yuwenma Do we plan to add it in our SDK?

contrib/functions/go/delete-annotations/main.go
mengqiy: Does it make sense to provide a counter here?

g-awmalik: Sure. I've added a counter for the no. of resources pruned.

contrib/examples/delete-annotations-simple/README.md
g-awmalik: Acked

contrib/examples/delete-annotations-simple/README.md
g-awmalik: Removed.

contrib/examples/delete-annotations-simple/resources.yaml
g-awmalik: Acked. Used a ConfigMap and Namespace instead.

contrib/functions/go/delete-annotations/README.md
g-awmalik: Acked

contrib/functions/go/delete-annotations/README.md
g-awmalik: Acked

contrib/functions/go/delete-annotations/go.mod
g-awmalik: Updated.

yuwenma:
> It appears it doesn't have it our golang SDK.

@mengqiy @g-awmalik I'll add the annotation

g-awmalik:
Thanks @yuwenma. Let me know when its ready.

mengqiy (approved):

contrib/functions/go/delete-annotations/main.go
```diff
@@ -0,0 +1,108 @@
...
+				return nil, err
+			}
+
+			if removed {
+				if o.GetName() != currObjName {
+					prunedObjCtr++
+					currObjName = o.GetName()
+				}
+
+				itemFilePath := o.GetAnnotations()["internal.config.kubernetes.io/path"]
```
mengqiy, line 67: Please add a TODO here to remind us to switch over.

mengqiy:
@yuwenma PTAL to see if you have any concerns or objections.

PR #777, opened 2022-03-16
fix: use default value for build args

sdowell:
On some versions of bash an error was being thrown due to the build_args
being interpreted as an unbound variable. This change adds a default
empty string value to the variable expansion so that this error does not
occur.

Issues: GoogleContainerTools/kpt#2902

### DIFF
--- a/scripts/docker.sh
+++ b/scripts/docker.sh
@@ -51,10 +51,11 @@ function docker_build {
   [[ -f "${override_dockerfile}" ]] && dockerfile="${override_dockerfile}"
   [[ -f "${dockerfile}" ]] || err "Dockerfile does not exist: ${dockerfile}"
 
+  # Use + conditional parameter expansion to protect from unbound array variable
   docker build \
     -t "${GCR_REGISTRY}/${name}:${UNSTABLE_TAG}" \
     -f "${dockerfile}" \
-    "${build_args[@]}" \
+    "${build_args[@]+"${build_args[@]}"}" \
     "${function_dir}"
 }
 


mengqiy:
The CI is failing with the same error.

scripts/docker.sh
```diff
@@ -54,7 +54,7 @@ function docker_build {
   docker build \
     -t "${GCR_REGISTRY}/${name}:${UNSTABLE_TAG}" \
     -f "${dockerfile}" \
-    "${build_args[@]}" \
+    "${build_args[@]+"${build_args[@]}"}" \
```
mengqiy, line 57: Can you please explain what are we doing here?

sdowell: The syntax is a bit arcane, I agree :)

From https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html:
```
${parameter:+word}
If parameter is null or unset, nothing is substituted, otherwise the expansion of word is substituted.
```

So basically if `build_args` is unset or null, it will substitute nothing. Otherwise, it will expand the `build_args` array.

According to [this stackoverflow post](https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u) this is the safest way to achieve this across bash versions.

I can also add a comment to help future readers

mengqiy: Thanks for you explanation

mengqiy (approved):
I approved it now.
Please feel free to merge it after adding the comment.
Thank you!

PR #778, opened 2022-03-16
refactor starlark function

mengqiy:
This PR refactors the starlark function to make its logic importable by Porch.

This PR needs https://github.com/GoogleContainerTools/kpt/pull/2927 to have the tests passing.


mengqiy:
This PR needs https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/534 to make the CI passing.

functions/go/starlark/starlark/config.go
```diff
@@ -124,39 +124,58 @@ func validateStarlarkRun(sr *StarlarkRun) error {
...
+	return nil
+}
 
+func (sfc *StarlarkFnConfig) Decode(obj *fn.KubeObject) error {
+	if obj == nil {
+		return fn.ErrMissingFnConfig{}
+	}
 	switch {
-	case meta.APIVersion == configMapApiVersion && meta.Kind == string(configMapKind):
+	case obj.GetAPIVersion() == configMapApiVersion && obj.GetKind() == string(configMapKind):
```
yuwenma, line 169: nit `case obj.IsGVK(configMapApiVersion, configMapKind):`

functions/go/starlark/starlark/processor.go
```diff
@@ -0,0 +1,37 @@
+package starlark
+
+import (
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+type Processor struct{}
+
+func (gkp *Processor) Process(resourceList *fn.ResourceList) error {
```
yuwenma, line 9: Just curious, what is `gkp`?

functions/go/starlark/starlark/processor.go
```diff
@@ -0,0 +1,37 @@
...
+	err := func() error {
+		sfc := &StarlarkFnConfig{}
```
yuwenma, line 11: I found `StarlarkFnConfig.Transform` confusing because `Transform` is the transformer's main function, but `StarlarkFnConfig` more looks like the starlark functionConfig. 

Also, is that possible to use `ResourceListProcessorFunc` but not a custom `Processor`?

mengqiy:
The 2nd the commit contains the new changes since last review.

functions/go/starlark/starlark/processor.go
mengqiy: I renamed `Transform` to `Evaluate`. Hope it makes more sense now.
I also updated it to use `ResourceListProcessorFunc`.

functions/go/starlark/starlark/processor.go
```diff
@@ -0,0 +1,37 @@
+package starlark
+
+import (
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+type Processor struct{}
```
yuwenma, line 7: remove?

functions/go/starlark/starlark/processor.go
```diff
@@ -0,0 +1,37 @@
...
+		if err := sfc.Default(); err != nil {
+			return err
+		}
+		if err := sfc.Validate(); err != nil {
+			return err
+		}
+		if te := sfc.Evaluate(resourceList); te != nil {
+			return te
+		}
+		return nil
```
yuwenma, line 24: Suggest adding docstring to the public functions since they are not following the typical `Config` `Transform` convention.

Actually, do you think we can merge Decode, Default and Validate to `Config`? Functions here are served as example code for SDK users, so they should follow the same pattern.

functions/go/starlark/starlark/processor.go
yuwenma: sorry, I was suggesting to keep `Transform` but rename "StarlarkFnConfig" to something other than`FnConfig`. How about `StarlarkEvaludator` or `StarlarkFn`?

yuwenma (approved):
LGTM. some minor comments about naming conventions.

mengqiy:
@yuwenma I did some further refactor. PTAL

functions/go/starlark/starlark/config.go
```diff
@@ -0,0 +1,116 @@
...
+
+	configMapApiVersion = "v1"
+	configMapKind       = "ConfigMap"
+
+	sourceKey = "source"
+
+	defaultProgramName = "stalark-function-run"
+)
+
+type ConfigMap struct {
```
yuwenma, line 26: I might miss something, but why do we have two Config and Transform functions?

functions/go/starlark/starlark/processor.go
```diff
@@ -0,0 +1,43 @@
...
+			if err := sr.Config(resourceList.FunctionConfig); err != nil {
+				return err
+			}
+			return sr.Transform(resourceList)
+		default:
+			return fmt.Errorf("`functionConfig` must be either %v or %v, but we got: %v",
+				schema.FromAPIVersionAndKind(configMapApiVersion, configMapKind).String(),
+				schema.FromAPIVersionAndKind(starlarkRunAPIVersion, starlarkRunKind).String(),
+				schema.FromAPIVersionAndKind(resourceList.FunctionConfig.GetAPIVersion(), resourceList.FunctionConfig.GetKind()).String())
+		}
```
yuwenma, line 30: Some suggestions to simplify the logic and avoid code duplication (Config and Transform) and use the new KubeObject functions rather than `k8s.io/api/core/v1` or `kustomize/yaml` `Marshal`.
```go
type StarlarkFn struct {
     Name string
     Source string
     Params map[string]interface{}
}
func (s *StarlarkFn) Config(fn *fn.KubeObject) {
      s.Name = fnCfg.GetName()
      switch {
      case fnCfg.IsGVK(configMapApiVersion, configMapKind):
            s.Source = fnCfg.GetStringOrDie("data.source")
      case fnCfg.IsGVK(starlarkRunAPIVersion, starlarkRunKind):
            s.Source = fnCfg.GetStringOrDie("source")
            var params map[string]interface{}
            found, err = fnCfg.Get(&params, "params")
            s.Params = params
      default:
            error
      }  
}
func (s *StarlarkFn) Transform() {
    	return evaluateStarlarkToTransformResourceList(sr.Name, sr.Source, rl)
}
func Process() {
      err := func(){
             s := StarlarkFn{}
             s.Config(resourceList.FunctionConfig) 
             s.Transform(resourceList)    
      }
}
```

yuwenma: then we don't need the `ConfigMap` and `StarlarkRun` struct, right?

mengqiy: If we remove `StarlarkRun`, developers who want to do something programmatically with `StarlarkRun` type will need to define this type on their own. e.g. build another function that construct the `starlarkRun` functionConfig.

IMO when users work with types, I'd suggest using strongly typed object if possible since it can provide compile-time check. If strongly typed objects are not available, I'd suggest using unstructured approach next like `KubeObject` we have in the SDK.

mengqiy:
@yuwenma I updated it according to our discussion yesterday.

mengqiy:
We should update the expected test results in a separate PR.

.github/workflows/ci.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
         go-version: 1.17
     - name: Install kpt
       run: |
-        go install github.com/GoogleContainerTools/kpt@main
+        go install github.com/GoogleContainerTools/kpt@0faf830e9f0ca2742bf15f3bb4560763eaa5d489
```
mengqiy, line 101: @droot I pinned kpt to a version to unblock this PR.

functions/go/starlark/starlark/config.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+			sr.Params[k] = v
+		}
+	case fnCfg.IsGVK(starlarkRunAPIVersion, starlarkRunKind):
+		if err := fnCfg.As(sr); err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("`functionConfig` must be either %v or %v, but we got: %v",
+			schema.FromAPIVersionAndKind(configMapApiVersion, configMapKind).String(),
```
yuwenma, line 58: ah, good to know this!

functions/go/starlark/starlark/config.go
```diff
@@ -0,0 +1,109 @@
...
+			schema.FromAPIVersionAndKind(starlarkRunAPIVersion, starlarkRunKind).String(),
+			schema.FromAPIVersionAndKind(fnCfg.GetAPIVersion(), fnCfg.GetKind()).String())
+	}
+
+	// Defaulting
+	if sr.Name == "" {
+		sr.Name = defaultProgramName
+	}
+	// Validation
+	if sr.Source == "" {
```
yuwenma, line 68: nit: suggest adding this logic inside the switch because ConfigMap and StarlarkRun have different source path (`data.source` vs `source`)

PR #779, opened 2022-03-17
fix: Skip clusterless resources in exported Terraform

morgante:
Since Terraform is a clusterless service, the exported config should exclude any resources which are specific to Config Controller.

Context: b/225041968


mengqiy (approved):
LGTM

PR #780, opened 2022-03-17
replacements should implement processor interface

natasha41575:
realized that the replacements type did not implement the processor interface. This fixes that, so that we can add helper functions to the sdk to run it as a builtin.


natasha41575:
Actually never mind, I realized this is not necessary.

PR #781, opened 2022-03-17
update fn sdk in apply-replacements to better handle panic

mengqiy:
To pick up commit in https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/531


yuwenma (approved):
Sweet!

natasha41575 (approved):

PR #782, opened 2022-03-21
[apply-replacements] return all function errors

natasha41575:
The function needs to return errors, otherwise kpt will assume the function executed successfully. 

Combined with https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/534, this fixes https://github.com/GoogleContainerTools/kpt/issues/2901.

Before: 

```
Package "apply-replacements-error": 
[RUNNING] "gcr.io/kpt-fn/apply-replacements:unstable"
[PASS] "gcr.io/kpt-fn/apply-replacements:unstable" in 1.1s
  Results:
    [error]: nothing selected by Pod.[noVer].[noGrp]/my-po.[noNs]:spec
```

After:

```
Package "apply-replacements-error": 
[RUNNING] "gcr.io/kpt-fn/apply-replacements:unstable"
[FAIL] "gcr.io/kpt-fn/apply-replacements:unstable" in 1.1s
  Results:
    [error]: nothing selected by Pod.[noVer].[noGrp]/my-po.[noNs]:spec
  Exit code: 1
```

### DIFF
--- a/functions/go/apply-replacements/replacements/replacements.go
+++ b/functions/go/apply-replacements/replacements/replacements.go
@@ -2,7 +2,6 @@ package replacements
 
 import (
 	"fmt"
-
 	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
 	"sigs.k8s.io/kustomize/api/filters/replacement"
 	"sigs.k8s.io/kustomize/api/types"
@@ -24,7 +23,7 @@ type Replacements struct {
 // Config initializes Replacements from a functionConfig fn.KubeObject
 func (r *Replacements) Config(functionConfig *fn.KubeObject) error {
 	if functionConfig.GetKind() != fnConfigKind || functionConfig.GetAPIVersion() != fnConfigApiVersion {
-		return fmt.Errorf("received functionConfig of kind %s and apiVersion %s, " +
+		return fmt.Errorf("received functionConfig of kind %s and apiVersion %s, "+
 			"only functionConfig of kind %s and apiVersion %s is supported",
 			functionConfig.GetKind(), functionConfig.GetAPIVersion(), fnConfigKind, fnConfigApiVersion)
 	}
@@ -43,11 +42,7 @@ func (r *Replacements) Process(rl *fn.ResourceList) error {
 	}
 	transformedItems, err := r.Transform(rl.Items)
 	if err != nil {
-		rl.Results = append(rl.Results, &fn.Result{
-			Message: err.Error(),
-			Severity: fn.Error,
-		})
-		return nil
+		return err
 	}
 	rl.Items = transformedItems
 	return nil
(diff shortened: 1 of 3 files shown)


functions/go/apply-replacements/replacements/replacements.go
```diff
@@ -43,11 +42,7 @@ func (r *Replacements) Process(rl *fn.ResourceList) error {
 	}
 	transformedItems, err := r.Transform(rl.Items)
 	if err != nil {
-		rl.Results = append(rl.Results, &fn.Result{
```
natasha41575, line 46: I'm using `fn.AsMain` which already does this step, so we don't need to do it here.

natasha41575:
This PR should be rebased on top of https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/534 when it is in.

functions/go/apply-replacements/replacements/replacements.go
droot: I commented in the other PR. Does it make sense to change the `Process(rl)` to return `result, err := Process(rl)` to avoid overloading of return err as `result` and `error` ? If result is returned, it is always added to the `rl` and returned error is used to indicate overall success or failure.

natasha41575: Yes, I agree that having `Process` return both `result` and `error` will simplify the `AsMain` code. We would have to change this in the definition of `ResourceListProcessor` and make sure other functions implementing `Process` make the same change.

@mengqiy had a brief discussion and he brought up that the KRM function spec indicates that when a function fails, it should exit with a nonzero exit code. That means if Process runs into an error here, that error should be returned. 

That would mean that `Process` should record the error in `result`, and then return `result, err`.  i.e. `return &fn.Result{Message: err.Error(), Severity: fn.Error}, err`.

Then in `AsMain`, we can just append the result, do the yaml processing, and return the err from the function.

droot: Totally agree with the approach described above. Since, we only have two functions written using new SDK, changing the `ResourceListProcessor` definition shouldn't be a big deal, so we should go ahead and do it.

natasha41575: I pushed a commit that pins this PR to my PR in the SDK https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/534, which makes `Process` return both Result and error, so we can see what it will look like. 

We should wait for https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/534 to go in, before this one (and remove the pinning to my fork).

natasha41575:
Marking this as a draft, pending the decision made in https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/534

natasha41575:
@yuwenma @mengqiy PTAL. This is the last change needed for apply-replacements so it can be released.

yuwenma (approved):

PR #785, opened 2022-03-23
feat: Refactor set-namespace

yuwenma:
New logic
-  `set-namespace` supports three modes
     - support a new `namespaceSelector` parameter in the functionConfig. If given, the  `namespaceSelector`  value is the "oldNs"
     - allow no more than one `namespace` object in the resourcelist.items. If found (and `namespaceSelector` is not given), use the namespace `metadata.name` value as "oldNs"
     -  If either `namespaceSelector` nor `namespace` object are found, we require all namespace scoped resources have the same namespace value, which is the "oldNs". 
- `set-namespace` only replace the namespace fields which match the "oldNs" to the new Namespace.        
- Skip resources which have annotation `config.kubernetes.io/local-config`

"ClusterRoleBinding" and "RoleBinding" changes
- Remove code about `RoleBinding` `subjects` namespace check. Because `RoleBinding` is namespace-scoped so it only works for resources under the same namespace. Itself `metadata.namespace` should be updated.
- Simplify the `ClusterRoleBinding` check to not check subjects "ServiceAccount" kind. For clusterrolebinding, only ServiceAccount may be namespace scoped.

New Code Structure
- Define `VisitNamespace` with callback function as params to iterations `resourcelist.items`. 
- Define `Namespace` struct to provide callback functions for different iterations, including `FindAllNamespaces` to check duplication, `UpdateNamespace` to update namespace value via string pointer, `GetDependsOnMap` to build a key value pairs by resources `depends-on` annotation, and `UpdateAnnotation` to update the namespace segment in the `depends-on` annotation.  
- Call the new function in go-sdk 
    - GetSlice to handle the ClusterRoleBinding "subjects" modification
    - GetXXXOrDie to get a KRM resource component from a field path, and panic if field path is wrong (expecting the go SDK to handle the panic and gives user friendly guidance)
    - SetXXXOrDie
    - `IsGVKMatch` for duck typing check.
    
- Update the KRM resources' namespace value via string pointer.

Rewrite the README


functions/go/set-namespace/go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-na
 
 go 1.17
 
-require github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220316202203-f9115a993ebd
+require github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220322205615-b3b2aa33ba46
```
justinsb, line 5: Aside: This probably should be a replace to `../fn` (?)  But not really relevant to this PR.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -98,45 +71,152 @@ func (p *NamespaceTransformer) Config(o *fn.KubeObject) error {
...
 func (p *NamespaceTransformer) Transform(objects []*fn.KubeObject) {
 	p.SetDependsOnMap(objects)
 	for _, o := range objects {
 		switch {
+		// Skip local resource which `kpt live apply` skips.
+		case o.IsLocalConfig():
+			continue
 		case o.IsGVK("v1", "Namespace"):
-			o.SetName(p.Namespace)
+			o.SetName(p.NewNamespace)
```
justinsb, line 84: I do think we want to check whether the old namespace matches.  If we feel the semantics of this function are sufficiently different, we might just want to make a new function and deprecate this one.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -98,45 +71,152 @@ func (p *NamespaceTransformer) Config(o *fn.KubeObject) error {
...
-			}
+			p.updateNamespace(o, "spec", "service", "namespace")
 		case o.GetKind() == "ClusterRoleBinding" || o.GetKind() == "RoleBinding":
 			var subjects []*Subject
 			o.GetOrDie(&subjects, "subjects")
 			for _, s := range subjects {
-				if s.Name == "default" {
-					s.Namespace = p.Namespace
+				if s.Kind == "ServiceAccount" {
+					if p.isSelectorMode() && s.Namespace != p.MatchingNamespace {
```
justinsb, line 94: Might be nice to have a version of the `updateNamespace` helper that takes a `*string`, just so it's obvious that this is the same transformation.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -98,45 +71,152 @@ func (p *NamespaceTransformer) Config(o *fn.KubeObject) error {
...
-		// Shall we accept custom fieldspec?
-		// p.SetCustomFieldSpec(o)
 		p.UpdateAnnotation(o)
 	}
 }
 
+// updateNamespace updates the namespace in three modes:
+// - Restrict mode: always update namespace from the given `fieldpaths`
+// - NsObjectSelector mode: update the namespace which matches the Namespace object found in resourceList.items.
+// - CustomSelector mode: update the namespace which matches the `namespaceSelector` in resourceList.functionConfig.
```
justinsb, line 116: Are NsObjectSelector and CustomSelector different for `updateNamespace`?  I feel like the logic could just be "if MatchingNamespace != ""`, then the old namespace must match.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -98,45 +71,152 @@ func (p *NamespaceTransformer) Config(o *fn.KubeObject) error {
...
+			p.updateNamespace(o, "spec", "conversion", "webhook", "clientConfig", "service", "namespace")
 		case o.IsGVK("apiregistration.k8s.io/v1", "APIService"):
-			err := o.Set(p.Namespace, "spec", "service", "namespace")
-			if err != nil {
-				p.Errors = append(p.Errors, err.Error())
-			}
+			p.updateNamespace(o, "spec", "service", "namespace")
 		case o.GetKind() == "ClusterRoleBinding" || o.GetKind() == "RoleBinding":
 			var subjects []*Subject
 			o.GetOrDie(&subjects, "subjects")
```
justinsb, line 91: Do we have GetSlice?  Because if we do then we don't need to treat this specially...

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -98,45 +71,152 @@ func (p *NamespaceTransformer) Config(o *fn.KubeObject) error {
...
+func (p *NamespaceTransformer) listCurrentNamespaces(objects []*fn.KubeObject) (map[string]int, map[string]bool) {
+	nsCount := map[string]int{}
+	fromResources := map[string]bool{}
+	for _, o := range objects {
+		// Skip local resource which `kpt live apply` skips.
+		if o.IsLocalConfig() {
+			continue
+		}
+		switch {
+		case o.IsGVK("v1", "Namespace"):
```
justinsb, line 196: I think it would be great to combine the logic with the transformation logic.

A good starting point might be something like:

`func VisitNamespaces(objects []*fn.KubeObject, visitor func (ns string) string)`

(I can see us replacing `ns string` with something like `v Namespace` and supporting `v.Set()`, `v.Node()`, `v.FieldPath()` etc, but I don't think we need to do that yet)

functions/go/set-namespace/transformer/namespace.go
yuwenma: Good point! yeah, added.

functions/go/set-namespace/go.mod
justinsb: Oh whoops - never mind - I missed that this was a different repo!

functions/go/set-namespace/transformer/namespace.go
yuwenma: I like the idea!

functions/go/set-namespace/transformer/namespace.go
yuwenma: Thanks! So in `transform` we always update the namespace object. I like the simplification

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -98,70 +70,208 @@ func (p *NamespaceTransformer) Config(o *fn.KubeObject) error {
...
+	setter        func(newNs string)
+	annoGetter    func() string
+	annoSetter    func(newDependsOnAnno string)
+}
+
+func (n *Namespace) Set(newNs string) {
+	n.setter(newNs)
+}
+
+func (n *Namespace) SetDependsOnAnno(newDependsOnAnno string) {
```
justinsb, line 271: Nit: I would just spell out `SetDependsOnAnnotation` etc rather than using the "Anno" abbreviation.

You _could_ argue that annotation is just an implementation detail, and make it something like `SetDependsOn(gvknn dependsOn)` -  that's consistent with e.g. a future method that gets all dependencies of an object.

justinsb (approved):

functions/go/set-namespace/transformer/namespace.go
yuwenma: renamed. I won't argue but agree +50 on the full spell. That's exactly the buggy moment I got confused when first see the short form.

yuwenma:
> I like this callback approach a lot 👍

Thank you so much for the suggestions!

functions/go/set-namespace/README.md
```diff
@@ -17,68 +17,114 @@ Namespaces are often used in the following scenarios:
...
+  namespace: newNamespace
+  namespaceSelector: example # update namespace whose value is "example" to "newNamespace"
+```
+
+SetNamespace as functionConfig
+```yaml
+apiVersion: fn.kpt.dev/v1alpha1
+kind: SetNamespace
+namespace: newNamespace
+namespaceSelector: example # update namespace whose value is "example" to "newNamespace"
```
mengqiy, line 120: Maybe `namespaceMatcher` or `originalNamespace` is a better name.
You may want to hear other folks' opinion.

functions/go/set-namespace/README.md
```diff
@@ -17,68 +17,114 @@ Namespaces are often used in the following scenarios:
...
+  if the field is set and matches the "oldNamespace"
+- This function updates `APIService` `spec/service/namespace` field if the field is set and matches the "oldNamespace"
+- This function updates the KRM resources annotation `config.kubernetes.io/depends-on` if this annotation contains the 
+  matching namespace.
+
+### FunctionConfig
+
+This function supports the default `ConfigMap` as function config and a custom `SetNamespace`. See below examples
+
+ConfigMap as functionConfig
```
mengqiy, line 39: Styling nit: `ConfigMap` should have backtick when it's a Kind name.

Similarly for `SetNamespace` and `Namespace`. Please audit this doc.

functions/go/set-namespace/README.md
```diff
@@ -17,68 +17,114 @@ Namespaces are often used in the following scenarios:
...
-  name: confluent-operator
-  apiGroup: rbac.authorization.k8s.io
+kind: ConfigMap
+data:
+  namespace: newNamespace # required
+  namespaceSelector: example # update namespace whose value is "example" to "newNamespace"
 ```
 
-This function can be used both declaratively and imperatively.
+SetNamespace as functionConfig
```
mengqiy, line 48: Same here.

functions/go/set-namespace/README.md
```diff
@@ -17,68 +17,114 @@ Namespaces are often used in the following scenarios:
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+data:
+  namespace: newNamespace # update all namespace fields to "newNamespace"
+```
+
+##### DefaultNamespace Mode
+
+The input `resourcelist.items` contains one and only one namespace object. The function matches the namespace `metadata.name`
```
mengqiy, line 74: > contains one and only one namespace object

contains **one and only one** `Namespace` object

functions/go/set-namespace/README.md
```diff
@@ -17,68 +17,114 @@ Namespaces are often used in the following scenarios:
...
+apiVersion: fn.kpt.dev/v1alpha1
+kind: SetNamespace
+namespace: newNamespace
+namespaceSelector: example # update namespace whose value is "example" to "newNamespace"
 ```
 
+### DependsOn annotation
+
+DependsOn annotation is a [kpt feature](https://kpt.dev/reference/annotations/depends-on/). This function updates the 
+namespace segment in a depends-on annotation if matches oldNs
```
mengqiy, line 126: > oldNs

This is informal.

functions/go/set-namespace/README.md
```diff
@@ -17,68 +17,114 @@ Namespaces are often used in the following scenarios:
...
+    name: the-service1
+    namespace: example  # updated to "newNs"
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: the-service2
+    namespace: irrelevant # skip since namespace does not match "example".
+```
+
+##### Selector Mode
```
mengqiy, line 102: We should have at least one example for each mode.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -65,103 +40,229 @@ func SetNamespace(rl *fn.ResourceList) error {
...
+	return fromResources[0], true
 }
 
-func (p *NamespaceTransformer) UpdateAnnotation(o *fn.KubeObject) {
-	anno, ok := o.GetAnnotations()[dependsOnAnnotation]
-	if !ok {
-		return
+type NamespaceTransformer struct {
+	NewNamespace      string
+	namespaceSelector string
```
mengqiy, line 229: Why this field is private but `Errors` is public?

functions/go/set-namespace/README.md
yuwenma: Sure, I prefer `namespaceMatcher` over `orignialNamespace`. The latter doesn't tell why the functions need the orignal namespace.

yuwenma: looks like [`namespaceSelector`](https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors) indeed can confuse users.

functions/go/set-namespace/README.md
yuwenma: done.

functions/go/set-namespace/README.md
yuwenma: done.

functions/go/set-namespace/README.md
yuwenma: done.

functions/go/set-namespace/README.md
yuwenma: Done.

functions/go/set-namespace/README.md
yuwenma: I'll do that in a follow up PR.

mengqiy (approved):

PR #787, opened 2022-03-23
Use the same version as in the sdk in set-namespace fn

mengqiy:
https://github.com/kubernetes/kube-openapi/pull/285 will break
things if we are using inconsitent version of kube-openapi


yuwenma (approved):
Sweet!

PR #792, opened 2022-03-25
preparing for go 1.18

mikebz:
Go 1.18 is a bit stricter about go install so this is an update to get rid of `go get` everywhere we have it. 

Also it seems like blackfriday is not really used anywhere and it gives us problems because it's not go install compatible.  Yanking that.


mengqiy (approved):
LGTM
Thanks

PR #793, opened 2022-03-29
sync apply-replacement doc

mengqiy:


natasha41575 (approved):
Thanks! PTAL https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/782 before releasing this function.

PR #794, opened 2022-03-30
[starlark] builtin match_gvkn function

natasha41575:
This adds a builtin to the starlark runtime called `match_gvkn`. Example usage in the value propagation demo: https://github.com/natasha41575/rolebinding-replacement/blob/9c26aedae15c36dfe9912ffa9706b4d49288a1f2/starlark-and-replacements-function/generate-rolebinding-group.yaml#L14

This needs to be rebased on https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/778 for CI to pass (some dependency issues).

Some questions:
- What do we want to call the module? I named it "krmfn", but wondering should it be "fn" to be consistent with the go SDK?
- Eventually we are moving this code and the forked starlark lib here to the kpt-functions-sdk. If I should change the directory structure to make moving it easier, let me know. Or if we want to move it over first before making this change, let me know.
- How do we document these builtin starlark functions? Docstrings in the Go code, a page on the kpt website, somewhere else?

cc @yuwenma

### DIFF
--- a/functions/go/starlark/krmfn/krmfn.go
+++ b/functions/go/starlark/krmfn/krmfn.go
@@ -26,8 +26,6 @@ func matchGVKN(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple
 	if err != nil {
 		return nil, err
 	}
-	if obj.GetAPIVersion() == apiVersion && obj.GetKind() == kind && obj.GetName() == name {
-		return starlark.Bool(true), nil
-	}
-	return starlark.Bool(false), nil
+	match := obj.GetAPIVersion() == apiVersion && obj.GetKind() == kind && obj.GetName() == name
+	return starlark.Bool(match), nil
 }
--- a/functions/go/starlark/krmfn/krmfn.go
+++ b/functions/go/starlark/krmfn/krmfn.go
@@ -11,21 +11,48 @@ const ModuleName = "krmfn.star"
 var Module = &starlarkstruct.Module{
 	Name: "krmfn",
 	Members: starlark.StringDict{
-		"match_gvkn": starlark.NewBuiltin("match_gvkn", matchGVKN),
+		"match_gvk":       starlark.NewBuiltin("match_gvk", matchGVK),
+		"match_name":      starlark.NewBuiltin("match_name", matchName),
+		"match_namespace": starlark.NewBuiltin("match_namespace", matchNamespace),
 	},
 }
 
-func matchGVKN(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+func matchGVK(_ *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var resourceList starlark.Value
-	var apiVersion, kind, name string
-	if err := starlark.UnpackPositionalArgs("match_gvkn", args, kwargs, 4,
-		&resourceList, &apiVersion, &kind, &name); err != nil {
+	var apiVersion, kind string
+	if err := starlark.UnpackPositionalArgs("match_gvk", args, kwargs, 3,
+		&resourceList, &apiVersion, &kind); err != nil {
 		return nil, err
 	}
 	obj, err := fn.ParseKubeObject([]byte(resourceList.String()))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


tests/starlark/load-library/fn-config.yaml
```diff
@@ -18,3 +18,4 @@ source: |
   load('time.star', 'time')
   load('xlsx.star', 'xlsx')
   load('zipfile.star', 'ZipFile')
+  load('krmfn.star', 'krmfn')
```
mengqiy, line 21: I'd suggest adding an simple example using it.

functions/go/starlark/krmfn/krmfn.go
```diff
@@ -0,0 +1,31 @@
...
+	var apiVersion, kind, name string
+	if err := starlark.UnpackPositionalArgs("match_gvkn", args, kwargs, 4,
+		&resourceList, &apiVersion, &kind, &name); err != nil {
+		return nil, err
+	}
+	obj, err := fn.ParseKubeObject([]byte(resourceList.String()))
+	if err != nil {
+		return nil, err
+	}
+	match := obj.GetAPIVersion() == apiVersion && obj.GetKind() == kind && obj.GetName() == name
```
mengqiy, line 29: We can use `obj.IsGVK(apiVersion, kind)`

functions/go/starlark/krmfn/krmfn.go
```diff
@@ -0,0 +1,31 @@
...
+	Name: "krmfn",
+	Members: starlark.StringDict{
+		"match_gvkn": starlark.NewBuiltin("match_gvkn", matchGVKN),
+	},
+}
+
+func matchGVKN(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var resourceList starlark.Value
+	var apiVersion, kind, name string
+	if err := starlark.UnpackPositionalArgs("match_gvkn", args, kwargs, 4,
```
mengqiy, line 21: We can support match GVK or GVKN or GVKNN. We choose to match GVKN here while our golang SDK choose to match GVK.
@yuwenma What's your opinions? IIRC you implemented matching GVK in golang SDK.

mengqiy (commented):
Have you compared this vs writing the SDK in starlark script? What are the pros and cons of the current approach?

functions/go/starlark/krmfn/krmfn.go
yuwenma: GVK is type and NN is meta. the GVKNN  or GVKN is a short form mainly used in kustomize, my main question is how easy it is for users to understand the N or NN. @natasha41575 Do you know how popular this GVKNN convention among kustomize users? If not, separate match_gvk and match_namespace might be easier to use.

natasha41575: I'll change this to have separate match_gvk, match_name, and match_namespace so that it is easier to use.

tests/starlark/load-library/fn-config.yaml
natasha41575: added!

functions/go/starlark/krmfn/krmfn.go
natasha41575: done

functions/go/starlark/krmfn/krmfn.go
natasha41575: done!

natasha41575:
@mengqiy @yuwenma @droot CI here is failing due to the replace directive in kpt. Is it possible to remove the replace directive there?

yuwenma (approved):
lgtm. thank you for this change

mengqiy:
I'm going to release starlark fn today or tomorrow. IMO it makes more sense to merge this after the release. Otherwise, the starlark SDK will be public. I guess our starlark SDK is still in an early stage.

natasha41575:
> Have you compared this vs writing the SDK in starlark script? What are the pros and cons of the current approach?

One benefit of this approach is that we will be able to easily provide all of the same functions in the starlark SDK and the go SDK if both are written in Go. In this PR, we wrapped the `KubeObject.IsGVK` function to provide it for Starlark. This is a simple example, but we can do the same for more complex ones. For example `KubeObject.SetNestedString` is somewhat complex. If we want to provide the same function in Starlark, we can either rewrite all of its internal logic in Starlark from scratch, or we can just import it to this Go package here. IMO this approach seems more sustainable in the long term. WYDT @mengqiy?

mengqiy:
@natasha41575 What you said makes sense. Thanks!

PR #796, opened 2022-04-04
Handle kpt fn behavior change

droot:
`kpt` has introduced a breaking change recently where `kpt fn` commands now include functionConfigs and meta resources such as `Kptfile` in the function input. This PR adjusts tests/examples and `ensure-name-substring` function to match the new `kpt fn` behavior.


mengqiy:
We can unblock this PR by updating:
- https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/9501ec9fc4fbced92003034aa1d44332bbc2a149/functions/go/Makefile#L116
- https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/9501ec9fc4fbced92003034aa1d44332bbc2a149/contrib/functions/go/Makefile#L96

mengqiy:
We need to change [this line](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/9501ec9fc4fbced92003034aa1d44332bbc2a149/.github/workflows/ci.yaml#L101) back to a version that contains your breaking change in kpt but not https://github.com/GoogleContainerTools/kpt/pull/2961, so that we can verify the behavior against the desired kpt CLI version.

mengqiy (approved):

mengqiy:
Thanks for fixing it!

yuwenma:
Sweet! Thank you so much for the quick fix!

PR #805, opened 2022-04-07
use kube-openapi that is compatible with k8s v0.23.5

mengqiy:
Reverted the dep change in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/785.
This PR is similar as https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/787.


functions/go/set-namespace/go.mod
```diff
@@ -27,10 +25,9 @@ require (
 	github.com/xlab/treeprint v1.1.0 // indirect
 	golang.org/x/net v0.0.0-20220403103023-749bd193bc2b // indirect
 	golang.org/x/text v0.3.7 // indirect
-	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 	k8s.io/klog/v2 v2.60.1 // indirect
-	k8s.io/kube-openapi v0.0.0-20220401212409-b28bf2818661 // indirect
+	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
```
mengqiy, line 31: Changes in the indirect section are easier to miss when we review PRs.
Even if I tried to pin it in the 1st `require` section, running `go mod tidy` will move it back :(

yuwenma (approved):

PR #808, opened 2022-04-07
feat: set-namespace special treatment for RoleBinding

yuwenma:
RoleBinding is namespace scoped, but it can bind other namespace's ServiceAccount to the namespace it lives. When replacing a namespace, if the rolebinding's own namespace matches the target namespace, itself should be updated, if its subjects namespace matches the target namespace (even if itself is in another namespace), the subject's namespace should be updated.

The code change merges the ClusterRoleBinding and Rolebinding's `subjects` handling together, and include `RoleBinding` itself to the `metadata.namespace` updates.


droot (approved):
Would be good to add test coverage for this.

yuwenma:
> Would be good to add test coverage for this.

examples and tests are coming!

PR #816, opened 2022-04-09
Optimize openapi parsing in starlark

mengqiy:
Added a wrapper to enable lazy initialization for openapi schema.
The schema is parsed only when the script tries to read it.

When the starlark script doesn't need to read openapi, it saves 0.7s.

### DIFF
--- a/functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/context.go
+++ b/functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/context.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/qri-io/starlib/util"
 	"go.starlark.net/starlark"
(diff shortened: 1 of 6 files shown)


mengqiy:
cc: @martinmaly @natasha41575

functions/go/starlark/third_party/sigs.k8s.io/kustomize/kyaml/fn/runtime/starlark/context.go
```diff
@@ -60,6 +58,56 @@ func env() (starlark.Value, error) {
...
+	once sync.Once
+	val  starlark.Value
+}
+
+var _ starlark.Mapping = &LazyInitializationOpenapi{}
+
+func (v *LazyInitializationOpenapi) init() {
+	o, err := oa()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "panic!!!!!! %v\n", err)
```
natasha41575, line 71: Do we need this?

mengqiy: Good catch!
Removed!

mengqiy:
PTAL

yuwenma:
We shall move the third-party to kpt-functions-sdk, but I know it has more overheads to contribute on both function and its dependent libs. @mengqiy Is there any more work need to be done for starlark function? If not, we can start the migration.

yuwenma (approved):

mengqiy:
> We shall move the third-party to kpt-functions-sdk, but I know it has more overheads to contribute on both function and its dependent libs. @mengqiy Is there any more work need to be done for starlark function? If not, we can start the migration.

I agree we should move it to the SDK repo.
There are no other work planned in the starlark function

PR #817, opened 2022-04-12
fix: check whether previous namespace field is set

yuwenma:
### Improve the user message when set-namespace passes

- If update the namespace:
```bash
[RUNNING] "./set-ns"
[PASS] "./set-ns" in 0s
  Results:
    [info]: namespace "yu" updated to "yu2", 3 values changed
```
- If no need to update namespace:
```bash
 kpt fn eval --exec ./set-ns yu --truncate-output=false -- namespace=yu
[RUNNING] "./set-ns"
[PASS] "./set-ns" in 700ms
  Results:
    [info]: namespace "yu" updated to "yu", 0 values changed
```

### Bug fix

Catch the risk condition when the input resources do not contain namespace field. This causes a panic, needs to import the new SDK once PR merges https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/557
The user message after the fix is: 
```bash 
 kpt fn eval --exec ./set-ns yu --truncate-output=false -- namespace=yu
[RUNNING] "./set-ns"
[PASS] "./set-ns" in 600ms
  Results:
    [error]: could not find any namespace fields to update. This function requires at least one of Namespace objects or namespace-scoped resources to have their namespace field set.
```

* Add an example set-namespace-ns-unset


examples/set-namespace-ns-unset/.expected/config.yaml
```diff
@@ -0,0 +1 @@
+exitCode: 0
```
mengqiy, line 1: This is the default value. So you can remove this line and then remove this file. It should not affect the tests.

examples/set-namespace-ns-unset/README.md
```diff
@@ -0,0 +1,47 @@
...
+Invoke the function by running the following commands:
+
+```shell
+$ kpt fn render set-namespace-ns-unset
+```
+
+### Expected result
+
+Users should receive error saying:
+  "could not find any namespace fields to update. This function requires at least one of Namespace objects or "
```
mengqiy, line 44: This behavior should be documented in the usage doc.

examples/set-namespace-ns-unset/README.md
```diff
@@ -0,0 +1,47 @@
+# set-namespace: ns Unset Example
```
mengqiy, line 1: It seems this is more testing an edge case than a useful example to users. If that's the case, we can move it to `tests/namespace/<test-name>`.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -31,18 +31,23 @@ func SetNamespace(rl *fn.ResourceList) (bool, error) {
 		return true, nil
 	}
 	// Update "namespace" to the proper resources.
-	tc.Transform(rl.Items)
+	count := tc.Transform(rl.Items)
```
mengqiy, line 34: Note that some of our functions in the catalog tracks not only `count` but also each field that is updated. e.g. https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/set-image/v0.1/examples/set-image-simple/.expected/results.yaml

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -145,13 +152,18 @@ func FindAllNamespaces(objects []*fn.KubeObject) ([]string, map[string]int) {
 }
 
 // ReplaceNamespace iterates the `objects` to replace the `OldNs` with `newNs` on namespace field.
-func ReplaceNamespace(objects []*fn.KubeObject, oldNs, newNs string) {
+func ReplaceNamespace(objects []*fn.KubeObject, oldNs, newNs string, count *int) {
+	*count = 0
```
mengqiy, line 156: Is this line necessary when you have `count := new(int)` before passing it in?
Why don't we simply return this value instead of passing in an int pointer?

yuwenma: good catch! 

"Why don't we simply return this value instead of passing in an int pointer?" --> because `replaceNamespace` should not expect a return with an integer telling how many fields are updated. The function name should expect an error or silence.

functions/go/set-namespace/transformer/namespace.go
yuwenma: Good point. The `count` comes from the fishfooding issue b/228617774. I will add the function specification result as well.

examples/set-namespace-ns-unset/README.md
yuwenma: yeah, and the 3 examples for update mode... I have another draft PR (but waiting for the decision about the namespace default behavior)

examples/set-namespace-ns-unset/.expected/config.yaml
yuwenma: thanks!

functions/go/set-namespace/transformer/namespace.go
yuwenma: Actually, adding the result info in resource level might end up output too much details. created https://github.com/GoogleContainerTools/kpt/issues/2996 to track further discussions. I would prefer keeping the output short as an update summary.

mengqiy: There is no knob in the function spec that controls the output of a function.
IMO a function should provide as much information as possible, but the function orchestrator (kpt CLI in this case) should decide how verbose the output should be. (e.g. hide `info` and `warning` at different verbosity level).

It's not a hard requirement for this PR, but a nice-to-have feature.

mengqiy (approved):

functions/go/set-namespace/transformer/namespace.go
yuwenma: Right, so I propose adding a new kpt flag to control the output severity level, for both `kpt fn eval` and `kpt fn render`. https://github.com/GoogleContainerTools/kpt/issues/2996
But one thing I don't like is that `kpt fn` has already had too many flags. kustomize sets a preference to configure the result in kustomization.yaml. But I don't think that's applied to the severity either. Do you have any suggestions how we can configure that?

yuwenma:
updated set-namespace with the latest SDK which expose the panic with stack trace.

PR #818, opened 2022-04-13
Cherry-pick abea229 from master to set-namespace/v0.3

yuwenma:


mengqiy:
@sdowell It seems there is an inconsistency between what the verify-docs expects and what we really want: https://github.com/GoogleContainerTools/kpt-functions-catalog/runs/6014038143?check_suite_focus=true

mengqiy:
I'm going to merge this PR. We don't have to block this PR.

PR #819, opened 2022-04-13
docs: Update tags for go/set-namespace/v0.3.3 (#803)

yuwenma:


sdowell:
The target should be a release branch (in this case `set-namespace/v0.3`)

mengqiy:
The target branch should be the release branch instead of master

yuwenma:
> The target should be a release branch (in this case `set-namespace/v0.3`)

Ah, I see. That helps. Actually in my local I used `upstream/set-namespace/v0.3` since my origin and upstream have the same branch it causes the following error.  Would be great to add some user guide.  
```bash
hint: If you'd like to always have checkouts of an ambiguous <name> prefer
hint: one remote, e.g. the 'origin' remote, consider setting
hint: checkout.defaultRemote=origin in your config.
fatal: 'set-namespace/v0.3' matched multiple (2) remote tracking branches
```

PR #827, opened 2022-04-16
Cleanup `set-namespace` output message

yuwenma:
Current set-namespace gives redundant message as below
```bash
  Results:
    [info]: namespace "example" updated to "example", 0 values changed
    [info] : set image from example to example
```
The second line is per KubeObject level info and should be removed.

### DIFF
--- a/functions/go/set-namespace/transformer/namespace.go
+++ b/functions/go/set-namespace/transformer/namespace.go
@@ -41,16 +41,6 @@ func SetNamespace(rl *fn.ResourceList) (bool, error) {
 	}
 
 	rl.Results = append(rl.Results, result)
-	rl.Results = append(rl.Results, &fn.Result{
-		Message: fmt.Sprintf("set image from %s to %s", tc.NamespaceMatcher, tc.NewNamespace),
-		Field: &fn.Field{
-			Path:          "",
-			CurrentValue:  tc.NamespaceMatcher,
-			ProposedValue: tc.NewNamespace,
-		},
-		File:     &fn.File{Path: "test", Index: 0},
-		Severity: fn.Info,
-	})
 	return true, nil
 }
 
--- a/tests/set-namespace/ns-unset/.expected/results.yaml
+++ b/tests/set-namespace/ns-unset/.expected/results.yaml
@@ -9,10 +9,3 @@ items:
     results:
       - message: could not find any namespace fields to update. This function requires at least one of Namespace objects or namespace-scoped resources to have their namespace field set.
         severity: error
-      - message: set image from  to example-ns
-        severity: info
-        field:
-          currentValue: ""
-          proposedValue: example-ns
-        file:
-          path: test


mengqiy:
@yuwenma Can you give before vs after? It seems you only give before now.

yuwenma:
> @yuwenma Can you give before vs after? It seems you only give before now.

The output after this change is 
```bash
kpt fn eval --exec set-ns --fn-config package-context.yaml
[RUNNING] "set-ns"
[PASS] "set-ns" in 800ms
  Results:
    [info]: namespace "user-service" updated to "user-service", 0 values changed
```

mengqiy:
@yuwenma Can you please also paste before vs after for the structured result?

yuwenma:
> @yuwenma Can you please also paste before vs after for the structured result?

We have a test covering the [results diff](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/827/files#diff-6725431b164e68d9566780ef0d1b456eabc4493f3aec464c6585e89c5a997eff). Do you mean something else?

mengqiy:
> We have a test covering the [results diff](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/827/files#diff-6725431b164e68d9566780ef0d1b456eabc4493f3aec464c6585e89c5a997eff). Do you mean something else?

That's why I was asking. It seems the structured results no longer has that information, is that desired?

yuwenma:
> > We have a test covering the [results diff](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/827/files#diff-6725431b164e68d9566780ef0d1b456eabc4493f3aec464c6585e89c5a997eff). Do you mean something else?
> 
> That's why I was asking. It seems the structured results no longer has that information, is that desired?

Yes, That's expected. The removed part is a per-KubeObject result (and not logged correctly). In [our previous discussion](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/817#discussion_r848972538), I no longer log the per-kubeobject result due to the redundant message it will introduce. 

qq: Do you know why the function specification only has info, warning and error level Severity? but no debug and fatal?

mengqiy:
> qq: Do you know why the function specification only has info, warning and error level Severity? but no debug and fatal?

IMO maybe it will help to add `debug` severity, but I can imagine how `fatal` severity can help in a KRM function. Do you have a use case in your mind?

mengqiy (approved):

sdowell:
It seems the proposed UX for this function differs a bit from the UX for similar functions such as `set-labels` or `set-annotations`. It may be worth considering to provide a similar UX for the alike functions

yuwenma:
> It seems the proposed UX for this function differs a bit from the UX for similar functions such as `set-labels` or `set-annotations`. It may be worth considering to provide a similar UX for the alike functions

@sdowell Yes, we are analyzing and applying the new sdk to these functions. set-namespace, apply-replacement are the first set of functions.

PR #832, opened 2022-04-18
docs: Update tags for go/set-namespace/v0.3.4

yuwenma:
fix version drift


yuwenma:
going to accept https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/833/files which uses the right base to run `make update-function-docs`

PR #844, opened 2022-05-03
set-namespace: Simplify the logic

yuwenma:
### Description
This change allows `set-namespace` to be used with more flexibility and users are no longer required to be aware of the original namespace value they plan to update.

### Changes
- Remove the validation for original namespaces. 
- Update all namespace values to the new value whatever its original value is. 
- Remove the 3 update Mode since all namespaces are squashed to the new one.
- Remove the `namespaceMatcher` feature, users won't be able to only update a specific namespace value to a new value. 
- Add post transformation validation check to detect duplicate GVKNN resources, which can be a common resource error when squashing all namespaces into a new one.
- Delete 3 update Mode examples
- Add e2e tests to cover new behavior
- Added two WYSIWYG e2e tests to cover the scenarios shown in the WYSIWYG Configuration as Data demo.
   - [wysiwyg-add-blueprint-resource](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/844/files#diff-66e6bd5d16fb0b34410b6b5ae454d805241e3a916352f1715c1630cbf15f23c1)
   - [wysiwyg-authroing-blueprint-from-basens](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/844/files#diff-292ce1889e275c44fdd9dc5e467a61134760c6e4208ae37ce016c01d3dc34d1e)


yuwenma:
Added two WYSIWYG e2e tests to cover the scenarios shown in the [WYSIWYG Configuration as Data demo](https://www.youtube.com/watch?v=CNGp5niim9c).
[wysiwyg-add-blueprint-resource](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/844/files#diff-66e6bd5d16fb0b34410b6b5ae454d805241e3a916352f1715c1630cbf15f23c1)
[wysiwyg-authroing-blueprint-from-basens](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/844/files#diff-292ce1889e275c44fdd9dc5e467a61134760c6e4208ae37ce016c01d3dc34d1e)

bgrant0607:
Thanks. I think this should be less error-prone for now. We can discuss what the long-term behavior should be in the future. kustomize name transforms record the original names. We will need to do similarly.

I don't have time to do a thorough review, but I did look at the test inputs / outputs, and this looks like it will do what I need.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -31,46 +30,51 @@ func SetNamespace(rl *fn.ResourceList) (bool, error) {
...
+	}
+	if dupErr := fn.CheckResourceDuplication(rl); dupErr != nil {
+		return fn.ErrorResult(dupErr)
+	}
+	if count == 0 {
+		msg := fmt.Sprintf("all namespaces are already %q. no value changed", p.NewNamespace)
+		return fn.GeneralResult(msg, fn.Info)
+	}
+
+	msg := fmt.Sprintf("namespace %q updated to %q, %d value(s) changed", reflect.ValueOf(oldNamespaces).MapKeys(),
```
mengqiy, line 53: > reflect.ValueOf(oldNamespaces).MapKeys()

IMO it's not necessary to use `reflect` here.
We can simply convert the map to a slice.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -100,68 +101,47 @@ func VisitNamespaces(objects []*fn.KubeObject, visitor func(namespace *Namespace
...
-			// skip the cluster scoped resource. We determine if a resource is cluster scoped by
-			// checking if the metadata.namespace is configured.
 		}
-		if o.HasNamespace() {
-			// We made the hypothesis that a namespace scoped resource should have its metadata.namespace field setup.
-			// All above special types are cluster scoped, except RoleBinding
-			namespace := o.GetStringOrDie("metadata", "namespace")
+		// "IsNamespaceScoped" detects resource scope by checking whether `metadata.namespace` is set.
+		// This is valid hypothesis is for custom k8s resources. For k8s resources, the function always gives accurate scope information.
+		if o.IsNamespaceScoped() {
```
mengqiy, line 132: Looking at the behavior, IMO `HasNamespace` is a better name than `IsNamespaceScoped`.

mengqiy (commented):
A few minor comments. LGTM otherwise.
We have breaking changes. Next release for set-namespace will be `v0.4.0`.

functions/go/set-namespace/transformer/namespace.go
yuwenma: HasNamespace and IsNamespaceScoped are actually different.
`HasNamespace` only checks whether the `metadata.namespace` is set, this is our hypothesis to tell whether a custom KRM resources is namespace scoped or not.

`IsNamespaceScoped` is the name from "func IsXXX() bool" and k8s CRD "scope: namespaced", which expects to tell more accurate namespace info, which is aligned with what the SDK does:  
The SDK now has a better knowledge for core k8s resource like `Deployment`, `Service`  and `RoleBinding`, and does not need to check the `metadata.namespace`  https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/568.

functions/go/set-namespace/transformer/namespace.go
mengqiy: It will be something like:
```
var oldNsList []string
for oldNs := range oldNamespaces {
	oldNsList = append(oldNsList, oldNs)
}
fmt.Sprintf("message...", strings.Join(oldNsList), foo, bar)
```

It has a few more lines.But IMO it's easier to reason about than using `reflect`.

functions/go/set-namespace/transformer/namespace.go
mengqiy: I misunderstood something :P

mengqiy (approved):

functions/go/set-namespace/transformer/namespace.go
yuwenma: done.

PR #845, opened 2022-05-04
Migrate set-image to use new sdk

yuwenma:
Move to use new SDK in `kpt-functions-catalog/go/fn`. This should fix the issue which errors end up as function "PASS" in `kpt fn eval` and `kpt fn render`:

- user errors will show as function "FAIL" with user guidance in the result. e.g.
```bash
kpt fn eval --exec ./set-image --fn-config fn-config.yaml 
[RUNNING] "./set-image"
[FAIL] "./set-image" in 200ms
  Results:
    [error] v1/ConfigMap/my-func-config: missing image newName, newTag, or digest
```  
- fn config error (parsing, yaml, resource errors) will show as function "FAIL". e.g.
```bash
 kpt fn eval --exec ./set-image --fn-config fn-config.yaml       
[RUNNING] "./set-image"
[FAIL] "./set-image" in 0s
  Results:
    [error]: Resource(apiVersion=fn.kpt.dev/v1alpha1, kind=SetImage, Name=my-func-config) has unmatched field type: `
  Exit code: 1
```

TODO: SDK should better  support `As` error handling, the second example does not help users to fix the issue.


functions/go/set-image/image_transformer_test.go
```diff
@@ -344,9 +352,9 @@ items:
       name: test-container2
 `,
 			ExpectedResults: `
-[info] v1/Pod/the-namespace/the-pod spec.containers.image: set image from foo:1.2.3 to bar:4.5.6
```
mengqiy, line 347: We return less information now. Why?

functions/go/set-image/image_transformer.go
```diff
@@ -5,7 +5,7 @@ import (
 	"strconv"
 	"strings"
 
-	sdk "github.com/GoogleContainerTools/kpt-functions-catalog/thirdparty/kyaml/fnsdk"
```
mengqiy, line 8: Thanks!

functions/go/set-image/image_transformer.go
```diff
@@ -114,62 +109,72 @@ func (si *SetImage) validateInput() error {
...
-		}
-	case validGVK(functionConfig, fnConfigAPIVersion, fnConfigKind):
-		if err := functionConfig.As(si); err != nil {
-			return fmt.Errorf("unable to convert functionConfig to %s %s:\n%w",
-				fnConfigAPIVersion, fnConfigKind, err)
-		}
+	case functionConfig.IsGVK( "v1", "ConfigMap"):
+		functionConfig.GetOrDie(&si.Image, "data")
+	case functionConfig.IsGVK(fnConfigAPIVersion, fnConfigKind):
+		functionConfig.AsOrDie(si)
```
sdowell, line 119: What is the convention for when to panic vs return an error?

functions/go/set-image/image_transformer.go
```diff
@@ -114,62 +109,72 @@ func (si *SetImage) validateInput() error {
...
 	if err != nil {
-		return err
+		return err, false
 	}
 	si.AdditionalImageFields = append(si.AdditionalImageFields, defaultImageFields...)
-	return nil
+	return nil, true
 }
 
 // Transform set image out of place and returns a new []*sdk.KubeObject
```
sdowell, line 134: It seems this comment is not accurate now since the input is modified in place

functions/go/set-image/image_transformer_test.go
yuwenma: We should have `ParseResourceList` parse `results` from the input []byte, right now it only parses `items` and `functionConfig`. So I have to use KubeObject and pull the result message one by one. See `runImageTransformerResults`

Once the `ParseResourceList` attaches `results` to `ResourceList`, we can have the unittests include more complete info. 

qq: I am in the impression we are more leveraging the e2e test but not unittests to cover the functions. Or is that only for kpt CLI?

functions/go/set-image/image_transformer.go
yuwenma: SDK processor handles specific panics like mismatch field type (what `XXXOrDie` raises). This is to give user a better experience. 
Back to your question about when to panic vs return an error:
If a SDK function returns an error, we expect users to handle the error; if a SDK function does not return an error, users should *not* expect panic **raised directly from the SDK side**.

yuwenma: The second use case in the description is from the `AsOrDie` panic
```bash
[error]: Resource(apiVersion=fn.kpt.dev/v1alpha1, kind=SetImage, Name=my-func-config) has unmatched field type: 
  Exit code: 1
```

functions/go/set-image/image_transformer.go
yuwenma: ah, good catch! Thanks

functions/go/set-image/image_transformer_test.go
```diff
@@ -19,20 +19,31 @@ func itemsToYaml(items []*sdk.KubeObject) string {
...
-	functionConfig, err := sdk.ParseKubeObject([]byte(config))
+	functionConfig, err := fn.ParseKubeObject([]byte(config))
 	if err != nil {
 		return nil, err
 	}
 	rl.FunctionConfig = functionConfig
-
-	if err = setImageTags(rl); err != nil {
+	in, _ := rl.ToYAML()
+	out, err := fn.Run(fn.ResourceListProcessorFunc(setImageTags), in)
```
sdowell, line 33: What is the motivation for changing the main entry point in this test scaffolding?

yuwenma: Calling `SetImageTags` directly won't go through the SDK panic handling process. In fact, when using sdk methods, we expects users to either use `fn.AsMain` or `fn.Execute` which they will pass in the `fn.ResourceListProcessorFunc(setImageTags)` and not calling the method `setImageTags` directly

sdowell (approved):

PR #850, opened 2022-05-05
Empty functionconfig check

yuwenma:
Rebase to resolve panic on empty functionConfig, change fnConfig failure return status from "PASS" to "FAIL" 
All three porch builtin functions are updated.

The output is 
``` 
kpt fn render --allow-exec --truncate-output=false
Package "empty-fn": 
[RUNNING] "./set-ns"
[FAIL] "./set-ns" in 800ms
  Results:
    [error]: FunctionConfig is missing. Expect `ConfigMap` or `SetNamespace`
  Exit code: 1
```

Even without the dummy check code, the fnConfig won't panic with fix  [#570](https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/570), it will fall into the default error which is
```
Package "empty-fn": 
[RUNNING] "./set-ns"
[FAIL] "./set-ns" in 800ms
  Results:
    [error]: unknown functionConfig Kind= ApiVersion=, expect `SetNamespace` or `ConfigMap`
  Exit code: 1
```


functions/go/starlark/starlark/config.go
```diff
@@ -34,6 +34,8 @@ type StarlarkRun struct {
 
 func (sr *StarlarkRun) Config(fnCfg *fn.KubeObject) error {
 	switch {
+	case fnCfg.IsEmpty():
+		return fmt.Errorf("FunctionConfig is missing. Expect `ConfigMap` or `StarlarkRun`")
```
yuwenma, line 38: @mengqiy  This is actually not needed for the new SDK. if not check, this will fall into the default case
`functionConfig` must be either ...`

yuwenma: Or we can always check the IsEmpty. wdyt?

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -25,8 +25,8 @@ func SetNamespace(rl *fn.ResourceList) (bool, error) {
 	// Get "namespace" arguments from FunctionConfig
 	err := tc.Config(rl.FunctionConfig)
 	if err != nil {
-		rl.Results = append(rl.Results, fn.ErrorConfigObjectResult(err, rl.FunctionConfig))
-		return true, nil
+		rl.Results = append(rl.Results, fn.ErrorResult(err))
+		return false, nil
```
mengqiy, line 29: We should document the `ResourceListProcessor` interface in the SDK repo. Otherwise, users may not know what does the bool mean here.
https://pkg.go.dev/github.com/GoogleContainerTools/kpt-functions-sdk/go/fn#ResourceListProcessor

functions/go/starlark/starlark/config.go
mengqiy: It seems this also applies to the other 2 functions, right?

yuwenma: If not check, the falls into the default with error message 
"functionConfig` must be either ConfigMap or StarlarkRun, but we got: " . which gives users the same kind of suggestion as  `IsEmpty`

yuwenma: Basically the SDK fixes the nil pointer references when calling a KRM methods. This `IsEmpty` is merely for better user messaging. Not to handle risk conditions.

mengqiy (approved):

mengqiy:
@yuwenma It seems CI is faling

functions/go/set-namespace/transformer/namespace.go
yuwenma: SG! I'm adding examples and tests on the sdk. (though that PR keeps on postponed by May blocker issues).

PR #860, opened 2022-05-07
fix: Update RELEASE.md

yuwenma:
We cannot hard reset the git references since it will confuse the local git cache and causes git ref conflict.
Here's the error I get https://linuxpip.org/git-error-cannot-lock-ref. I think running the `RELEASE_BRANCH=${RELEASE_BRANCH} make update-function-docs` the first time works, but after that, the git ref becomes dirty and hard to fix. 

Propose a safer way to run the doc-update script. I tried this new pattern to release `set-namespace:v0.4.1`, `apply-replacements:v0.1.1` and `starlark:v0.4.3`.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -23,28 +23,38 @@ be created on the [releases pages]. `RELEASE_BRANCH` is in the form of
 `${FUNCTION_NAME}/v${MAJOR_VERSION}.${MINOR_VERSION}`.
 For example `set-namespace/v0.3`, `kubeval/v0.1`, etc.
 
-
-1. Run the script:
+1. Setup the release branch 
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -23,28 +23,38 @@ be created on the [releases pages]. `RELEASE_BRANCH` is in the form of
...
+5. Send out a Pull Request. 
+	Your local git reference is now pointing to the local RELEASE BRANCH.
+	A new git commit is auto-generated which contains the function document referring to the latest function version in the form of 
+	`<FUNCTION_NAME>/v<MAJOR>.<MINOR>.<PATCH>`
+	You should be ready to submit the Pull Request against the upstream <RELEASE_BRANCH>. 
 ```shell
-# Push the commit to your remote (branch name can be anything)
-# e.g. git push origin HEAD:update-docs-set-namespace
-git push origin HEAD:<remote-branch>
+> git push -f origin ${RELEASE_BRANCH}
```
sdowell, line 56: A minor nit: I would probably direct the user to create a new branch to represent the change rather than using release branch ref

sdowell (approved):

PR #861, opened 2022-05-11
Use node.js 14 in CI

mengqiy:


.github/workflows/ci.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
-        node-version: [12.x]
```
sdowell, line 40: Do any of the docker images need to be updated? Upon a quick glance it looks like the [common ts dockerfile](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/build/docker/ts/Dockerfile#L1) uses `node:lts-alpine`, which actually points to node.js 16 currently. We might want to pin that to a fully qualified tag. As far as I can tell the other TS dockerfiles are pinned to `node:14.17-alpine3.13`, so I think those are fine

sdowell (approved):

.github/workflows/ci.yaml
mengqiy: Updated. PTAL

sdowell (approved):
LGTM.

If the intent is to keep the base image for each of these in lockstep, then I think as a separate improvement we could define the base image in a single place and pass it into each of these dockerfiles as an argument. That would make this a bit easier to manage IMO

mengqiy:
I'd like to see it happen if you already have an idea how to do it.

PR #866, opened 2022-05-17
test kubeval with kptfile

mengqiy:


sdowell:
Could you provide some context for this change? It's a simple change so I'm inclined to approve but I'm not sure I understand the scope of the change

droot (approved):

droot:
> Could you provide some context for this change? It's a simple change so I'm inclined to approve but I'm not sure I understand the scope of the change

Current tests were configured to skip `Kptfile` kind because the openapi schema packaged in the `kubeval` function didn't include schema for `Kptfile` and used to result in failure. Recently @mengqiy added the schema for `Kptfile`, so this change removes the special treatment (skipping) for `Kptfile` kind.

sdowell (approved):

PR #870, opened 2022-05-20
support building multi-arch image

mengqiy:
In the `build` target, we only build the matching one. This is because `docker buildx` doesn't support exporting an multi-arch image to local. And building multi-arch image locally is not useful for users and it's more time consuming.
In the `push` target, we build for both amd64 and arm64.

fixes: https://github.com/GoogleContainerTools/kpt/issues/2874
fixes: https://github.com/GoogleContainerTools/kpt/issues/2775

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,9 +7,9 @@ on:
   pull_request:
     branches:
       - master
-      - '*/v[0-9]+.[0-9]+'
+      - "*/v[0-9]+.[0-9]+"
   schedule:
-    - cron:  '0 17 * * *'
+    - cron: "0 17 * * *"
 
 concurrency:
   # If a job with a concurrency group is running with cancel-in-progress, that job is cancelled when a new job with same concurrency group is started.
@@ -17,7 +17,7 @@ concurrency:
   # Concurrency group will be of form ${{ github.head_ref }}-ci when triggered via PR creating 1 group per PR. Older jobs are cancelled when new commits are pushed to that PR branch.
   # Concurrency group will be of form ${{ github.ref }}-ci when triggered from repo branch or tag ref. Older jobs are cancelled when new jobs are triggered from that same branch/tag.
   # `-ci` suffix is to namespace the concurrency group incase you want to add a group for another workflow in the future.
-  group: '${{ github.head_ref || github.ref }}-ci'
+  group: "${{ github.head_ref || github.ref }}-ci"
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -40,44 +40,44 @@ jobs:
...
+      - name: Run unit tests
+        run: |
+          make unit-test-go
+      - name: Check licenses
+        run: |
+          cd functions/go
+          make check-licenses
 
   e2e-ci:
-    needs: [ts-unit-test-ci, go-unit-test-ci]
```
mengqiy, line 80: It's not necessary to block the e2e tests.

scripts/docker.sh
```diff
@@ -58,26 +59,28 @@ function docker_build {
...
+        -f "${dockerfile}" \
+        "${build_args[@]+"${build_args[@]}"}" \
+        "${function_dir}"    
+      ;;
+    push)
+      # build and push multi-arch image.
+      docker buildx build --push \
+        -t "${GCR_REGISTRY}/${name}:${tag}" \
+        -f "${dockerfile}" \
+        --platform "linux/amd64,linux/arm64" \
```
mengqiy, line 78: Build multi-arch images when pushing to a registry.

mengqiy:
@droot @sdowell PTAL

.github/workflows/ci.yaml
droot: Does this mean, this job will start in parallel with the `ts-unit-test-ci` and `go-unit-test-ci` ?

scripts/docker.sh
droot: Will it work for `OSX` or just `linux` ?

scripts/docker.sh
```diff
@@ -58,26 +59,28 @@ function docker_build {
...
+        "${build_args[@]+"${build_args[@]}"}" \
+        "${function_dir}"    
+      ;;
+    *)
+      echo "action must be load or push"
+      exit 1
+  esac
```
droot, line 85: I don't see any updates related to documentation for building/publishing functions ?

droot (approved):
Looks good to me. I have a few questions.

.github/workflows/ci.yaml
mengqiy: Yes. Do you have any concerns?

scripts/docker.sh
mengqiy: It will work on mac.
`container` is a linux-only thing. On mac, containers are executed in a linux VM.

ARM has v7 and v8. From my understanding, Apple's chips are all v8. The default is v8 when we are using linux/arm64.
We can add v7 in the future if needed.

.github/workflows/ci.yaml
droot: No concerns. Just happy :)

scripts/docker.sh
droot: Cool.

scripts/docker.sh
mengqiy: I added something in `CONTRIBUTING.md`. I mentioned that `docker buildx` should be installed. Everything else is the same.

mikebz:
exciting progress here!

mikebz:
Quick follow up question at what junction will the current functions have both images, are we re-running the build to produce multi architecture stuff?

PR #873, opened 2022-05-31
feat:adds export for Log Bucket

g-awmalik:


g-awmalik:
@bharathkkb - can you sanity check this please?

functions/go/export-terraform/terraformgenerator/templates/log-export.tf
```diff
@@ -45,4 +45,14 @@ module "{{ .GetResourceName }}-destination" {
     is_locked             = {{ .GetBool "spec" "retentionPolicy" "isLocked" }}
   }{{end}}
 }
+{{end}}{{with $logsink.References.LoggingLogBucket }}
+module "{{ .GetResourceName }}-destination" {
+  source  = "terraform-google-modules/log-export/google//modules/logbucket"
+  version = "~> 7.3.0"
```
bharathkkb, line 51: Will only land in the next release
```suggestion
  version = "~> 7.4.0"
```

functions/go/export-terraform/terraformgenerator/templates/log-export.tf
```diff
@@ -45,4 +45,14 @@ module "{{ .GetResourceName }}-destination" {
...
+
+  project_id                  = module.{{ .Parent.GetResourceName }}.project_id
+  log_bucket_name             = "{{ .GetResourceName }}"
```
bharathkkb, line 54: Shouldn't this be name?
https://github.com/terraform-google-modules/terraform-google-log-export/blob/eb777cae1f7a7ef7df4f7f9651797b0213b3412f/modules/logbucket/variables.tf#L22

functions/go/export-terraform/terraformgenerator/templates/log-export.tf
```diff
@@ -45,4 +45,14 @@ module "{{ .GetResourceName }}-destination" {
...
+  location                    = "{{.}}"{{end}}{{ with .GetStringFromObject "spec" "location" }}
+  retention_days              = "{{.}}"{{end}}{{ if .GetInt "spec" "retentionDays"}}
+}
```
bharathkkb, line 57: We should also expose https://github.com/GoogleCloudPlatform/blueprints/blob/aa4a37d846c816de4649fa555c327f9ede89cb26/catalog/log-export/folder/logbucket-export/export.yaml#L58 in TF for parity above with the storage

bharathkkb (commented):
@g-awmalik some comments below and lets also add a test to validate expected TF.

functions/go/export-terraform/terraformgenerator/templates/log-export.tf
g-awmalik: ack

functions/go/export-terraform/terraformgenerator/templates/log-export.tf
g-awmalik: Yes, you're right.

functions/go/export-terraform/terraformgenerator/templates/log-export.tf
g-awmalik: I wanted to but it's not exposed in the [TF resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/logging_project_bucket_config) for some reason. I'll open up an issue in the provider repo but for now we'll have to let this go.

bharathkkb: @g-awmalik sgtm

g-awmalik: https://github.com/hashicorp/terraform-provider-google/issues/11798

g-awmalik:
@mengqiy - the upsert-resource-simple example is erroring out with `kpt@main`. I've tried running it with the latest version of kpt i.e. 1.0.0-beta.16 is goes through fine. 

Seems like there's a feature (or bug) that doesn't allow two resources with the same GVK loaded in the list which is what the upsert example is trying to do.

Is there a reason we're using the `main` version for the CI? Can we switch that to a more stable (latest) version?

mengqiy:
> Is there a reason we're using the main version for the CI? Can we switch that to a more stable (latest) version?

We use the kpt built from HEAD of `main` on purpose, since if a new kpt feature can break a function, we want to be aware of it before our users. Using the latest kpt can help us catch this signal earlier.

g-awmalik:
Thanks for the update.

g-awmalik:
@mengqiy - can we merge this since the tests for this function are all passing? This required for a feature we're working on for Customer Onboarding and we need to start testing soon.

mengqiy:
@g-awmalik The e2e CI has been fixed on HEAD. A rebase on top of HEAD should make the tests passing.

functions/go/export-terraform/terraformgenerator/resources.go
```diff
@@ -206,6 +218,8 @@ func (resource *terraformResource) attachReferences() error {
 		{kind: "ComputeRouter", path: []string{"spec", "routerRef", "name"}},
 		{kind: "ComputeAddress", customRetriever: singleComputeAddressRetriever([]string{"spec", "natIps"})},
 		{kind: "ComputeAddress", customRetriever: singleComputeAddressRetriever([]string{"spec", "reservedPeeringRanges"})},
+		//todo:awmalik@ - remove customerRetriver when this issue is addressed: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/665
```
mengqiy, line 221: nit: We should use `TODO` by convention.

mengqiy (commented):
Please update the usage doc to reflect the change.
PR LGTM otherwise.

functions/go/export-terraform/terraformgenerator/resources.go
g-awmalik: Ack

g-awmalik:
@mengqiy - ping.

mengqiy:
@g-awmalik Has you updated the usage doc (https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/export-terraform/README.md) as I requested in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/873#pullrequestreview-995488436.
It seems `LoggingLogBucket` is something new that's not reflected in the usage doc.

mengqiy (approved):

g-awmalik:
Thanks for the approval @mengqiy. I don't think I can merge though, will need your help with that.

PR #874, opened 2022-06-02
Multi-arch image build

mengqiy:
This PR contains the following changes:
- In the builder image we use the native platform (linux/amd64) to cross compile for the target platform (linux/arm64). And then copy the cross compilation result to the final base image for the target platform (linux/arm64). This approach is way more faster than doing everything with the target platform. The reason is that when we compile something in an arm64 image on amd64 platform, we are using an emulator which has huge performance penalty. Ref reading: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
- Overhaul the image build process for kubeval
  - One dependency (https://github.com/instrumenta/openapi2jsonschema) is lack of maintenance and can't be installed on arm64 platform. So we generate the json schema beforehand and bake them in the image.
  - Kubeval CLI is lack of maintenance, it doesn't provide a build for arm64. We compile it by ourselves.
- There are a few functions that download other binaries e.g. render-helm-chart. They download the correct binary now.

The motivation to speed up the image building process is that it seems the GCP auth token used in our release pipeline won't last multiple hours and it will cause some image push to failing when it expired.


droot (approved):

mengqiy:
The e2e test has been fixed on HEAD.
PR is rebased on top of HEAD.

PR #875, opened 2022-06-02
add oci and private repo support to render-helm-chart

natasha41575:
Add private repo and OCI support to render-helm-chart function. This PR includes doc updates and also changes the function to use the go sdk.


natasha41575:
Part of this PR needs to undo skipping the directories here: https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/877#discussion_r891761548

yuwenma:
Consider the size of this change, I'd suggest switching to use Go SDK, so once this function is suggested to helm users, if they want to contribute, it can be easier to get started on the Go SDK.

natasha41575:
> Consider the size of this change, I'd suggest switching to use Go SDK

Will do in a separate commit as part of this PR.

natasha41575:
@yuwenma I have updated the fn to use the go sdk instead of kyaml. There is a CI failure because for some reason the OCI test isn't able to get the credentials. I will investigate tomorrow.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -211,49 +254,103 @@ func (p *HelmChartInflationGeneratorPlugin) cleanup() {
 	if p.tmpDir != "" {
 		os.RemoveAll(p.tmpDir)
 	}
+	if strings.HasPrefix(p.Repo, "oci://") && p.password != "" {
```
yuwenma, line 257: IIRC, there are possibilities that the OCI repo may not start with `oci://` prefix. I'd suggest using the same OCI library as kpt here since that's more reliable.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -211,49 +254,103 @@ func (p *HelmChartInflationGeneratorPlugin) cleanup() {
...
-func (p *HelmChartInflationGeneratorPlugin) Generate() (rm resmap.ResMap, err error) {
+func (p *HelmChartInflationGeneratorPlugin) Generate() (objects fn.KubeObjects, err error) {
 	defer p.cleanup()
 	if err = p.checkHelmVersion(); err != nil {
 		return nil, err
 	}
 	if _, exists := p.chartExistsLocally(); !exists {
-		if _, err := p.runHelmCommand(p.pullCommand()); err != nil {
+		var pullArgs []string
+		switch strings.HasPrefix(p.Repo, "oci://") {
```
yuwenma, line 274: ditto. We may need more reliable way to determine whether the repo is OCI or not.

yuwenma: For readability, true/false may not be a good signal to use switch/case :)

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -211,49 +254,103 @@ func (p *HelmChartInflationGeneratorPlugin) cleanup() {
...
+	s := strings.Split(string(stdout), "---")
+	for i := range s {
+		if len(s[i]) == 0 {
+			continue
+		}
+		o, err := fn.ParseKubeObject([]byte(s[i]))
+		if err != nil {
+			if strings.Contains(err.Error(), "expected exactly one object, got 0") {
+				// sometimes helm produces some messages in between resources, we can safely
+				// ignore these
```
yuwenma, line 309: ah! good to know

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -291,21 +388,65 @@ func (p *HelmChartInflationGeneratorPlugin) templateCommand() []string {
...
+			}...)
+		}
+		return args, strHash
+	}
+	return nil, ""
+}
+
+func (p *HelmChartInflationGeneratorPlugin) registryLoginArgs() []string {
+	if p.Repo != "" {
+		args := []string{
```
yuwenma, line 414: from the latest discussion with CS team. does this function still need the login run?

natasha41575: It's not needed for CS because they are using it in exec mode. But it will be needed for the general case for passing credentials to the containerized function.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
```diff
@@ -63,9 +62,64 @@ func (p *HelmChartInflationGeneratorPlugin) establishTmpDir() (err error) {
...
+		return err
+	}
+	if !found {
+		return fmt.Errorf("could not find username in Secret")
+	}
+
+	pass, found, err := targetSecret.NestedString("data", "password")
+	if err != nil {
+		return err
+	}
```
yuwenma, line 102: Another format can be using the `As` method.

yuwenma: then you can call the `corev1.Secret` functions directly. It might be easier for other SWE to understand.
FYI, error happens when the type of `data/password` path is not a string. I guess the main reason you use `NestedString` but not `NestedStringOrDie` is for `found`. But how about `pass` value is ""? Shall it give the same user message "secret not given" or acceptable?

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
yuwenma: yeah, I agree with still supporting the authentication. I'm wondering whether the `helm registry login` is needed with the password and user flags

natasha41575: `helm registry login` is still needed, I don't think `helm pull` with OCI works even with user and password flags unless you do `helm registry login` first.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: What kpt library are you referring too? Seems like kpt checks for "oci://" prefix:

https://github.com/GoogleContainerTools/kpt/blob/919b09de3b8c3a6663557a997a604f2379ae9232/internal/cmdreporeg/command.go#L108-L109

natasha41575: Also I looked through the email thread again and CS said "Helm uses oci:// for pushing/pulling OCI charts". Maybe I misunderstood, but I thought that meant that we can safely use "oci://" prefix?

yuwenma: kk. If this is aligned with kpt. I'm fine. 

I'm thinking of the oci module porch uses `github.com/google/go-containerregistry`.
For other format, I think container registry uses the format `https://{{REGION|ZONE}}-docker.pkg.dev/{{GCP_PROJECT}}/{{SUB_PATHS}}` 

I can't find the URL standards in the OCI repo though.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: Hm. yeah I think empty `pass` should be an error.

I updated the logic to use `As` and `corev1.Secret`.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: Changed to if statements here.

functions/go/render-helm-chart/third_party/sigs.k8s.io/kustomize/api/builtins/HelmChartInflationGenerator.go
natasha41575: Ah I'll look into `github.com/google/go-containerregistry` some more tomorrow, thanks for the tip

yuwenma: It looks like helm doesn't do any better. Seems to be good enough for us 
https://github.com/helm/helm/blob/c137bfb68b4f8b7a10687f043b0dbccb09574969/pkg/registry/util.go#L37

natasha41575: I can directly import helm's `IsOCI` function. At least then we will always be consistent with helm even if it changes.

yuwenma: That will increase the size of the function (for builtin helm in kustomize, that increases the kustomize binary), not sure whether that worths it. Let's keep your original approach.

yuwenma (approved):
After the CI tests are fixed, I think this PR is good to merge!

PR #877, opened 2022-06-07
preprocessing step for render-helm-chart private repo tests

natasha41575:
Related to https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/875

Writes the Secrets to the necessary files as a preprocessing step. I'm doing this separately from https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/875 because PRs from forks cannot access Github Secrets. So this needs to go in first, then https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/875 can use it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -103,6 +103,12 @@ jobs:
         if: matrix.platform == 'ubuntu-latest'
         run: |
           make build
+      - name: Write creds to files for render-helm-chart tests
+        if: matrix.platform == 'ubuntu-latest'
+        run: |
+          echo -e 'username=${{ secrets.RENDER_HELM_CHART_BOT_TOKEN }}\npassword=${{ secrets.RENDER_HELM_CHART_BOT_TOKEN }}' \
+          > ./examples/render-helm-chart-kustomize-private-git/credentials.env
+          echo ${{ secrets.RENDER_HELM_CHART_GCP_SERVICE_ACCOUNT }} | base64 -d > ./examples/render-helm-chart-kustomize-private-oci/password
       - name: Run all tests
         if: matrix.platform == 'ubuntu-latest'
         run: |
--- a/examples/render-helm-chart-kustomize-private-git/credentials.env
+++ b/examples/render-helm-chart-kustomize-private-git/credentials.env
@@ -1,3 +1,4 @@
-# This will contain credentials for accessing the private Git repo.
+# This will contain credentials for accessing the private Git repo. The <GH_TOKEN> is populated by
+# the CI, so this test won't work locally.
 username=<GH_TOKEN>
 password=<GH_TOKEN>
--- a/examples/render-helm-chart-kustomize-private-git/credentials.env
+++ b/examples/render-helm-chart-kustomize-private-git/credentials.env
@@ -1,0 +1,3 @@
+# This will contain credentials for accessing the private Git repo.
+username=<GH_TOKEN>
+password=<GH_TOKEN>
--- a/examples/render-helm-chart-kustomize-private-oci/password
+++ b/examples/render-helm-chart-kustomize-private-oci/password
@@ -1,1 +1,2 @@
-# This will contain the SA key for acessing the private OCI registry.
+# This will contain the SA key for acessing the private OCI registry. This file will be populated
+# by the CI, so this test won't work locally.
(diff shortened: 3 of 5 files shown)


natasha41575:
cc @mengqiy

scripts/verify-docs.py
```diff
@@ -35,7 +35,7 @@
 examples_directory = 'examples'
 functions_directory = 'functions'
 lang_dirs = ['go', 'ts']
-examples_directories_to_skip = ['_template']
+examples_directories_to_skip = ['_template', 'render-helm-chart-kustomize-private-oci', 'render-helm-chart-kustomize-private-git']
```
mengqiy, line 38: Please don't forget to remove this change in the other PR.

examples/render-helm-chart-kustomize-private-git/credentials.env
```diff
@@ -0,0 +1,3 @@
+# This will contain credentials for accessing the private Git repo.
+username=<GH_TOKEN>
```
mengqiy, line 2: Do we need the placeholders here?

.github/workflows/ci.yaml
```diff
@@ -103,6 +103,12 @@ jobs:
         if: matrix.platform == 'ubuntu-latest'
         run: |
           make build
+      - name: Write creds to files for render-helm-chart tests
+        if: matrix.platform == 'ubuntu-latest'
+        run: |
+          echo -e 'username=${{ secrets.RENDER_HELM_CHART_BOT_TOKEN }}\npassword=${{ secrets.RENDER_HELM_CHART_BOT_TOKEN }}' \
```
mengqiy, line 109: If a contributor run the tests locally, will all tests pass?

scripts/verify-docs.py
natasha41575: Left a comment to remind myself https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/875#issuecomment-1149289269

examples/render-helm-chart-kustomize-private-git/credentials.env
natasha41575: Not needed, but thought it might be helpful for us to remember where and how the credentials are stored and generated. Do you prefer we don't include these files?

.github/workflows/ci.yaml
natasha41575: No. Is there a good way to mark these directories as unable to be tested locally? 

In kustomize, we gate such tests with an environment variable that we set in CI. I could do something similar here if desired.

mengqiy: > Is there a good way to mark these directories as unable to be tested locally?

The test harness doesn't support it now. Currently we only have `skip`: https://github.com/GoogleContainerTools/kpt/blob/main/pkg/test/runner/README.md

Here is one option: You can add another config field maybe called `skipLocal` in the test harness in the kpt repo. When a developer test is locally, these tests will be skipped. When tests is run in the GH Action, we can use another env var to make all tests (including the ones with `skipLocal: true`) are tested.

If you have better solutions, please feel free to use.

examples/render-helm-chart-kustomize-private-git/credentials.env
mengqiy: No need to delete it. But I'd suggest adding more comments to make it clear that the values will be populated by the CI and it doesn't work locally.

.github/workflows/ci.yaml
natasha41575: Thanks! Yeah that's what I was thinking too. I created a PR to add this option in the kpt repo: https://github.com/GoogleContainerTools/kpt/pull/3295

examples/render-helm-chart-kustomize-private-git/credentials.env
natasha41575: Good idea. Added more details to the comment!

natasha41575:
@mengqiy is this PR blocked by anything?

examples/render-helm-chart-kustomize-private-oci/password
```diff
@@ -0,0 +1,2 @@
+# This will contain the SA key for acessing the private OCI registry. This file will be populated
+# by the CI, so this test won't work locally.
```
yuwenma, line 2: is this the secret yaml resource or user/password? I'd suggest give an example to show the required format, but using fake data

yuwenma (approved):
one minor comment. Defer to @mengqiy to approve

PR #878, opened 2022-06-21
set label refactor

zyy98:
refactor set-labels function using go/sdk instead of kyaml.


functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	transformer := LabelTransformer{}
+	if err := transformer.Config(rl.FunctionConfig); err != nil {
+		rl.Results = append(rl.Results, fn.ErrorResult(err))
+		return false, nil
+	}
+	if err := transformer.Transform(rl.Items); err != nil {
+		rl.Results = append(rl.Results, fn.ErrorResult(err))
+		return false, nil
+	}
+	return true, nil
```
yuwenma, line 29: Suggest also add a result if the set-labels passes. It can be a summary to tell users what happened and how many resources are changed.
FYI, I like your code about displaying the resource level result, but it might be verbose to users. Here's an improvement which allows users to display results by severity level.  Let me know if this makes sense to you   https://github.com/GoogleContainerTools/kpt/issues/2996

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf("FunctionConfig is missing. Expect `ConfigMap` or `SetNamespace`")
+	case o.IsGVK("", "", "ConfigMap"):
+		p.NewLabels = o.NestedStringMapOrDie("data")
+		if len(p.NewLabels) == 0 {
+			return fmt.Errorf("`data` should not be empty")
+		}
+	case o.IsGVK(fnConfigGroup, fnConfigAPIVersion, legacyFnConfigKind):
+		fallthrough
+	case o.IsGVK(fnConfigGroup, fnConfigAPIVersion, fnConfigKind):
+		p.NewLabels = o.NestedStringMapOrDie("labels")
```
yuwenma, line 57: Another option is to use `As` and defines `SetLabels` method to read labels.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	default:
+		return fmt.Errorf("unknown functionConfig Kind=%v ApiVersion=%v, expect `%v` or `ConfigMap`",
+			o.GetKind(), o.GetAPIVersion(), fnConfigKind)
+	}
+	// add default fields
+	if err := p.addDefaultLabelFields(); err != nil {
+		return err
+	}
+	// add additional fields
```
yuwenma, line 69: let's drop the additional fields

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return err
+	}
+	if !exist {
+		oldLabels = make(map[string]string)
+	}
+	result.Field.CurrentValue = copyStringMap(oldLabels)
+	replaceLabels(oldLabels, newLabels, create)
+	result.Field.ProposedValue = copyStringMap(oldLabels)
+	o.SetNestedStringMapOrDie(oldLabels, paths...)
```
yuwenma, line 145: This makes it looks like we need more result related functions in the SDK. 
I'd suggest 1. drop the resource level result to unblock this PR 2. add result methods in the Go SDK to make this kinds of `copyStringMap` easier. 3 support https://github.com/GoogleContainerTools/kpt/issues/2996 to display results by severity level 4. add back the resource level result.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil
+}
+
+// replaceLabels replace old labels map with new labels map according to create
+// oldLabels must not be nil
+func replaceLabels(oldLabels map[string]string, newLabels map[string]string, create bool) {
+	for k, v := range newLabels {
+		_, exist := oldLabels[k]
+		if create || exist {
```
yuwenma, line 154: `create` should be changed to be a user configurable field. Users can choose to replace, append, delete if not shown up a set-label field. I'm thinking of [strategic merge patch](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#basic-patch-format)
Can you add a todo here and let's discuss this with a broader audience.

yuwenma (commented):
Overall looks really good :D. Add some links and extended readings.

functions/go/set-labels/main.go
```diff
@@ -1,174 +1,14 @@
...
-	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
-	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
-	"sigs.k8s.io/yaml"
-)
-
-const (
-	fnConfigGroup      = "fn.kpt.dev"
-	fnConfigVersion    = "v1alpha1"
-	fnConfigAPIVersion = fnConfigGroup + "/" + fnConfigVersion
-	legacyFnConfigKind = "SetLabelConfig"
```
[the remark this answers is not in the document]
mengqiy: `SetLabelConfig` has been deprecated for more than 1 year.
I believe we can safely remove it

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	Results fn.Results
+}
+
+// Config parse the functionConfig kubeObject to the fields in the LabelTransformer
+func (p *LabelTransformer) Config(o *fn.KubeObject) error {
+	// parse labels to NewLabels
+	switch {
+	case o.IsEmpty():
+		return fmt.Errorf("FunctionConfig is missing. Expect `ConfigMap` or `SetNamespace`")
+	case o.IsGVK("", "", "ConfigMap"):
```
mengqiy, line 49: > o.IsGVK("", "", "ConfigMap")

It should be `o.IsGVK("", "v1", "ConfigMap")`

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	// Results is used to track labels that have been applied
+	Results fn.Results
+}
+
+// Config parse the functionConfig kubeObject to the fields in the LabelTransformer
+func (p *LabelTransformer) Config(o *fn.KubeObject) error {
+	// parse labels to NewLabels
+	switch {
+	case o.IsEmpty():
+		return fmt.Errorf("FunctionConfig is missing. Expect `ConfigMap` or `SetNamespace`")
```
mengqiy, line 48: Copy-paste error. Use`SetLabel`

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	case o.IsGVK("", "", "ConfigMap"):
+		p.NewLabels = o.NestedStringMapOrDie("data")
+		if len(p.NewLabels) == 0 {
+			return fmt.Errorf("`data` should not be empty")
+		}
+	case o.IsGVK(fnConfigGroup, fnConfigAPIVersion, legacyFnConfigKind):
```
mengqiy, line 54: We can remove this case.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	// add default fields
+	if err := p.addDefaultLabelFields(); err != nil {
+		return err
+	}
+	// add additional fields
+	if o.IsGVK(fnConfigGroup, fnConfigAPIVersion, fnConfigKind) {
+		arr, exist, err := o.NestedSlice("additionalLabelFields")
+		if err != nil {
+			return fmt.Errorf("`additionalLabelFields` format is wrong")
```
mengqiy, line 73: We should surface `err` as well.

functions/go/set-labels/main.go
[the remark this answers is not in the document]
zyy98: under tests/set-labels directory, the fn-config.yaml in those test cases uses `kind: SetLabelConfig`, if deprecated, cannot pass the test cases.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf("unknown functionConfig Kind=%v ApiVersion=%v, expect `%v` or `ConfigMap`",
+			o.GetKind(), o.GetAPIVersion(), fnConfigKind)
+	}
+	// add default fields
+	if err := p.addDefaultLabelFields(); err != nil {
+		return err
+	}
+	// add additional fields
+	if o.IsGVK(fnConfigGroup, fnConfigAPIVersion, fnConfigKind) {
+		arr, exist, err := o.NestedSlice("additionalLabelFields")
```
mengqiy, line 71: I guess you can do something like:
```
var addFields []FieldSpec
err := o.As(addFields, "additionalLabelFields")
```

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
+	for _, o := range objects {
+		for _, sp := range p.AdditionalLabelFields {
+			if (sp.Group == "" && sp.Version == "" && sp.Kind == "") || o.IsGVK(sp.Group, sp.Version, sp.Kind) {
+				newResult := fn.Result{
+					Message:     "Replace labels",
+					Severity:    "INFO",
+					ResourceRef: nil, // TODO: initialize in kyaml
```
mengqiy, line 97: Do you want to do this in this PR or in a followup PR.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	result.Field.CurrentValue = copyStringMap(oldLabels)
+	replaceLabels(oldLabels, newLabels, create)
+	result.Field.ProposedValue = copyStringMap(oldLabels)
+	o.SetNestedStringMapOrDie(oldLabels, paths...)
+	return nil
+}
+
+// replaceLabels replace old labels map with new labels map according to create
+// oldLabels must not be nil
+func replaceLabels(oldLabels map[string]string, newLabels map[string]string, create bool) {
```
mengqiy, line 151: This replacement is not going to work, since this is not modifying the original object.
You should use something like `o.Set()` to modify the original object.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,168 @@
...
+	paths := strings.Split(fieldPath, "/")
+
+	oldLabels, exist, err := o.NestedStringMap(paths...)
+	if err != nil {
+		return err
+	}
+	if !exist {
+		oldLabels = make(map[string]string)
+	}
+	result.Field.CurrentValue = copyStringMap(oldLabels)
```
mengqiy, line 142: You don't have to do a deepcopy of the map.

functions/go/set-labels/transformer/labels.go
zyy98: In the original code, it will add default type and the documentation also said that.

functions/go/set-labels/transformer/labels.go
zyy98: OK. I changed the implementation to use o.Get() and surface the error

functions/go/set-labels/transformer/labels.go
zyy98: It seems o.As only accepts one parameter. But I found the o.Get() which is very useful. I have changed my implementation now.

functions/go/set-labels/transformer/labels.go
zyy98: I have now changed my code to use `o.SetNestedString()`. It is more concise now. Thank you for your advice!

functions/go/set-labels/transformer/labels_test.go
```diff
@@ -0,0 +1,185 @@
...
+	result, _ := fn.ParseKubeObject([]byte(input))
+	err := transformer.Transform(fn.KubeObjects{result})
+	if err != nil {
+		return
+	}
+	fmt.Println(transformer.Results)
+	exp, _ := fn.ParseKubeObject([]byte(expected))
+
+	if exp.String() != result.String() {
+		fmt.Println("Actual:")
```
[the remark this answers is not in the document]
zyy98: I have read the link and now I understand I can use `cmp.diff` to compare the diff of two string. But I didn't figure out how to compare two `KubeObject`. 
Could you please provide me with some examples about comparing two KubeObject?

functions/go/set-labels/transformer/labels.go
zyy98: I reconsider this and now I think since the E2E test's result format doesn't contains the RecourceRef. So I can simply set it to be nil. Is it OK for me to leave `ResourceRef` to `nil`?

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// Config parse the functionConfig kubeObject to the fields in the LabelTransformer
+func (p *LabelTransformer) Config(o *fn.KubeObject) error {
+	// parse labels to NewLabels
+	switch {
+	case o.IsEmpty():
+		return fmt.Errorf("FunctionConfig is missing. Expect `ConfigMap` or `SetNamespace`")
+	case o.IsGVK("", "", "ConfigMap"):
+		p.NewLabels = o.NestedStringMapOrDie("data")
```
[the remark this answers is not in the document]
zyy98: I have tried using o.As(&cm). But it is not working because the `corev1.ConfigMap` is not has yaml annotation. So the As() failed. 
the comfigMap struct's Data field is: `Data map[string]string `json:"data,omitempty" protobuf:"bytes,2,rep,name=data"``, it doesn't has yaml. 
For the runner, I have looked into the code in details. Their functionality is pretty similar, I think they have the same level of simplicity. 
Do I need to switch the implementation to Runner? it is about changing 20-30 lines of code.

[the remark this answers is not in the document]
yuwenma: 😄 Yeah, this is one of the reasons I mentioned before about getting rid of the `yaml.Unmarshal`, so we do not have this limitations. Two changes are needed before dropping yaml, one is disabling the `AdditionalFieldSpec`, which Mengqi has verified it's safe to do so, the other is changing the const `commonLabelFieldSpecs`, you can change that to the `set-namespace` approach by comparing the GVK, and nestedFields.     

Using `fn.Runner` should much simplify the logic and give a good example for other fn authors.

functions/go/set-labels/main.go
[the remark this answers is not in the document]
yuwenma: can you change the kind from `SetLabelConfig` (legacy) to `SetLabel` (custom fn config type)?

yuwenma:
@zyy98 Some general suggestion to make this PR in good shape without compromising test coverage: 
- Let's make the minimum change to `./tests`, some tests may have duplicate coverage like `additional-labels` and `both-additionalLabelFields-fieldSpecs`. It's fine for now. Those tests are just to help us guarantee the refactoring does not break the UI (except additionalFieldSpec). 

- We need some more tests in `./tests`: 
  - If the input resources.yaml does not have any KRM resources
  - If the input contains wrong functionConfig Group or Kind name (for `SetLabel`)
  - If the input FunctionConfig contains wrong Data type, e.g. it's not a string map but a string struct or a slice, etc (for `ConfigMap`)

functions/go/set-labels/transformer/const.go
```diff
@@ -0,0 +1,195 @@
+package transformer
+
+const (
+	fnConfigGroup         = "fn.kpt.dev"
+	fnConfigAPIVersion    = "v1alpha1"
+	fnConfigKind          = "SetLabels"
+	commonLabelFieldSpecs = `
```
mengqiy, line 7: Does it make more sense to store it as yaml?
@yuwenma Thoughts?

tests/set-labels/both-additionalLabelFields-fieldSpecs/.expected/results.yaml
```diff
@@ -5,8 +5,8 @@ metadata:
 exitCode: 1
 items:
   - image: gcr.io/kpt-fn/set-labels:unstable
-    stderr: '[error]: failed to configure function: `fieldSpecs` has been deprecated, please rename it to `additionalLabelFields`'
+    stderr: 'failed to evaluate function: `additionalLabelFields` has been deprecated'
```
mengqiy, line 8: Are we going to to deprecate `additionalLabelFields`?

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,137 @@
...
+
+	rl.Results = append(rl.Results, transformer.Results...)
+	return true, nil
+}
+
+type LabelTransformer struct {
+	// Desired labels
+	NewLabels map[string]string `json:"labels,omitempty" yaml:"labels,omitempty"`
+	// FieldSpecs is deprecated, please use AdditionalLabelFields instead.
+	FieldSpecs []FieldSpec `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
```
mengqiy, line 39: Why are we using `fieldSpecs` as tag name?
Is the comment above up to date?

zyy98: Yes I should update the comment, FieldSpecs is used to store commonLabelFieldSpecs

tests/set-labels/both-additionalLabelFields-fieldSpecs/.expected/results.yaml
yuwenma: Yes. `set-labels` is a general function and the `additionalLabelFields` is more like a temporary workaround for me. 
To support  `additionalLabelFields`, verifying the custom fields do mean "labels" are necessary. Also, it should avoid overriding the core k8s resources that do not mean "labels". From that perspective, I think we either need  OpenApi schema or let users keep their own function for custom labels, and  `additionalLabelFields` seems not to be the proper solution.

tests/set-labels/both-additionalLabelFields-fieldSpecs/Kptfile
```diff
@@ -6,3 +6,6 @@ pipeline:
   mutators:
     - image: gcr.io/kpt-fn/set-labels:unstable
       configPath: fn-config.yaml
+      selectors:
+        - kind: ConfigMap
+        - kind: MyResource
```
yuwenma, line 11: what is this change for?

tests/set-labels/legacy-config/.expected/diff.patch
```diff
@@ -21,7 +21,4 @@ index 1e641f7..0dde4b3 100644
 +    fruit: apple
  spec:
    selector:
-     labels:
-       foo: bar
```
yuwenma, line 25: this change is skeptical. Falling legacy function config to `SetLabels` should give the same result, right?

tests/set-labels/legacy-config/fn-config.yaml
```diff
@@ -1,13 +1,7 @@
 apiVersion: fn.kpt.dev/v1alpha1
-kind: SetLabelConfig
+kind: SetLabels
```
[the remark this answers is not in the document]
yuwenma: yeah, we discussed this and we will remove the tests in the following up PR. Because this PR refactors the entire function and we do want to make sure it is under good test coverage. 
 Here's the follow-up work https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/878#issuecomment-1167671033

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,137 @@
...
+	"strings"
+)
+
+// the similar struct esxits in resid.GVK, but there is no function to create an GVK struct without using kyaml
+type FieldSpec struct {
+	Group              string `json:"group,omitempty" yaml:"group,omitempty"`
+	Version            string `json:"version,omitempty" yaml:"version,omitempty"`
+	Kind               string `json:"kind,omitempty" yaml:"kind,omitempty"`
+	Path               string `json:"path,omitempty" yaml:"path,omitempty"` // seperated by /
+	CreateIfNotPresent bool   `json:"create,omitempty" yaml:"create,omitempty"`
```
yuwenma, line 17: Can you also add a TODO here about we will change this "createIfNotPresent" to be a configurable field?

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func SetLabels(rl *fn.ResourceList) (bool, error) {
+	transformer := LabelTransformer{}
+	if err := transformer.Config(rl.FunctionConfig); err != nil {
+		rl.Results = append(rl.Results, fn.ErrorResult(err))
+		return false, err
```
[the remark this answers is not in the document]
yuwenma: you can check the difference of return `true/false` via `kpt fn render` (e.g. pipeline.mutators set-labels, set-namespace) and make set-labels fails.

functions/go/set-labels/transformer/labels.go
yuwenma: +1

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+		}
+	}
+	return nil
+
+}
+
+func (p *LabelTransformer) addDefaultLabelFields() error {
+	var defaultFieldSpecs []FieldSpec
+	err := json.Unmarshal([]byte(commonLabelFieldSpecs), &defaultFieldSpecs)
```
yuwenma, line 131: Can you add a TODO here to convert commonLabelFieldSpecs to use the IsGVK format?

functions/go/set-labels/transformer/const.go
yuwenma: +1 to change this. 
I think it is better to use the IsGVK and other KRM resource type specific functions (We can provide some more functions on the SDK side to make it simpler). 

Json/Yaml parsing the resources (even with `json`) set the wrong view on users for KRM resources editing. We expect the functions to change KRM resource by its definition (so relying on OpenApi schema is reasonable), not by the yaml/json raw data (similar as helm values and kustomize vars).

tests/set-labels/legacy-config/fn-config.yaml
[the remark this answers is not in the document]
mengqiy: Got it.

tests/set-labels/both-additionalLabelFields-fieldSpecs/.expected/results.yaml
mengqiy: `additionalLabelFields` is a short-term solution. Using OpenAPI schema is the long term solution, but this solution is still not complete yet.
`additionalLabelFields` is intended to be a lightweight solution until we have the OpenAPI solution ready. If a user only want to change one more label field in their CRD, it'd be too heavy weight to ask them to build a custom function.

> it should avoid overriding the core k8s resources that do not mean "labels"

This may happen if a user configure it wrong.

IMO if we deprecate it now, we actually increase the barrier for adoption. WDYT?

examples/set-labels-advanced/fn-config.yaml
```diff
@@ -7,9 +7,3 @@ metadata:
 labels:
   color: orange
   fruit: apple
-additionalLabelFields:
-  - kind: MyResource
-    create: true
-    group: dev.example.com
-    path: spec/selector/labels
-    version: v1
```
mengqiy, line 15: IMO it doesn't make a lot of sense to remove these lines even if we want to deprecated `additionalLabelFields`.
We can change the example name to something like `set-labels-deprecated`.
If some users are using it, it can be a problem if they can't find an example.

functions/go/set-labels/transformer/const.go
```diff
@@ -0,0 +1,196 @@
+package transformer
+
+const (
+	fnConfigGroup         = "fn.kpt.dev"
+	fnConfigAPIVersion    = "v1alpha1"
+	fnConfigKind          = "SetLabels"
+	fnDeprecateField      = "additionalLabelFields"
+	CommonLabelFieldSpecs = `
```
mengqiy, line 8: Have we agreed to change this yaml?

The content here doesn't appear to be used anywhere.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,235 @@
...
+
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+// the similar struct esxits in resid.GVK, but there is no function to create an GVK struct without using kyaml
+type FieldSpec struct {
+	Group              string `json:"group,omitempty" yaml:"group,omitempty"`
+	Version            string `json:"version,omitempty" yaml:"version,omitempty"`
+	Kind               string `json:"kind,omitempty" yaml:"kind,omitempty"`
+	Path               string `json:"path,omitempty" yaml:"path,omitempty"` // seperated by /
```
mengqiy, line 17: It's more visible to change it from line comment to head comment.

functions/go/set-labels/transformer/labels.go
mengqiy: > we will change this "createIfNotPresent" to be a configurable field

@yuwenma Can you please explain this a bit?

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,235 @@
...
+
+	for _, o := range objects {
+		var path = "metadata/labels"
+		err := updateLabels(o, path, p.NewLabels, true)
+		if err != nil {
+			return err
+		}
+		p.LogResult(o, path)
+		// check all default path to update labels
+		switch {
```
mengqiy, line 80: I believe you can simplify the logic below by using the config of `CommonLabelFieldSpecs` and `fieldSpec`.
Currently it seems you are hardcoding it below.

zyy98: I thought it is best not to use json or yaml unmarsh to read in the **CommonLabelFieldSpecs**, and I should use **IsGVK**, which is a bit hard-coding as I write here. Is this what you expect or I misunderstood the point? @yuwenma

functions/go/set-labels/go.mod
```diff
@@ -3,31 +3,35 @@ module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-la
 go 1.17
 
 require (
-	github.com/stretchr/testify v1.7.0
-	sigs.k8s.io/kustomize/api v0.10.2-0.20220110181313-a1a0a4982226
-	sigs.k8s.io/kustomize/kyaml v0.13.2-0.20220110181313-a1a0a4982226
-	sigs.k8s.io/yaml v1.2.0
+	github.com/stretchr/testify v1.7.1 // indirect
+	sigs.k8s.io/kustomize/kyaml v0.13.7-0.20220418212550-9d5491c2e20c // indirect
```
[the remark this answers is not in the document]
yuwenma: @zyy98 please fix as Mengqi suggested.

examples/set-labels-advanced/fn-config.yaml
yuwenma: Why cannot users find examples? I think the idea is that users find docs and examples from the website, which points to a specific released version. I guess that could be more confusing to keep an unsupported example.

functions/go/set-labels/transformer/const.go
yuwenma: +1, please cleanup @zyy98

functions/go/set-labels/transformer/labels.go
yuwenma: agree. @zyy98 do you mind explaining why we need json tag these fields?

functions/go/set-labels/transformer/labels.go
yuwenma: I think we want to avoid `path = "metadata/labels"` as well.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,235 @@
...
+			path = "spec/template/spec/affinity/podAntiAffinity/preferredDuringSchedulingIgnoredDuringExecution/podAffinityTerm/labelSelector/matchLabels"
+			err = updateLabels(o, path, p.NewLabels, false)
+		case o.IsGVK("apps", "", "StatefulSet"):
+			path = "spec/template/spec/affinity/podAntiAffinity/requiredDuringSchedulingIgnoredDuringExecution/labelSelector/matchLabels"
+			err = updateLabels(o, path, p.NewLabels, false)
+		case o.IsGVK("apps", "", "StatefulSet"):
+			path = "spec/template/spec/topologySpreadConstraints/labelSelector/matchLabels"
+			err = updateLabels(o, path, p.NewLabels, false)
+		case o.IsGVK("apps", "", "StatefulSet"):
+			path = "spec/volumeClaimTemplates[]/metadata/labels"
```
yuwenma, line 148: volumeClaimTemplates is a slice. (Suggest checking the struct of `StatefulSet`)? line 213 won't be able to handle this type. please consider using the SDK method.

yuwenma (approved):

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,235 @@
...
+	sort.Strings(keys)
+	for i := 0; i < len(keys); i++ {
+		key := keys[i]
+		val := newLabels[key]
+		newPath := append(basePath, key)
+		_, exist, err := o.NestedString(newPath...)
+		if err != nil {
+			return err
+		}
+		if exist || create {
```
yuwenma, line 227: can we add a TODO about changing create to be a configurable field?

functions/go/set-labels/transformer/labels.go
yuwenma: @mengqiy and I have an offline discussion about the code style here.
Something you can consider.

```go
type FieldPath []string
type GVK []string
// for fieldpath that has simple stringMap field.   
stringMapLabels := map[fn.ResourceIdentifier]FieldPath{
        GVK{"", "v1", "Service"}: FieldPath{"spec", "selectors"} 
      .........
}

for _, o := range objects {
          o.SetLabels()... // or SetLabel
            // for pure stringMap fieldspec
           setLabelsInSpecs(o)
           // handle special cases
           if o.IsGVK("apps", "", "StatefulSet") {
                  for _, vctObj:=  range o.GetMap("spec").GetSlice("volumeClaimTemplates")  {
                        setVolumeClaimTemplatesLabel(vctObj)
                    }            
           }
}

func setLabelsInSpecs(o fn.KubeObject) {
           for gvk, paths := range stringMapLabels { 
                     if o.IsGVK(gvk...) {
                             o.SetNestedStringMap(newLabels, paths...) 
                     }
            }
}
```

functions/go/set-labels/transformer/common.go
```diff
@@ -0,0 +1,162 @@
+package transformer
```
yuwenma, line 1: nit: copyright.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+type FieldPath []string
+type GVK struct {
+	group   string
+	version string
+	kind    string
+}
+
+// the similar struct esxits in resid.GVK, but there is no function to create an GVK struct without using kyaml
```
yuwenma, line 19: typo:  "esxits"

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+type FieldSpec struct {
+	Identifier         GVK
```
yuwenma, line 21: can we just call it GVK? because GVK is not an identifier.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+	Path               FieldPath
```
yuwenma, line 22: ditto, it seems to be easier to understand by just naming it FieldPath

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+	CreateIfNotPresent bool
+	//TODO: should support user configurable field for labels
```
yuwenma, line 24: super nit:
```go
// TODO(zyy98): ...
CreateIfNotPresent bool
```

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type FieldSpecs []FieldSpec
+
+type LabelTransformer struct {
+	// Desired labels
+	NewLabels map[string]string
+	// FieldSpecs storing default label fields
+	FieldSpecs []FieldSpec
+	// Results is used to track labels that have been applied
```
yuwenma, line 34: nit: "that have been applied" infers the change happens before set-labels.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+	Results fn.Results
+}
+
+// perform the whole set labels operation according to given resourcelist
+func SetLabels(rl *fn.ResourceList) (bool, error) {
```
yuwenma, line 39: nit: godoc shall start with the function name so this docstring need to be `// SetLabels performs ...`

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return nil
+}
+
+// set labels according to the generated common label
+func (p *LabelTransformer) setLabelsInSpecs(o *fn.KubeObject) error {
+	for _, spec := range CommonSpecs {
+		if o.IsGVK(spec.Identifier.group, spec.Identifier.version, spec.Identifier.kind) {
+			err := updateLabels(&o.SubObject, spec.Path, p.NewLabels, spec.CreateIfNotPresent)
+			p.LogResult(o, spec.Path)
```
yuwenma, line 82: should ln82 be moved after the error handling in ln83-85?

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Transform updates the labels in the right path using configured logic
```
yuwenma, line 91: nit: "in the right path using configured logic" is a little bit unclear.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
+	for _, o := range objects {
+		// this label need to set for all GKV
```
yuwenma, line 94: nit: GKV -> GVK?

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+		defaultPath := FieldPath{"metadata", "labels"}
+		err := updateLabels(&o.SubObject, defaultPath, p.NewLabels, true)
+		p.LogResult(o, defaultPath)
+		if err != nil {
+			return err
+		}
+		// set other common labels according to specific GKV
```
yuwenma, line 101: super nit: GKV-> GVK

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+		err = p.setLabelsInSpecs(o)
+		if err != nil {
+			return err
+		}
+		// handle special case with slice
+		if o.IsGVK("apps", "", "StatefulSet") {
+			for _, vctObj := range o.GetMap("spec").GetSlice("volumeClaimTemplates") {
```
yuwenma, line 108: this can be panic when "spec" return nil.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// Logs the result of each operation, can also modify into other logs user wants
```
yuwenma, line 120: ditto: `// LogResult...`

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+func (p *LabelTransformer) LogResult(o *fn.KubeObject, path []string) {
+	res, _ := json.Marshal(p.NewLabels)
```
yuwenma, line 122: Do we still have to use `json.Marshal`?

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	sort.Strings(keys)
+	for i := 0; i < len(keys); i++ {
+		key := keys[i]
+		val := newLabels[key]
+		newPath := append(labelPath, key)
+		_, exist, err := o.NestedString(newPath...)
+		if err != nil {
+			return err
+		}
```
yuwenma, line 155: nit: o.NestedStringOrDie?

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return nil
+}
+
+// Transform updates the labels in the right path using configured logic
+func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
+	for _, o := range objects {
+		// this label need to set for all GKV
+		defaultPath := FieldPath{"metadata", "labels"}
```
yuwenma, line 95: nit: "default" is not appropriate here. Maybe metaLabelsPath

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+	keys := make([]string, 0)
+	for k := range newLabels {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for i := 0; i < len(keys); i++ {
+		key := keys[i]
+		val := newLabels[key]
+		newPath := append(labelPath, key)
+		_, exist, err := o.NestedString(newPath...)
```
yuwenma, line 152: Are you appending the label key to the fieldpaths?  Does it work?

yuwenma: well. do you think it's possible to use `SetNestedStringMap` and `NestedStringMap` here

yuwenma:
Overall it looks much better! Good job @zyy98  leave several minor comments.

functions/go/set-labels/transformer/labels.go
yuwenma: Here's the guidance https://go.dev/blog/godoc

functions/go/set-labels/transformer/labels.go
zyy98: the result format is using json format to record the changes. So using json package is the most straightforward way

functions/go/set-labels/transformer/labels.go
zyy98: append works here and pass the test cases. Using SetNestedStringMap would overwrite the labels field, but here I want to upsert the labels.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,167 @@
...
+type FieldSpec struct {
+	Gvk       GVK
+	FieldPath FieldPath
+	// TODO: should support user configurable field for labels
+	CreateIfNotPresent bool
+}
+
+type FieldSpecs []FieldSpec
+
+// LabelTransformer stores information during the transform label process
```
yuwenma, line 29: this struct does the actual "transform" operation. Can we change this to something like providing the main workflow ... it uses Config to parse functionConfig... it uses transform to change the labels ...

functions/go/set-labels/transformer/labels.go
yuwenma: kk.

functions/go/set-labels/transformer/labels.go
yuwenma: how does the `set-labels:v0.1.5` sort the label keys for the following case:
```yaml
// function config
kind: ConfigMap
data:
    aNewKey1: val1,
    bNewKey2: val2,
    cModifyKey3: val3,
    dModifyKey4: val4,
    eNewKey5: val5,
    fNewKey6: val6 
```
```yaml
// one of the ResourceList.Items
Kind: Deployment
metadata:
   labels:
       aOtherKey1: val,
       aOtherKey2: val,
       cModifyKey3: val,
       dModifyKey4: val,
       eOtherKey5: val,
       fOtherKey6: val
```

tests/set-labels/legacy-config/.expected/diff.patch
yuwenma: nice fix!

tests/set-labels/empty-list/simple/.expected/results.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 exitCode: 1
 items:
   - image: gcr.io/kpt-fn/set-labels:unstable
-    stderr: '[error]: failed to configure function: `functionConfig` must be either a `ConfigMap` or `SetLabels`'
+    stderr: 'failed to evaluate function: error: function failure'
```
yuwenma, line 8: Do you know why this does not give "`functionConfig` must be either a `ConfigMap` or `SetLabels`" error?
I think for empty function config, giving "`ConfigMap` or `SetLabels`" is more helpful.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,167 @@
...
+	// Results records the operations performed, user can log here what information they want
+	Results fn.Results
+}
+
+// SetLabels perform the whole set labels operation according to given resourcelist
+func SetLabels(rl *fn.ResourceList) (bool, error) {
+	transformer := LabelTransformer{}
+	if err := transformer.Config(rl.FunctionConfig); err != nil {
+		rl.Results = append(rl.Results, fn.ErrorResult(err))
+		return false, nil
```
yuwenma, line 44: `return true, nil`
See the comment below

tests/set-labels/both-additionalLabelFields-fieldSpecs/Kptfile
```diff
@@ -6,3 +6,6 @@ pipeline:
   mutators:
     - image: gcr.io/kpt-fn/set-labels:unstable
       configPath: fn-config.yaml
+      selectors:
```
yuwenma, line 9: Why do we use selectors here?

zyy98: The original file does not have selectors because it fails before applying the selectors. Here it is added because I modified the test to perform the actual operation (this test is duplicate coverage and should be merged with additional- labels later)

tests/set-labels/empty-list/simple/.expected/results.yaml
zyy98: I can modify the error message here. Using (true, nil) would cause confusion in fn render, as the function would be marked PASS instead of FAIL

functions/go/set-labels/transformer/labels.go
yuwenma: yeah, I'm thinking of sorting the labels *after* the update: basically consider the labels as a whole (you can do that via `NestedStringMap` --> update labels --> `SetNestedStringMap`).

zyy98: I think maps cannot be sorted, SetNestedStringMap cannot ensure the sequence. It would be more effort by setting the original one to nil, then iterate all sorted keys using SetNestedString

yuwenma: Not sorting the map. I mean firstly merging the current label keys and the new label keys, and then sort the new label slice, and based on the sorted slice to update the whole labels. But I can see the downside that git will show more diff if the existing labels' order are changed.
@mengqiy Any insights what label order may be more useful to users?

mengqiy: > But I can see the downside that git will show more diff if the existing labels' order are changed.

We should avoid unnecessary diff since it causes noise but not provide useful info.

functions/go/set-labels/transformer/labels.go
```diff
@@ -0,0 +1,164 @@
...
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+type FieldPath []string
+type GVK struct {
```
mengqiy, line 13: Please add a new line.

functions/go/set-labels/transformer/labels.go
mengqiy: > esxits

typo

functions/go/set-labels/transformer/labels.go
mengqiy, line 125: The results generated this way is not always useful. It should only show the labels that have been added or updated.
For example, running the set-labels functions with the same config again should not generate any results since labels have been updated in the first run.
cc: @yuwenma

functions/go/set-labels/transformer/labels.go
yuwenma: Exactly, that's what I'm concerning about. On the other hand, appending new labels to the end and updating existing labels in place can end up a non-alphabetical label order. It's unclear to me whether it could have potential problems. 

Anyway, I think current approach is fair.

functions/go/set-labels/transformer/labels_test.go
```diff
@@ -0,0 +1,332 @@
...
+    quotedBoolean: "true"
+    quotedFruit: peach
+    unquotedBoolean: "true"
+`
+
+	transformer := LabelTransformer{}
+	config, _ := fn.ParseKubeObject([]byte(functionConfig))
+	_ = transformer.Config(config)
+	result, _ := fn.ParseKubeObject([]byte(input))
+	err := transformer.Transform(fn.KubeObjects{result})
```
yuwenma, line 58: super nit: you can use `ResourceList` to define the inputs (which contains both functionConfig and the KRM resources) and call `ParseResourceList`

functions/go/set-labels/transformer/labels_test.go
```diff
@@ -0,0 +1,332 @@
+package transformer
```
yuwenma, line 1: Some general suggestions for writing go unittests. 
- using for loop to parameterize the test cases.   See below example
- use some testdata generator method to reuse/simplify the code. See example `generateResourceList` below
example 1
```go
testdata := map[string]struct{
     resourcelist string
     expected string
} { 
      ""add labels to selector": {
           resourcelist: generateResourceList(cm, services)
          expected: ...
      } 
       ....
}

func generateResourceList(functionConfig string, singleItem string) {
        data := fmt.Sprintf("functionConfig: %s\nitems: \n... %s", functionConfig, singleItem)
        rl, _ := fn.ParseResourceList(data)
        return rl 
}
```

functions/go/set-labels/transformer/labels_test.go
yuwenma: you can make the suggested change in a follow-up PR.

zyy98: I am curious why we need to use `ResourceList` here. Is it because it is more intuitive for users, since each operation is considered transformation on resoureList? Because 
`data := fmt.Sprintf("functionConfig: %s\nitems: \n... %s", functionConfig, singleItem)` is actually not easy to format with indentation. If only consider the code simplicity, it would be best to parse config and resource separately

functions/go/set-labels/transformer/labels_test.go
zyy98: Using for loop means there is only one unit test to run all the tests? I think it might not be good for debugging. I should reuse the code and I will fix that part.

functions/go/set-labels/transformer/labels_test.go
yuwenma: I was sharing some pseudo code off the top of my head, agree the `fmt.Sprint` is not a good format.
In the fn sdk, we basically defines the entire `ResourceList` as input. the main reason I vote down on parsing the fnConfig and parsing the input resources separately is that KRM function testing more or less need a lot of code to generate testing resource. It's much less maintainable to handle each type of function resource separately than putting them all in the ResourceList in the form of yaml (string) format. 

Also, this gives users a more directly understanding about what the expected input is (ResourceList Yaml) and what the output is (modified ResourceList).

functions/go/set-labels/transformer/labels_test.go
yuwenma: that's true, that's why the sample code uses the map key which I expect to describe the test case.
This is more or less a go convention for test parameterization.

functions/go/set-labels/transformer/labels_test.go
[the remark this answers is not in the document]
zyy98: Yes that can work. I also applied assert in later update that compares 2 strings

tests/set-labels/both-additionalLabelFields-fieldSpecs/.expected/results.yaml
yuwenma: Per offline discussion, we need a more general solution for custom paths rather than relying on OpenApi which has a very long timeline to design and publish new features. 
We will drop additionalLabelFields for now.

PR #885, opened 2022-07-02
New set-namespace with origin logic

yuwenma:
## New Behavior
The `set-namespace` uses `internal.kpt.dev/upstream-identifier` annotation to decide whether resources are under the same original namespace. 
- It squash the local resources (no upstream origin) to the new value.
- It replaces the derived resources (with upstream origin) to the new value only if the origin are the same.
- It requires all derived resources under the same origin
- It gives warnings for raw KRM resources that do not have upstream origin. 
- It does not validate resource duplication, because kpt can accept duplicate GVKNN resources if they are under different kpt packages. 
- It provides `namespaceMatcher` to specifically match and replace a namespace. Using `namespaceMatcher` no longer requires unique origin. 
- `namespaceMatcher` matches both current namespace and the origin namespace. 

### TODO
- update README
- more examples to verify the following CUJs
 
#### No-op if the new and current namespace are the same. 
```bash
Results:
  [warning]: Resource(apiVersion=v1, kind=Namespace, namespace=, name=example) does not have upstream origin.
  [warning]: Resource(apiVersion=rbac.authorization.k8s.io/v1, kind=RoleBinding, namespace=example, name=app-admin) does not have upstream origin.
  [info]: all matching namespaces are already "example". no value changed
  [info]: all `depends-on` annotations are up-to-date. no `namespace` changed
```
#### Changes all “example” namespace to new namespace “basens”.
```bash
Results:
  [warning]: Resource(apiVersion=v1, kind=Namespace, namespace=, name=example) does not have upstream origin.
  [warning]: Resource(apiVersion=rbac.authorization.k8s.io/v1, kind=RoleBinding, namespace=example, name=app-admin) does not have upstream origin.
  [info]: namespace [example] updated to "basens", 2 value(s) changed
  [info]: all `depends-on` annotations are up-to-date. no `namespace` changed
```
#### Changes newly added resources to new namespace “basens” 
- If kube-gen adds a new resource (quota) after all other resources have been updated to “basens”, set-namespace will update the newly added resource to “basens”. 
- If users edit/change the existing namespace value, set-namespace will update the changed resource to “basens”.   
```bash
Results:
  [warning]: Resource(apiVersion=v1, kind=Namespace, namespace=, name=basens) does not have upstream origin.
  [warning]: Resource(apiVersion=v1, kind=ResourceQuota, namespace=example, name=default) does not have upstream origin.
  [warning]: Resource(apiVersion=rbac.authorization.k8s.io/v1, kind=RoleBinding, namespace=basens, name=app-admin) does not have upstream origin.
  [info]: namespace [example] updated to "basens", 1 value(s) changed
  [info]: all `depends-on` annotations are up-to-date. no `namespace` changed
```
#### Replace a specific namespace if the package contains multiple namespaces. 
Use namespaceMatcher in function config to specify the existing namespace value to update. 
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kptfile.kpt.dev
  annotations:
data:
  # "namespace: app" is required if not kptfile.kpt.dev
  name: app 
  namespaceMatcher: example
```
#### Replace multiple sets of namespace
If users want to update namespaces to more than one new values, they can run set-namespace in a pipeline with different functionConfig values.  
```yaml
// Run N set-namespace to replace N namespace from X to Y
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: backend
  annotations:
pipeline:
  mutators:
    - exec: set-ns
      configMap:
        namespace: basens
        namespaceMatcher: example
    - exec: set-ns
      configMap:
        namespace: new-app
        namespaceMatcher: custom
```
```bash
[RUNNING] "set-ns"
[PASS] "set-ns" in 0s
  Results:
    [info]: all matching namespaces are already "basens". no value changed
    [info]: all `depends-on` annotations are up-to-date. no `namespace` changed
[RUNNING] "set-ns"
[PASS] "set-ns" in 0s
  Results:
    [info]: namespace [custom] updated to "new-app", 1 value(s) changed
    [info]: all `depends-on` annotations are up-to-date. no `namespace` changed
```
#### Change namespace from upstream “example” to a local “backend”
set-namespace ignores the resource current namespace value, but relies on its origin to decide the correlation of each resource. Derived resources can have different namespace values,   set-namespace updates them all as long as their origin namespace is the same.
```bash
Results:
  [info]: namespace [example] updated to "backend", 4 value(s) changed
  [info]: all `depends-on` annotations are up-to-date. no `namespace` changed
```
#### Fail the execution if more than one upstream namespace is detected.
```bash
Results:
  [error]: unable to use origin `namespace` to match. expect a single value, found [example kube-public]. please use `namespaceMatcher`to specify the namespace value you want to change
```
#### Change local new resource to “backend”, ignoring its origin.
Users can customize their package by adding new resources via kube-gen. These resources use “example” as namespace. The derived resources have “backend” as namespace with origin “example”.
set-namespace can change “example” to “backend”  
```bash
Results:
  [warning]: Resource(apiVersion=v1, kind=ResourceQuota, namespace=example, name=default) does not have upstream origin.
  [info]: namespace [example] updated to "backend", 1 value(s) changed
  [info]: all depends-on annotations are up-to-date. no namespace changed
```
#### Replace a specific origin match namespace. 
The derived package has two origin namespaces kube-public and example. Users can use namespaceMatcher to only update the resources having kube-public origin.
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kptfile.kpt.dev
  annotations:
data:
  name: basens
  namespaceMatcher: kube-public
```
```bash
Results:
  [info]: [Resource(apiVersion=v1, kind=ResourceQuota, namespace=custom, name=default)] has matching origin kube-public
  [info]: namespace [custom] updated to "basens", 1 value(s) changed
  [info]: all `depends-on` annotations are up-to-date. no namespace changed
```
#### Replace a specific matching namespace.
namespaceMatcher matches both origin and current namespace field.

#### Replace a set of specific origin namespaces in a pipeline
If users want to update resources from a multi-namespace upstream package. They can run set-namespace in a pipeline with different functionConfig values.  
```yaml
// Run N set-namespace to replace N namespace from X to Y
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: basens
  annotations:
info:
  description: kpt package for provisioning namespace
pipeline:
  mutators:
    - exec: set-ns
      configMap:
        namespace: app
        namespaceMatcher: kube-public
    - exec: set-ns
      configMap:
        namespace: backend
        namespaceMatcher: example
```
```bash
[RUNNING] "set-ns"
[PASS] "set-ns" in 0s
  Results:
    [info]: [Resource(apiVersion=v1, kind=ResourceQuota, namespace=custom, name=default)] has matching origin kube-public
    [info]: namespace [custom] updated to "app", 1 value(s) changed
    [info]: all `depends-on` annotations are up-to-date. no `namespace` changed
[RUNNING] "set-ns"
[PASS] "set-ns" in 0s
  Results:
    [info]: [Resource(apiVersion=v1, kind=Namespace, namespace=, name=basens)] has matching origin example
    [info]: [Resource(apiVersion=rbac.authorization.k8s.io/v1, kind=RoleBinding, namespace=basens, name=app-admin)] has matching origin example
    [info]: [Resource(apiVersion=v1, kind=Service, namespace=basens, name=my-service)] has matching origin example
    [info]: namespace [basens] updated to "backend", 3 value(s) changed
```
#### Change namespace on new resource from upstream package
kpt pkg update adds a new resource Service from upstream to the local downstream.  Service has the same origin but different current namespace. Set-namespace updates its namespace from “example” to “backend”
```bash
Results:
  [info]: namespace [example] updated to "backend", 1 value(s) changed
  [info]: all depends-on annotations are up-to-date. no namespace changed
```


yuwenma:
@droot Thank you for listing the scenarios. I added the new behavior and rationale here https://docs.google.com/document/d/1YwxalwS45V1BhvzwWrgHJQPL2mZ1xchIzTm_eI6jXsk/edit?pli=1#heading=h.8axlv8ip44x7

To give a quick answer for your scenarios:

set-namespace are origin based and considered from the package revision level. raw resource without origin are considered as a draft, and I am leaning torwards making it as convenient to edit as possible. `set-namespace` does a squash behavior for no-origin resources.
  -   Abstract packages always have its resource changed to new namespace, whether they have different namespaces or not.
  - For non k8s core resources, unset namespace (no "metadata.namesapce" field) will be considered as cluster-scoped. Note: metadata.namespace = "" is considered as namespace set with value "default".  
  - For k8s core resources, set-namespace knows its scope and always update namespace-scoped resource. 
  - add new resource in the derived package, as long as it does not have origin, will always have their namespace changed to the new value. This is a squash
  -  set-namespace gives user message to differentiate the origin and non-origin resources, and this  tells what kinds of change is made. I listed the user message here https://docs.google.com/document/d/1YwxalwS45V1BhvzwWrgHJQPL2mZ1xchIzTm_eI6jXsk/edit?pli=1#heading=h.8axlv8ip44x7

natasha41575:
> Changes all “example” namespace to new namespace “basens”.
> ```
> Results:
>   [warning]: Resource(apiVersion=v1, kind=Namespace, namespace=, name=example) does not have upstream origin.
>   [warning]: Resource(apiVersion=rbac.authorization.k8s.io/v1, kind=RoleBinding, namespace=example, name=app-admin) does not have upstream origin.
>   [info]: namespace [example] updated to "basens", 2 value(s) changed
>   [info]: all `depends-on` annotations are up-to-date. no `namespace` changed
> ```

I have some question about whether giving a warning is the right behavior for root packages. I think it would make sense to throw this warning on derived packages, but for a root package it is expected that there will be no upstream origin. The user might be confused why are they getting a warning when this is expected behavior. 

We could potentially check the `Kptfile` to see if this is a root package, and only throw the warning if either Kptfile is not there or if we see that this is a derived package. WDYT?

(Apologies, I probably should have left this comment on your doc but didn't think of this when I read that part.)

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,310 @@ import (
...
-	return fn.GeneralResult(msg, fn.Info)
+// SetNamespace defines structs to parse KRM resource "SetNamespace" (the custom function config) and "ConfigMap" data.
+// it provides the method "Config" to read the function configs from ResourceList.FunctionConfig
+// it provides the method "Transform" to change the "namespace" and update the "config.kubernetes.io/depends-on" annotation.
+type SetNamespace struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+	NewNamespace      string            `json:"namespace,omitempty"`
+	NamespaceMatcher  string            `json:"namespaceMatcher,omitempty"`
+	Data              map[string]string `json:"data,omitempty"`
```
natasha41575, line 49: IMO it makes more sense to have a struct that just defines the `SetNamespace` type and not conflate it with `ConfigMap`. This is a little bit confusing, and I think it would be more clear what is happening here if we separate another struct called `ConfigMap` for the data field (or use the `corev1.ConfigMap` struct).

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,310 @@ import (
...
-				if p.NewNamespace == "" {
-					return fmt.Errorf("`data.name` should not be empty")
-				}
-			} else {
-				return fmt.Errorf("`data.namespace` should not be empty")
-			}
+		return fmt.Errorf("FunctionConfig is missing. Expect `ConfigMap(v1)` or `%s(%s/%s)`",
+			fnConfigKind, fnConfigGroup, fnConfigVersion)
+	case o.IsGVK("", "v1", "ConfigMap"):
+		o.AsOrDie(&p)
```
natasha41575, line 62: If we take my suggest above to separate `SetNamespace` from `ConfigMap` structs, we can initialize the `ConfigMap` type here and then copy its data field into `p`'s fields. 

We're already copying the `data` field into `p`'s other fields, so the separation wouldn't add any complexity.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,310 @@ import (
...
-			o.SetOrDie(&subjects, "subjects")
-		}
-		// "IsNamespaceScoped" detects resource scope by checking whether `metadata.namespace` is set.
-		// This is valid hypothesis is for custom k8s resources. For k8s resources, the function always gives accurate scope information.
-		if o.IsNamespaceScoped() {
-			namespace := o.NestedStringOrDie("metadata", "namespace")
-			visitor(NewNamespace(o, &namespace))
-			o.SetOrDie(&namespace, "metadata", "namespace")
+			if origin.Namespace == fn.UnknownNamespace {
+				// This should rarely happen.
```
natasha41575, line 149: In what case would this happen?

functions/go/set-namespace/transformer/consts.go
```diff
@@ -14,24 +14,21 @@
...
 
 const (
-	fnConfigAPIVersion  = "fn.kpt.dev/v1alpha1"
-	fnConfigKind        = "SetNamespace"
-	dependsOnAnnotation = "config.kubernetes.io/depends-on"
-	groupIdx            = 0
-	namespaceIdx        = 2
-	kindIdx             = 3
-	nameIdx             = 4
+	// Constants for FunctionConfig `SetLabel`
```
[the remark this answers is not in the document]
yuwenma: thanks! good catch

functions/go/set-namespace/transformer/namespace.go
yuwenma: Thanks! That works for me.

functions/go/set-namespace/transformer/namespace.go
yuwenma: done.

examples/set-namespace-depends-on/.expected/diff.patch
```diff
@@ -34,3 +35,8 @@ index e45cae5..2ff45d5 100644
  ---
  apiVersion: rbac.authorization.k8s.io/v1
  kind: ClusterRoleBinding
+@@ -43,3 +43,4 @@ subjects:
+ - apiGroup: rbac.authorization.k8s.io
+   kind: Group
+   name: admin
++  namespace: example-ns
```
mengqiy, line 42: Can you please confirm if this change is expected?

examples/set-namespace-depends-on/resources.yaml
```diff
@@ -3,6 +3,7 @@ kind: Deployment
 metadata:
   annotations:
     config.kubernetes.io/depends-on: apps/namespaces/default/StatefulSet/wordpress-mysql
+    internal.kpt.dev/upstream-identifier: apps|Deployment|example|wordpress
```
mengqiy, line 6: When will this annotation be set? When we `kpt pkg get` a package?

examples/set-namespace-kpt-package-context/Kptfile
```diff
@@ -2,6 +2,8 @@ apiVersion: kpt.dev/v1
 kind: Kptfile
 metadata:
   name: example
+  annotations:
+    config.kubernetes.io/local-config: "true"
```
mengqiy, line 6: Why we need to add this?

functions/go/set-namespace/go.mod
```diff
@@ -2,7 +2,11 @@ module github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-na
 
 go 1.17
 
-require github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220506190241-f85503febd54
+require (
+	github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220706221933-7181f451a663
+	k8s.io/api v0.24.2
+	k8s.io/apimachinery v0.24.2
```
mengqiy, line 8: We should try to use the same version as porch. Otherwise, it can cause problems when we try to import it in Porch.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,311 @@ import (
...
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func SetNamespace(rl *fn.ResourceList) (bool, error) {
-	tc := NamespaceTransformer{}
+// Run provides the main workflow to update the ResourceList.Items "namespace" value.
+func Run(rl *fn.ResourceList) (bool, error) {
```
mengqiy, line 28: Why changing the name?

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,311 @@ import (
 	"strings"
 
 	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	"k8s.io/apimachinery/pkg/util/sets"
+
```
mengqiy, line 22: nit: new line is not necessary. If you want to separate github imports from k8s imports, the newline should be after "github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,311 @@ import (
...
-		oldNss = append(oldNss, `"`+oldNs+`"`)
-	}
-	msg := fmt.Sprintf("namespace %v updated to %q, %d value(s) changed", strings.Join(oldNss, ","), p.NewNamespace, count)
-	return fn.GeneralResult(msg, fn.Info)
+// SetNamespace defines structs to parse KRM resource "SetNamespace" (the custom function config) and "ConfigMap" data.
+// it provides the method "Config" to read the function configs from ResourceList.FunctionConfig
+// it provides the method "Transform" to change the "namespace" and update the "config.kubernetes.io/depends-on" annotation.
+type SetNamespace struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
```
mengqiy, line 47: protobuf tag is not useful here.
Instead of using `metav1`, we can use https://github.com/kubernetes-sigs/kustomize/blob/b727febd08b3902b959e48e2f2649c77a3ed911d/kyaml/yaml/types.go#L121-L152

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,311 @@ import (
...
+	NewNamespace      string `json:"namespace,omitempty"`
+	NamespaceMatcher  string `json:"namespaceMatcher,omitempty"`
```
mengqiy, line 49: Should we consider adding yaml tag as well?

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,311 @@ import (
...
+		return fmt.Errorf("`data.namespace` should not be empty")
+	case o.IsGVK(fnConfigGroup, fnConfigVersion, fnConfigKind):
+		o.AsOrDie(&p)
 		if p.NewNamespace == "" {
 			return fmt.Errorf("`namespace` should not be empty")
 		}
 	default:
-		return fmt.Errorf("unknown functionConfig Kind=%v ApiVersion=%v, expect `%v` or `ConfigMap`",
-			o.GetKind(), o.GetAPIVersion(), fnConfigKind)
+		return fmt.Errorf("unknown functionConfig Kind=%v ApiVersion=%v, expect `ConfigMap(v1)` or `%s(%s/%s)`",
```
mengqiy, line 83: `Kind(group/version)` doesn't seem to be a common way to format GVK. The formatting used in kubectl is `kind.version.group`. e.g. you can run `kubectl get deployments.v1.apps`.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,311 @@ import (
...
-// GetDependsOnMap iterates `objects` to get the annotation which contains namespace value.
-func GetDependsOnMap(objects []*fn.KubeObject) map[string]bool {
-	dependsOnMap := map[string]bool{}
-	VisitNamespaces(objects, func(ns *Namespace) {
-		key := ns.GetDependsOnAnnotation()
-		dependsOnMap[key] = true
-	})
-	return dependsOnMap
+// VisitAll applies "visitor" function to both namespace scoped and cluster scoped resource.
+func VisitAll(objects fn.KubeObjects, visitor func(origin string, currentPtr *string, idStr ...string)) {
```
mengqiy, line 204: This might be something useful in the SDK after generalization.

functions/go/set-namespace/transformer/namespace.go
```diff
@@ -18,206 +18,311 @@ import (
...
 		}
-		return dependsOnKeyPattern(group, obj.GetKind(), obj.GetName())
-	}
-	return &Namespace{
-		Ptr:                 namespacePtr, //  obj.GetStringOrDie(path...),
-		IsNamespace:         obj.IsGVK("v1", "Namespace"),
-		DependsOnAnnotation: obj.GetAnnotations()[dependsOnAnnotation],
-		annotationGetter:    annotationGetter,
-		annotationSetter:    annotationSetter,
+		key := fmt.Sprintf("%v/namespaces/%v/%v/%v", id.Group, id.Namespace, id.Kind, id.Name)
```
mengqiy, line 266: It might be useful to have a helper to format a key string form an `id` object.
I'm curious why use choose this format as the key.

tests/set-namespace/replace-match/Kptfile
```diff
@@ -0,0 +1,10 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: example
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/set-namespace:unstable
+      configMap:
+        namespace: new-ns
+        namespaceMatcher: example2
```
mengqiy, line 10: Why this is a test instead of an example?

mengqiy:
Please update the usage doc.

examples/set-namespace-depends-on/resources.yaml
yuwenma: kpt CLI will add this annotation in `kpt pkg get ` and `kpt pkg udpate`. porch will add this annotation in clone and pkg update.

examples/set-namespace-kpt-package-context/Kptfile
yuwenma: This is a cleanup. Currently kpt adds this annotation to Kptfile in creation time. e.g. `kpt pkg init` gives 
```yaml
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: tm
  annotations:
info:
  description: sample description
```

functions/go/set-namespace/go.mod
yuwenma: This is really a good point! Do you have any ideas how we can guarantee it automatically? This requirement makes sense to me, but it is hard to be sustainable.

yuwenma: I manually updated this for now.

functions/go/set-namespace/transformer/namespace.go
yuwenma: Thanks for asking. 😄  I want to use the `SetNamespace` as the transformer name which is also the function config kind. So the transformer is served as the typed functionConfig parser at the same time. The advantage is that the `As` method can be called in a single line (well I believe you must already use this or something similar) 
```go
func (t *SetNamespace) Config(fnConfig *fn.KubeObject) {
     fnConfig.As(t)
}

File: functions/go/set-namespace/transformer/namespace.go (hunk 13 above)
Comment (yuwenma) [reply functions/go/set-namespace/transformer/namespace.go:83]: ah, like the reversed domain. Thanks for the suggestion. updated.

File: functions/go/set-namespace/transformer/namespace.go (hunk 15 above)
Comment (yuwenma) [reply functions/go/set-namespace/transformer/namespace.go:266]: ..same here..... it's a little bit error-prone .

File: functions/go/set-namespace/transformer/namespace.go (hunk 14 above)
Comment (yuwenma) [reply functions/go/set-namespace/transformer/namespace.go:204]: Totally agree!

File: examples/set-namespace-depends-on/.expected/diff.patch (hunk 5 above)
Comment (yuwenma) [reply examples/set-namespace-depends-on/.expected/diff.patch:42]: you are absolutely right. updated.

File: tests/set-namespace/replace-match/Kptfile (hunk 16 above)
Comment (yuwenma) [reply tests/set-namespace/replace-match/Kptfile:10]: yeah, example fits better here. I'd prefer updating the tests/examples and docs in a separate PR. It's more efficient to think them together.

File: functions/go/set-namespace/transformer/namespace.go (hunk 12 above)
Comment (yuwenma) [reply functions/go/set-namespace/transformer/namespace.go:49]: mind explaining what the benefit is? `preserving comment` seems not to be a concern since this is function config.

File: functions/go/set-namespace/transformer/namespace.go (hunk 11 above)
Comment (yuwenma) [reply functions/go/set-namespace/transformer/namespace.go:47]: I'm trying to not import kyaml for now. Or we can define similar things on the sdk.

Comment (mengqiy) [reply functions/go/set-namespace/transformer/namespace.go:47]: `sdk/go/fn` depends on `kyaml`. I guess you mean you don't want to depend on `kyaml` explicitly.
One thing we can do here is to define types in https://github.com/kubernetes-sigs/kustomize/blob/b727febd08b3902b959e48e2f2649c77a3ed911d/kyaml/yaml/types.go#L121-L152 as alias in `sdk/go/fn`.

Types in `metav1` only has json tags but not yaml tags.

File: functions/go/set-namespace/transformer/namespace.go (hunk 12 above)
Comment (mengqiy) [reply functions/go/set-namespace/transformer/namespace.go:49]: If a user depends on this functionConfig types (e.g. he/she is writing a generator function that generates this functionConfig) and he/she uses yaml serialization, this user may run into issues.

Adding yaml doesn't cost us anything. Why don't we add them?

File: tests/set-namespace/replace-match/Kptfile (hunk 16 above)
Comment (mengqiy) [reply tests/set-namespace/replace-match/Kptfile:10]: > I'd prefer updating the tests/examples and docs in a separate PR.

This seems to be a completely **new** test. I don't see the benefits of adding it as test first and then changing it to an example. For new example or test, I'd suggest to create it in the right place in the beginning.

File: tests/set-namespace/no-fnconfig-resource/.expected/results.yaml [hunk 17]
```diff
@@ -0,0 +1,14 @@
...
+exitCode: 0
+items:
+  - image: gcr.io/kpt-fn/set-namespace:unstable
+    exitCode: 0
+    results:
+      - message: FunctionConfig is missing. Expect `ConfigMap.v1` or `SetNamespace.v1alpha1.fn.kpt.dev`
+        severity: error
+        resourceRef: {}
+        file:
+          index: -1
```
mengqiy, line 14: Hmm... What does `-1` mean?

functions/go/set-namespace/transformer/namespace.go
yuwenma: missed this comments before (GitHub folds comments). 😆  
Done

functions/go/set-namespace/transformer/namespace.go
yuwenma: Yea, I want to simplify the go dependencies as much as possible, otherwise users can easily encounter dependency version conflict. I think it's better to not depend on kyaml directly in functions, but either provide the needed struct in SDK (like you suggested ) or even add the struct inline.  I added the structure in this function with a TODO to let go/fn provides the meta types eventually.

tests/set-namespace/no-fnconfig-resource/.expected/results.yaml
yuwenma: This is intended. file index -1 meaning the file path does not exist.  This test covers the cases where functionConfig is required but not given

functions/go/set-namespace/transformer/namespace.go
yuwenma: Great. Changing this removes another dependency metav1.
I also realized another potential risk condition, where that users may generate yaml serialized ConfigMap (as you mentioned in the other comment), does it mean if this ConfigMap cannot be parsed as below since the metav1 does not have yaml tag? Or did I miss anything? 
```yaml
var cm metav1.ConfigMap
fnConfigObject.As(&cm)
```

If this is a concern, then I guess we need to provide a yaml tag handler in the go/fn.

mengqiy: `As` function in `go/fn` has considered it. It relies on json tags to deserialize yaml.
If a user tries to deserialize yaml using `go-yaml` or `kyaml` directly (e.g. deserialize yaml bytes to metav1.ConfigMap), he/she will run into issues.

When we define types in our projects, we should always try to add both json and yaml tags.

tests/set-namespace/no-fnconfig-resource/.expected/results.yaml
mengqiy: Quoted from the functions spec: https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md

> index:
            type: number
            default: 0
            description: Index of the object in a multi-object YAML file.

The function spec has not define what does `-1` mean. `kpt` has not either.

> functionConfig is required but not given

IMO it's too hacky to use `-1` to communicate this. We can leave the `index` field empty and use the `message` field to communicate with the user.

mengqiy: I don't think using `-1` is the right thing to do here. But I'm OK if you want to fix this in a followup PR.

mengqiy (approved):
Please let me know what you want to do with the `-1` issue. That doesn't have to be a blocking issue.

functions/go/set-namespace/transformer/namespace.go
yuwenma: Wonderful!  Didn't pay close attention to json/yaml go libraries before.

tests/set-namespace/no-fnconfig-resource/.expected/results.yaml
yuwenma: Do you know what index value kpt uses to mark resources from generator?

yuwenma:
> Please let me know what you want to do with the `-1` issue. That doesn't have to be a blocking issue.

I'll leave the -1 as it is for now. Proposing a -1 value is straight forward. Do you know the progress to propose that in kpt?

PR #886, opened 2022-07-13
Add `set-labels` e2e tests and readme

zyy98:
add more e2e tests in tests/ and examples/; change readme to get rid of additional label fields


yuwenma:
Can we separate the update license into a different PR?

tests/set-labels/empty-resource/.expected/results.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kpt.dev/v1
+kind: FunctionResultList
+metadata:
+  name: fnresults
+exitCode: 0
+items:
+  - image: gcr.io/kpt-fn/set-labels:unstable
+    exitCode: 0
```
yuwenma, line 8: For empty resource, I would expect the function can still give a helpful user message, something like no resources found,  or updated 0 labels. So as users can take actions. 
Note: it's hard for users to debug an unexpected behavior of a KRM function, so it's more critical to give user messages.

tests/set-labels/wrong-config-group/fn-config.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: fn.kpt.dev/v1alpha1
+kind: SetLabel
```
yuwenma, line 2: super nit: can we change the `SetLabel` to `BadKind`? the plural `SetLabels` is less obvious 🤣

yuwenma:
Can we also add cases to cover:
- Empty fnConfig will give helpful user messages.
- Add some happy path to examples/, which covers the special cases like 
   - https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/set-labels/transformer/common.go#L153
   -  https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/set-labels/transformer/common.go#L148
   - https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/set-labels/transformer/labels.go#L111

examples/set-labels-full-coverage/README.md
```diff
@@ -34,25 +34,29 @@ apiVersion: fn.kpt.dev/v1alpha1
...
+$ kpt fn render set-labels-full-coverage
 ```
 
 ### Expected result
 
-Check all resources have 2 labels: `color: orange` and `fruit: apple`. 
-The `CronJob` should have new labels in `spec.jobTemplate.metadata.labels` and `"spec.jobTemplate.spec.template.metadata.labels"` as well.
+Check all resources, if the labels to the path should be upsert, 
+then we can see 3 labels have been updated, `color: orange`, `fruit: apple` and `app: new`. 
+If the path does not need to create, just update labels if the key already exists, otherwise, ignore.
```
yuwenma, line 60: How about "The following labels should be upserted to the `labels`, `matchingLabels` or `seletors`(labelSelector) fields: ..." .

Reasons: try to avoid describing the code logic, but explaining from the usage perspective.

examples/set-labels-full-coverage/fn-config.yaml
```diff
@@ -7,3 +7,4 @@ metadata:
 labels:
   color: orange
   fruit: apple
+  app: new
```
yuwenma, line 10: qq: any reason to add another field here?

functions/go/set-labels/transformer/labels.go
```diff
@@ -205,7 +205,7 @@ func checkExistAndSet(o *fn.SubObject, labelPath FieldPath, newLabels map[string
 				return err
 			}
 			if value == val || !exist {
-				return nil
+				continue
```
yuwenma, line 208: good catch

yuwenma (approved):
Looks great. One minor comment about README wording.

examples/set-labels-full-coverage/README.md
```diff
@@ -1,16 +1,16 @@
-# set-labels: Advanced Example
+# set-labels: Full Coverage Example
```
mengqiy, line 1: What do we want to do in this example? Include all built-in resources?

examples/set-labels-full-coverage/README.md
```diff
@@ -1,16 +1,16 @@
...
 
 ### Overview
 
 This example demonstrates how to declaratively run [`set-labels`] function
-to upsert labels to the `.metadata.labels` field on all resources.
+to upsert all common labels to different type of resources, as well as self-defined resource.
```
mengqiy, line 6: > different type of resources, as well as self-defined resource.

Maybe `different built-in resources and CustomResourceDefinition (CRD) resources`.

functions/go/set-labels/transformer/labels_test.go
```diff
@@ -45,7 +45,7 @@ spec:
 `
 
 	sliceInput := `
-apiVersion: apps/
```
mengqiy, line 48: The apiVersion for StatefulSet should be `apps/v1`.

functions/go/set-labels/transformer/labels_test.go
```diff
@@ -54,6 +54,11 @@ spec:
...
         labels:
           testkey: testvalue
+          app: testvalue
+          env: testvalue
+          quotedBoolean: testvalue
+          quotedFruit: testvalue
+          unquotedBoolean: testvalue
 `
 
 	sameLabelInput := `
```
mengqiy, line 64: The apiVersion for ConfigMap should be `v1`.
Please audit the file ensure all occurrences have been updated.

tests/set-labels/wrong-resource-type/.expected/config.yaml
```diff
@@ -0,0 +1 @@
+exitCode: 1
```
mengqiy, line 1: > tests/set-labels/wrong-resource-type

Maybe `ill-formatted-fn-config` is a better here.

tests/set-labels/empty-resource/Kptfile
```diff
@@ -0,0 +1,10 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: example
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/set-labels:unstable
+      configPath: fn-config.yaml
+      selectors:
```
mengqiy, line 9: You don't have to specify the selectors here, right?

tests/set-labels/wrong-config-data-type/Kptfile
```diff
@@ -7,5 +7,4 @@ pipeline:
     - image: gcr.io/kpt-fn/set-labels:unstable
       configPath: fn-config.yaml
       selectors:
```
mengqiy, line 9: Same here and below.

examples/set-labels-full-coverage/README.md
zyy98: Yes, this covers all the resources in `CommonSpecs`

tests/set-labels/empty-resource/Kptfile
zyy98: Yes, I am waiting for local-config PR to merge first, then I will make some changes regarding those tests

functions/go/set-labels/transformer/labels_test.go
```diff
@@ -54,6 +54,11 @@ spec:
     - metadata:
         labels:
           testkey: testvalue
+          app: testvalue
+          env: testvalue
+          quotedBoolean: testvalue
+          quotedFruit: testvalue
+          unquotedBoolean: testvalue
```
yuwenma, line 61: super nit: just a general suggestion that it's better to use different "testvalue" for each field.

yuwenma (changes requested):

functions/go/set-labels/transformer/labels.go
```diff
@@ -105,18 +108,6 @@ func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
...
-			if o.GetMap("spec") != nil {
-				for _, vctObj := range o.GetMap("spec").GetSlice("volumeClaimTemplates") {
-					updatedLabels, err = updateLabels(vctObj, metaLabelsPath, p.NewLabels, true)
-					if err != nil {
-						return err
-					}
-					p.LogResult(o, FieldPath{"spec", "volumeClaimTemplates[]", "metadata", "labels"}, updatedLabels)
-				}
-			}
-		}
```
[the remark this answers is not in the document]
zyy98: Because if we handle special cases one by one, there would be too many cases to hardcode. So I just add [] in commonFieldSpec to indicate lists

functions/go/set-labels/transformer/labels.go
```diff
@@ -92,28 +92,162 @@ func (p *LabelTransformer) setLabelsInSpecs(o *fn.KubeObject) error {
...
+
+func (p *LabelTransformer) setMetadataForAll(o *fn.KubeObject) error {
+	metaLabelsPath := FieldPath{"metadata", "labels"}
+	updatedLabels, err := updateLabels(&o.SubObject, metaLabelsPath, p.NewLabels, true)
+	if err != nil {
+		return err
+	}
+	p.LogResult(o, metaLabelsPath, updatedLabels)
+	return nil
+}
```
yuwenma, line 123: Suggest using the `o.SetLabel` or `o.SetLabels` functions directly.

functions/go/set-labels/transformer/labels.go
```diff
@@ -92,28 +92,162 @@ func (p *LabelTransformer) setLabelsInSpecs(o *fn.KubeObject) error {
...
+			if err := p.processPodSpec(podSpecObj, o); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// processPodSpec takes in podSpec object and parse its path, parent kubeObject is also passed in for logging
+func (p *LabelTransformer) processPodSpec(o *fn.SubObject, parentO *fn.KubeObject) error {
```
yuwenma, line 158: I tried this function in`set-default-labels` I still think it is too tedious to log every resource changes. Can we remove the resource level `LogResult`?

functions/go/set-labels/transformer/labels.go
zyy98: I am not sure what you mean here. Do you mean to change the name? `o.SetLabels` is the function that performs all label updates according to `resourceList`. Here I am separating all those updates into 3 parts, and this is part to `handle metadata.labels`.

functions/go/set-labels/transformer/labels.go
zyy98: By removing the resource level, do you mean only want a general count of how many resources have been updated in total?

yuwenma: yeah. `kpt fn` truncates the log if the output is too large. This is bad for PASS function because 1. normally the last log message contains the summary 2. truncated log is more used for FAIL function.  It seems that the set-labels can easily exceed the limitations.

functions/go/set-labels/transformer/labels.go
```diff
@@ -92,6 +92,11 @@ func (p *LabelTransformer) setLabelsInSpecs(o *fn.KubeObject) error {
 
 // Transform updates the labels in the right path using GVK filter and other configurable fields
 func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
+	if objects.Len() == 0 || objects[0] == nil {
+		newResult := fn.Result{Message: "no resources found, nothing to transform"}
```
yuwenma, line 96: Suggest using `GeneralResult` https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/fn/result.go#L52

yuwenma: message -> "no input resources"

examples/set-labels-full-coverage/.expected/diff.patch
```diff
@@ -0,0 +1,265 @@
+diff --git a/Kptfile b/Kptfile
+index 2ab6a18..cb09366 100644
+--- a/Kptfile
++++ b/Kptfile
+@@ -8,4 +8,3 @@ pipeline:
+   mutators:
+     - image: gcr.io/kpt-fn/set-labels:unstable
+       configPath: fn-config.yaml
+-
```
mengqiy, line 9: This diff is meaningless, you can tweak the input to make it disappear.

mengqiy (approved):
One nit, LGTM otherwise.

PR #887, opened 2022-07-16
update make rule

zyy98:


yuwenma:
Looks good. Can we add a rule in the root ./Makefile so users do not need to redirect to functions/go to run the update license rule?

Makefile
```diff
@@ -74,3 +74,6 @@ site-check: ## Test site for broken catalog links.
 update-function-docs: ## Update documentation for a function release branch
 	(cd scripts/update_function_docs/ && go build -o update_function_docs .)
 	RELEASE_BRANCH=$(RELEASE_BRANCH) ./scripts/update_function_docs/update_function_docs
+
+license:
+	cd functions/go && $(MAKE) license
```
yuwenma, line 79: Have you tried this? running `make license` under functions/go works well, but it seems that this rule does not actually update the license.

mengqiy: It seems you only add it for golang part. Root level makefile should be able to handle all golang/ts functions in curated (functions/) and contrib (contrib/functions/).

functions/go/Makefile
```diff
@@ -153,3 +153,5 @@ func-push:
 docs-generate:
 	$(GOBIN)/mdtogo $(CURRENT_FUNCTION) $(CURRENT_FUNCTION)/generated --license=none --strategy=cmdDocs
 
+license:
+	GOBIN=$(GOBIN) ../hack/update-license.sh
```
mengqiy, line 157: Please take a look at https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/Makefile#L141-L143

functions/go/Makefile
```diff
@@ -54,7 +54,7 @@ FUNCTION_BUILDS := $(patsubst %,%-BUILD,$(FUNCTIONS))
 # Targets to push images
 FUNCTION_PUSH := $(patsubst %,%-PUSH,$(FUNCTIONS))
 # Current function name used by individual function targets
-CURRENT_FUNCTION ?= unknown
+CURRENT_FUNCTION = set-labels
```
yuwenma, line 57: cleanup?

hack/update-license.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# don't add licenses to the site directory, it will break the docs
+# and will add them to the theme which is a submodule (bad)
+command -v addlicense || go install github.com/google/addlicense@v1.0.0
```
mengqiy, line 19: Can you please reconcile it with https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/scripts/install-go-licenses.sh?

functions/go/Makefile
```diff
@@ -153,3 +153,6 @@ func-push:
 docs-generate:
 	$(GOBIN)/mdtogo $(CURRENT_FUNCTION) $(CURRENT_FUNCTION)/generated --license=none --strategy=cmdDocs
 
+license:
```
mengqiy, line 156: Please reconcile with https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/Makefile#L108-L114

zyy98: I took another look, the `go-licenses` is more about analyzes a binary/package's dependencies and save those licenses to a specific directory, instead of add licenses for files, so it seems not the package we need.

yuwenma: @zyy98 I think using more dedicated PR description and subject can help the confusion here.
@mengqiy We want to add a handy tool for ourselves that can add missing COPYRIGHT to each new files. This is not to verify the dependent libraries LICENSE.

yuwenma:
@mengqiy Do you mind taking another review on this? Everything looks good on my side and I also replied to some of your comments.

functions/go/Makefile
mengqiy, line 156: I'd suggest moving it closer to check-license: https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/functions/go/Makefile#L102
Same comments for all Makefiles.

Makefile
```diff
@@ -74,3 +74,9 @@ site-check: ## Test site for broken catalog links.
 update-function-docs: ## Update documentation for a function release branch
 	(cd scripts/update_function_docs/ && go build -o update_function_docs .)
 	RELEASE_BRANCH=$(RELEASE_BRANCH) ./scripts/update_function_docs/update_function_docs
+
+license:
```
mengqiy, line 78: Same here.

mengqiy (approved):

PR #889, opened 2022-07-18
handle local config

zyy98:


functions/go/set-labels/transformer/labels.go
```diff
@@ -93,6 +93,11 @@ func (p *LabelTransformer) setLabelsInSpecs(o *fn.KubeObject) error {
 // Transform updates the labels in the right path using GVK filter and other configurable fields
 func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
 	for _, o := range objects {
+		// check if local config
+		localConfigPath := FieldPath{"metadata", "annotations", "config.kubernetes.io/local-config"}
```
yuwenma, line 97: can we use [`IsLocalConfig`](https://pkg.go.dev/github.com/GoogleContainerTools/kpt-functions-sdk/go/fn#KubeObject.IsLocalConfig) and [`WhereNot`](https://pkg.go.dev/github.com/GoogleContainerTools/kpt-functions-sdk/go/fn#KubeObjects.WhereNot)

zyy98: `IsLocalConfig` is a method, and `WhereNot` asks for a function input. Do I need to create an extra function that converts method to function?

functions/go/set-labels/transformer/labels.go
```diff
@@ -90,8 +90,15 @@ func (p *LabelTransformer) setLabelsInSpecs(o *fn.KubeObject) error {
...
+// IsLocalConfig checks if the object annotations is local config or not
+func IsLocalConfig(o *fn.KubeObject) bool {
+	return o.IsLocalConfig()
+}
+
 // Transform updates the labels in the right path using GVK filter and other configurable fields
 func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
+	// check if local config
+	objects = objects.WhereNot(IsLocalConfig)
 	for _, o := range objects {
```
yuwenma, line 102: Just to further simplify the code, we can making this into one-line as:
```go
for _, object := range objects.WhereNot(func (o *fn.KubeObject) bool{ return o.IsLocalConfig()})  
```

mengqiy (approved):

functions/go/set-labels/transformer/labels.go
```diff
@@ -92,7 +92,8 @@ func (p *LabelTransformer) setLabelsInSpecs(o *fn.KubeObject) error {
 
 // Transform updates the labels in the right path using GVK filter and other configurable fields
 func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
-	for _, o := range objects {
+	// check if local config
```
yuwenma, line 95: super nit: can you remove this comment or change it to something like "local config is ...., so it should not have labels added"

yuwenma (approved):
One minor comment. Approved

mengqiy (approved):

PR #892, opened 2022-07-26
`set-labels` feature change, handle slice, change log

zyy98:
fixed previous way of handling slice, add more path with slice and refactor that part of code after discussion with Yuwen.
change log to show general information to avoid being truncated.


functions/go/set-labels/transformer/labels.go
```diff
@@ -94,26 +101,147 @@ func (p *LabelTransformer) setLabelsInSpecs(o *fn.KubeObject) error {
 func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
```
yuwenma, line 101: func Transform() {
     // filter unrelated resources. like localconfig or Kptfile metaconfig
     err := func() {
       // set meta labels
       UpdateLabelsInMeta
       // set pod labels 
       UpdatePodLabels 
       // labelSelector
       ...
       // optional (Not CreateifNot exist)  
     }()
     if err != nil {
         LogError
     }
     LogSummary
}

functions/go/set-labels/transformer/labels.go
```diff
@@ -76,95 +84,361 @@ func (p *LabelTransformer) Config(functionConfig *fn.KubeObject) error {
...
+		return true, true
+	}
+	if o.IsGVK("batch", "", "Job") || o.IsGVK("policy", "", "PodDisruptionBudget") {
+		return true, false
+	}
+	return false, false
+}
+
+// hasPodTemplateSpec check if the resource contains struct PodTemplateSpec, ReplicationController, Deployment, ReplicaSet, DaemonSet, StatefulSet, Job kind has it
+func hasPodTemplateSpec(o *fn.KubeObject) bool {
```
yuwenma, line 257: super nit: can we use the switch/case format?

yuwenma:
The new format is much better! I think it's a good starting point to help us improving the SDK and make the `set-labels` code shorter and more readable.
I'd like to get another review from @natasha41575

functions/go/set-labels/transformer/labels.go
```diff
@@ -1,45 +1,35 @@
...
+	// ResultCount logs the total count of each labels change
+	ResultCount map[string]int
+}
+
+// NewTransformer is the constructor for labelTransformer
+func NewTransformer() *LabelTransformer {
+	transformer := LabelTransformer{}
+	resultCount := make(map[string]int)
+	transformer.ResultCount = resultCount
+	return &transformer
```
natasha41575, line 27: nit: this seems a bit verbose, suggest the following instead:

```
	resultCount := make(map[string]int)
	return &LabelTransformer {
		ResultCount: resultCount,
	}
```

functions/go/set-labels/transformer/labels.go
```diff
@@ -1,45 +1,35 @@
...
-	// FieldSpecs stores default label fields
-	FieldSpecs []FieldSpec
 	// Results logs the changes to the KRM resource labels
 	Results fn.Results
+	// ResultCount logs the total count of each labels change
+	ResultCount map[string]int
+}
+
+// NewTransformer is the constructor for labelTransformer
+func NewTransformer() *LabelTransformer {
```
natasha41575, line 23:
```suggestion
func NewLabelTransformer() *LabelTransformer {
```

functions/go/set-labels/transformer/labels.go
```diff
@@ -99,78 +75,365 @@ func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
...
+		err := func() error {
+			// set meta labels
+			if err := p.setObjectMeta(o); err != nil {
+				return err
+			}
+			// set selector
+			if err := p.setSelector(o); err != nil {
+				return err
+			}
+			// set PodTemplateSpec
```
natasha41575, line 87: nit: I think these comments (L79, L83, and L87) are not necessary, the name of the methods are descriptive enough

functions/go/set-labels/transformer/labels.go
```diff
@@ -99,78 +75,365 @@ func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
...
+}
+
+// hasJobTemplateSpec check if the object is CronJob, this kind might have JobTemplateSpec, which contains label fieldPath
+func hasJobTemplateSpec(o *fn.KubeObject) bool {
+	return o.IsGVK("batch", "", "CronJob")
+}
+
+// setJobTemplateSpecMeta set MetaObject in struct JobTemplateSpec
+func (p *LabelTransformer) setJobTemplateSpecMeta(o *fn.KubeObject) error {
+	if hasJobTemplateSpec(o) {
```
natasha41575, line 118: I think having this check inside the `setJobTemplateSpecMeta` is a little bit confusing, the name of the method `setJobTemplateSpecMeta` seems to imply that we already know that the object is a Job. 

I think a better structure would be to change lines 90-99 above to something like

```
if hasJobTempalteSpec(o) {
  p.setJobTemplateSpecMeta
}
```

and same for `p.setVolumeClaimTemplates(o)` and `p.setNetworkPolicyRule(o)`

functions/go/set-labels/transformer/labels.go
```diff
@@ -99,78 +75,365 @@ func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
...
+		return true
+	case o.IsGVK("", "", "Deployment"):
+		return true
+	case o.IsGVK("", "", "ReplicaSet"):
+		return true
+	case o.IsGVK("", "", "DaemonSet"):
+		return true
+	case o.IsGVK("apps", "", "StatefulSet"):
+		return true
+	case o.IsGVK("batch", "", "Job"):
```
natasha41575, line 257: Discussion point for @yuwenma, I think this would be a good use case for getting this information from the openapi data instead of having to hardcode all of these things.

functions/go/set-labels/transformer/labels.go
zyy98: I see. I left this change because for functions like `setSelector`, I need to check if it has different selector type inside the function. So for consistency I just left all `hasXXX` function inside `setXXX`. But for the corner cases, it seems a good idea to separate them out.

functions/go/set-labels/transformer/labels.go
```diff
@@ -99,78 +75,365 @@ func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
...
+		return
+	}
+	keys := make([]string, 0)
+	for k := range labelCount {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for i := 0; i < len(keys); i++ {
+		key := keys[i]
+		element := labelCount[key]
```
natasha41575, line 397: nit: suggest renaming 'element' to 'labelCountValue' or something similar

functions/go/set-labels/transformer/labels.go
```diff
@@ -99,78 +75,365 @@ func (p *LabelTransformer) Transform(objects fn.KubeObjects) error {
...
+		msg := fmt.Sprintf("set labels {%v} for %v times", key, element)
```
natasha41575, line 398:
```suggestion
		msg := fmt.Sprintf("set labels {%v} %v times", key, element)
```

natasha41575 (approved):
Code structure LGTM, defer to @yuwenma for final approval

yuwenma (approved):

yuwenma:
This PR looks great! This is a good starting point to let us continue improving the SDK and adding features. like the OpenApi schema as @natasha41575 suggested.

PR #898, opened 2022-08-05
support compiling set-ns and set-labels to wasm

mengqiy:
How to compile:
```
$ cd functions/go/set-namespace
$ GOOS=js GOARCH=wasm go build -o set-namespace.wasm 
```

### DIFF
--- a/functions/go/set-labels/run.go
+++ b/functions/go/set-labels/run.go
@@ -12,15 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !(js && wasm)
+
 package main
 
 import (
-	"os"
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-labels/transformer"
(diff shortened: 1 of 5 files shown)


yuwenma:
So cool! Is this for experimental purpose or you literally want `set-namespace` to be the pilot function for WASM? If it's the former, quick approval on my side. But please be aware that the latest set-namespace code is not released yet and already contains major changes like supporting upstream identifier.
 For the latter, I feel like the WASM `run` function may be better live in the Go SDK (But I'm totally fine if you want to have it land in this function). Just a note that before releasing `set-namespace`, we need to disable the warning message [here](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/d955e7d3a80086e7a400faeea53174e08faed314/functions/go/set-namespace/transformer/namespace.go#L123).

yuwenma (approved):
Approved on my side. Feel free to merge or update whatever it needs.

natasha41575 (approved):

PR #902, opened 2022-08-11
Remove the temporary fn sdk in catalog

mengqiy:
All functions that used to depend on it have been migrated to the new fn SDK.


yuwenma (approved):
what a lovely PR. Thanks for this cleanup

PR #904, opened 2022-08-12
refactor `set-image` with go sdk

zyy98:
To fix the issues https://github.com/GoogleContainerTools/kpt/issues/3444
Start by refactoring `set-image` functions using go sdk package.
Allow both `digest` and `tag` to be as input.
Replace unit tests with end-to-end tests.


functions/go/set-image/image_transformer_test.go
```diff
@@ -30,15 +30,16 @@ func runImageTransformerResults(input, config string) (*fn.ResourceList, error)
...
 	rl.FunctionConfig = functionConfig
 	in, _ := rl.ToYAML()
-	out, err := fn.Run(fn.ResourceListProcessorFunc(setImageTags), in)
+	out, err := fn.Run(fn.ResourceListProcessorFunc(SetImageTagSDK), in)
 	if err != nil {
 		return nil, err
 	}
 	rl, _ = fn.ParseResourceList(out)
 
+	rl.Results = nil
```
zyy98, line 39: The original method keeps an internal struct `setImageResults` to store all the results, then its `LogResult` method convert the internal struct to be `fn.Results` when the function finishes. For me I just directly append the results while updating the image, as I did in `set-labels`. So unit test differs here, as I reset my Results to nil.

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+				Severity: fn.Warning,
+			}
+			return true, warning
+		}
+	}
+	return true, nil
+}
+
+// updateImages the update process for each image, if error happened, it panics
+func (imageTrans *ImageTransformer) updateImages(o *fn.SubObject, newImage *Image, parentO *fn.KubeObject, fieldPath FieldPath) {
```
zyy98, line 197: Because I am operating upon `SubObject`, and I also need information for logging, so I have its parent KubeObject and its relative path passed in. This makes me think of the error message PR in sdk, as I try to add those information in SubObject error to help debugging. The same information is needed here again, I wonder if those fields might be part of `SubObject` some day.

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+	oldValue := o.NestedStringOrDie("image")
+	container := o.NestedStringOrDie("name")
+
+	matched, err := matchImage(oldValue, container, newImage)
+	if matched {
+		newName := getNewImageName(oldValue, newImage)
+		o.SetNestedStringOrDie(newName, "image")
+		fieldPath = append(fieldPath, "image")
+		imageTrans.LogResult(parentO, oldValue, newName, fieldPath)
+		imageTrans.ResultCount += 1
```
zyy98, line 207: I keep a total count, but did not display it. Because I hope to be like its original logging format, so I did not add the total logging for now. And also because we can only update one image per operation, so I wonder keeping a total count is needed or not

functions/go/set-image/image_transformer_test.go
```diff
@@ -30,15 +30,16 @@ func runImageTransformerResults(input, config string) (*fn.ResourceList, error)
 	}
 	rl.FunctionConfig = functionConfig
 	in, _ := rl.ToYAML()
-	out, err := fn.Run(fn.ResourceListProcessorFunc(setImageTags), in)
+	out, err := fn.Run(fn.ResourceListProcessorFunc(SetImageTagSDK), in)
```
yuwenma, line 33: Using "SDK" can be confusing. The `SetImageTagSDK` supposes to transform the KubeObject-based resources with "image" logic. It is not and we have no plan to use it as a SDK.

yuwenma, line 33: Suggest using the `fn.Runner` rather than `ResourceListProcessorFunc`. Here's an [example ](https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/fn/examples/example_asmain_runner_test.go)

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+	"sigs.k8s.io/kustomize/api/image"
+	"sigs.k8s.io/kustomize/api/types"
+)
+
+const FnConfigKind = "SetImage"
+
+type FieldPath []string
+
+type Image struct {
+	// Name is a tag-less image name. should be deprecate, means image name
```
yuwenma, line 17: `// (DEPRECATED): Name ...`

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+	if err := imageTrans.validateInput(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// validateInput validates the inputs passed into via the functionConfig
+func (imageTrans *ImageTransformer) validateInput() error {
+	if imageTrans.Image.Name == "" && imageTrans.Image.ContainerName == "" && imageTrans.Image.ImageName == "" {
+		return fmt.Errorf("missing image name or container name")
```
yuwenma, line 72: How about `either `imageName` or `containerName` is required`. This helps users to navigate the exact function configs.

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	if imageTrans.Image.NewName == "" && imageTrans.Image.NewTag == "" && imageTrans.Image.Digest == "" {
+		return fmt.Errorf("missing image newName, newTag, or digest")
```
yuwenma, line 75: ditto

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	if imageTrans.Image.NewTag != "" && imageTrans.Image.Digest != "" {
+		return fmt.Errorf("image newTag and digest both set")
```
yuwenma, line 78: ditto

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+			functionConfig.GetKind(), functionConfig.GetAPIVersion(), FnConfigKind)
+	}
+	if err := imageTrans.validateInput(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// validateInput validates the inputs passed into via the functionConfig
+func (imageTrans *ImageTransformer) validateInput() error {
```
yuwenma, line 70: For validation function, I feel like it is better to contain the logic about mapping the deprecating `name` to `imageName`.

functions/go/set-image/image_transformer_test.go
zyy98: I am checking out `Runner` interface, it seems the `*fn.Context` does not provide good result logging. Its related functions only provide message and severity, no further information about `fn.ResourceRef`, `fn.Field` or `fn.File`

yuwenma (approved):
Haven't checked the unittests. Two major areas to consider:
1. fn.Result seems to be redundant, can you either use existing functions (e.g. `ConfigObjectResult`) or develop something new in the sdk?
2. it seems that there are some duplicated code, e.g. `validateInput`, shall we cleanup them? FYI, to keep `additionalFields`, I'm expecting that we can completely separate the solutions using kyaml lib and the go sdk lib.  The pseudo code can be like
parse input yaml --> initialize `imageTransformer` --> transform configs from A to B --> `additionalField` detected --> transform modified config B to C --> log results.

yuwenma (changes requested):
undo approve since this PR adds some more changes.

functions/go/set-image/image_transformer_test.go
yuwenma: If so, we shall add a "ConfigObjectResult" method to `Context`

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+	imageTrans.Results = append(imageTrans.Results, warning)
+}
+
+func (imageTrans *ImageTransformer) setPodSpecContainers(o *fn.KubeObject) {
+	if spec := o.GetMap("spec"); spec != nil {
+		if template := spec.GetMap("template"); template != nil {
+			if podSpec := template.GetMap("spec"); podSpec != nil {
+				for _, vecObj := range podSpec.GetSlice("containers") {
+					fieldPath := FieldPath{"spec", "template", "spec", "containers"}
+					imageTrans.updateImages(vecObj, &imageTrans.Image, o, fieldPath)
```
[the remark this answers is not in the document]
zyy98: Because it is a slice of object, so I have to use `GetSlice` to iterate each subObject. The entire field spec only works when there is no slice involved

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+
+// validateInput validates the inputs passed into via the functionConfig
+func (imageTrans *ImageTransformer) validateInput() error {
+	if imageTrans.Image.Name == "" && imageTrans.Image.ContainerName == "" && imageTrans.Image.ImageName == "" {
+		return fmt.Errorf("either image name or container name is required")
+	}
+	if imageTrans.Image.NewName == "" && imageTrans.Image.NewTag == "" && imageTrans.Image.Digest == "" {
+		return fmt.Errorf("either image newName, newTag, or digest is required")
+	}
+	if imageTrans.Image.NewTag != "" && imageTrans.Image.Digest != "" {
```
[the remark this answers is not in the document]
zyy98: I see. I plan to do it in the next PR, will add TODO

functions/go/set-image/image_transformer_test.go
zyy98: Oh I see, I missed this important method.

zyy98: `ConfigObjectResult` does not contain `field`, and would miss some of the information. Should I add it in the function? Or it is ok to leave out those information `path`, `currentValue`, `proposedValue`.

zyy98: There are some other issues with this `fn.Runner`, I am not sure how to test it in unit tests. Because I cannot create `Context`, its field is not exported, there is no way for me to test its Results because I cannot create `Context`. In e2e test, this creation of `Context` is completed by `AsMain`.

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
...
+	// deprecated name means image name
+	if imageTrans.Image.Name != "" {
+		imageTrans.Image.ImageName = imageTrans.Image.Name
+	}
+	return nil
+}
+
+// Transform updates the image in pod
+func (imageTrans *ImageTransformer) Transform(objects fn.KubeObjects) error {
+	// using unit test and pass in empty string would provide a nil; an empty file in e2e would provide 0 object
```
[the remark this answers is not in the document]
zyy98: I am not sure what you meant by **fix**, it is just some comment explaining my code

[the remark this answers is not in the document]
yuwenma: The comment makes it looks like you are making some workaround/code changes specifically for the unittest and e2e tests to pass. Or maybe it's just the wording.

functions/go/set-image/transformer/images.go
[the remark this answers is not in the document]
yuwenma: My suggestion is to not pass fieldpath and logs the result in the caller function

functions/go/set-image/transformer/images.go
```diff
@@ -0,0 +1,237 @@
+package transformer
```
yuwenma, line 1: main.go
```go
	if err := fn.AsMain(&SetImage{}); err != nil {
```
transformers/setImage.go
```go

var _ fn.Runner = &SetImage{}

type SetImage struct {
	 // DEPRECATED
        Name string `yam: "", json:"name,omitempty"`
	NewImage string `json:"newImage,omitempty"`
	NewTag string `json:"newTag,omitempty"`
        // ONLY for kustomize
        //additionalFieldSpec is private attribute, different than other config because if importing SetImage, we do not want users to use additionalFieldSpec. 
        additionalFieldSpec ....  
}

func (r *SetLabels) Run(ctx *fn.Context, _ *fn.KubeObject, items fn.KubeObjects) {
        err := r.Validate()
        if err != nil { 
               ctx.ResultErrorOrDie()
        }
        for _, o := range items.Where(hasPodSpec){
                    err := o.replaceImage()
                    if err != nil {
                         ctx.ResultError(err.Error(), o)
                    } else {
                          ctx.ResultInfoXXX("", o)
                    }
        }
          
        for _, o := range items.Where(hasPodTemplateSpec){
               ...
        }
         if lr.additionalFieldSpec != nil && len(r.additionalFieldSpec) != 0 {
               results := custom.SetAdditionalFieldSpec
               // converts results to results
               // log results
         } 
        ctx.ResultInfo("summary")
}
```
custom/addadditionalfield
```go

```


///

functions/go/set-image/transformer/images.go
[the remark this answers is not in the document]
yuwenma: Or you can delete the comment. I think the confusion I have is that normally comment do not explain how a code can help a unit test. If it does, it usually means the code is specifically modified that way to have unittest pass.

functions/go/set-image/image_transformer_test.go
zyy98: There are more issues with `fn.Runner` than I think. Its config function is not very generalized, it can only deal with map type, such as `SetLabels` kind, but for `SetImage`, the type is Image instead of map. More debugging error details in 1on1 doc.

functions/go/set-image/custom/custom.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	"sigs.k8s.io/kustomize/api/filters/imagetag"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/filtersutil"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+func SetAdditionalFieldSpec(image types.Image, objects fn.KubeObjects, additionalImageFields types.FsSlice, ctx *fn.Context) {
```
yuwenma, line 13: nit: can you add a docstring to explain what `SetAdditionalFieldSpec` does. and the long term plans we discussed why and when additionalFieldSpec should be dropped.

functions/go/set-image/custom/custom.go
```diff
@@ -0,0 +1,43 @@
...
+	for i, obj := range objects {
+		objRN, err := yaml.Parse(obj.String())
+		if err != nil {
+			ctx.ResultInfo(err.Error(), obj)
```
yuwenma, line 17: Is there a reason why the log is "info" but not "error"?

functions/go/set-image/custom/custom.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		filter := imagetag.Filter{
+			ImageTag: image,
+			FsSlice:  additionalImageFields,
+		}
+
+		filter.WithMutationTracker(mutationTracker(ctx, obj))
+		err = filtersutil.ApplyToJSON(filter, objRN)
+		if err != nil {
+			ctx.ResultInfo(err.Error(), obj)
```
yuwenma, line 27: ditto ResultErr?

functions/go/set-image/custom/custom.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		newObj, err := fn.ParseKubeObject([]byte(objRN.MustString()))
+		if err != nil {
+			ctx.ResultInfo(err.Error(), obj)
```
yuwenma, line 31: ditto

functions/go/set-image/custom/custom.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		objects[i] = newObj
+	}
+}
+
+func mutationTracker(ctx *fn.Context, ko *fn.KubeObject) func(key, value, tag string, node *yaml.RNode) {
+	return func(key, value, tag string, node *yaml.RNode) {
+		currentValue := node.YNode().Value
+		msg := fmt.Sprintf("set image from %v to %v", currentValue, value)
```
yuwenma, line 40: nit: "update image" or "replace image", "set" is kind of vague here.

functions/go/set-image/custom/custom.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		newObj, err := fn.ParseKubeObject([]byte(objRN.MustString()))
+		if err != nil {
+			ctx.ResultInfo(err.Error(), obj)
+		}
+		objects[i] = newObj
+	}
+}
+
+func mutationTracker(ctx *fn.Context, ko *fn.KubeObject) func(key, value, tag string, node *yaml.RNode) {
```
yuwenma, line 37: Suggest renaming this function to "LogResultCallback". 
Reasons:
- The function names `mutationTracker` and `WithMutationTracker` give redundant information. 
- In general a function should serve at least one of the following goals:
   - explain what the code does. `CreateXX` `ResetXX`
   - follow some name conventions. e.g. `Request` `Response` `Get` in a http library.
  the goal is to give reviewers an idea how closely they need to dig into the code.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-image/custom"
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	"sigs.k8s.io/kustomize/api/image"
+	"sigs.k8s.io/kustomize/api/types"
+)
+
+// SetImage supports the set-image workflow, it uses Config to parse functionConfig, Transform to change the image
+type SetImage struct {
+	// Image is the desired image
+	Image types.Image `json:"image,omitempty" yaml:"image,omitempty"`
```
yuwenma, line 15: Is it necessary to import the kustomize/api

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,179 @@
...
+			imageTrans.Image.Name = val
+		} else if key == "newName" {
+			imageTrans.Image.NewName = val
+		} else if key == "newTag" {
+			imageTrans.Image.NewTag = val
+		} else if key == "digest" {
+			imageTrans.Image.Digest = val
+		} else {
+			return fmt.Errorf("SubObject has unmatched field type: `data`")
+		}
```
yuwenma, line 39: Suggest using switch/case here.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return nil
+}
+
+// validateInput validates the inputs passed into via the functionConfig
+func (imageTrans *SetImage) validateInput() error {
+	if imageTrans.Image.Name == "" {
+		return fmt.Errorf("missing image name or typo")
```
yuwenma, line 47: I see many of my previous comments are not addressed. Can you iterate those comments so we can make the review more efficient? 
e.g.
https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/904#discussion_r946242946
https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/904#discussion_r946243457

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,179 @@
...
+				}
+				for _, vecObj := range podSpec.GetSlice("iniContainers") {
+					if err := imageTrans.updateImages(vecObj, &imageTrans.Image, o); err != nil {
+						return err
+					}
+				}
+			}
+		}
+	}
+	return nil
```
yuwenma, line 75: My previous comment seems to be not resolved  https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/904#discussion_r946246928

yuwenma: ditto for the following `setPodContainers`, `hasPodSpecContainers` functions.

functions/go/set-image/custom/custom.go
zyy98: Leftover after my refactor for custom, my bad. Should be error.

functions/go/set-image/image_util/image_util.go
```diff
@@ -0,0 +1,78 @@
...
+	Kind    string `json:"kind,omitempty" yaml:"kind,omitempty"`
+}
+
+// IsImageMatched returns true if the value of t is identical to the
+// image name in the full image name and tag as given by s.
+func IsImageMatched(s, t string) bool {
+	// Tag values are limited to [a-zA-Z0-9_.{}-].
+	// Some tools like Bazel rules_k8s allow tag patterns with {} characters.
+	// More info: https://github.com/bazelbuild/rules_k8s/pull/423
+	pattern, _ := regexp.Compile("^" + t + "(@sha256)?(:[a-zA-Z0-9_.{}-]*)?$")
```
natasha41575, line 45: Normally `regexp.Compile` is assigned to a global var. Otherwise, this line gets executed every time `updateImages` is run even though the value of `pattern` is always the same. Suggest moving `pattern` out to be a global var.

Normally global vars are not good, but this is an exception because it is more like a const (the value never changes).

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+		case "name":
+			imagetrans.Image.Name = val
+		case "newName":
+			imagetrans.Image.NewName = val
+		case "newTag":
+			imagetrans.Image.NewTag = val
+		case "digest":
+			imagetrans.Image.Digest = val
+		default:
+			return fmt.Errorf("SubObject has unmatched field type: `data`")
```
natasha41575, line 38: nit: surface the value of `key` in the error message

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}
+	return nil
+}
+
+// validateInput validates the inputs passed into via the functionConfig
+func (imagetrans *SetImage) validateInput() error {
+	// TODO: support container name and only one argument input in the next PR
+	if imagetrans.Image.Name == "" {
+		return fmt.Errorf("missing `name` to select the image")
```
natasha41575, line 48:
```suggestion
		return fmt.Errorf("`name` field is missing from image selector")
```

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if imagetrans.Image.NewName == "" && imagetrans.Image.NewTag == "" && imagetrans.Image.Digest == "" {
+		return fmt.Errorf("missing `newName`, `newTag`, or `digest`, at least one of the fields should not be empty")
```
natasha41575, line 51:
```suggestion
		return fmt.Errorf("must specify one of `newName`, `newTag`, or `digest`")
```

See the kpt error message style guide: go/kpt-errors-style-guide

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if imagetrans.Image.NewTag != "" && imagetrans.Image.Digest != "" {
+		return fmt.Errorf("image `newTag` and `digest` both set, only one field can be set")
```
natasha41575, line 54:
```suggestion
		return fmt.Errorf("image `newTag` and `digest` both set, set only one")
```

natasha41575 (commented):
Looks pretty good! Left some nits

functions/go/set-image/transformer/images_transformer.go
zyy98: Thanks a lot for the guide!

examples/set-image-digest/.expected/results.yaml
```diff
@@ -7,28 +7,22 @@ items:
...
           kind: Pod
           name: app2
-        field:
-          path: spec.containers.image
-          currentValue: nginx:1.20.2
-          proposedValue: bitnami/nginx@sha256:e1d6f00f191515945233f562777fc9ab3b7637ea75169b3dd628d46c9b24400f
         file:
           path: resources.yaml
           index: 1
+      - message: 'summary: updated a total of 2 image(s)'
```
yuwenma, line 27: the new message is much nicer!

functions/go/set-image/image_util/image_util.go
```diff
@@ -0,0 +1,80 @@
+package image_util
```
yuwenma, line 1: Does this (partially) come  from kustomize?

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	// ONLY for kustomize, AdditionalImageFields is the user supplied fieldspec
+	AdditionalImageFields image_util.FsSlice `json:"additionalImageFields,omitempty" yaml:"additionalImageFields,omitempty"`
+	// context logs each detailed result
+	context *fn.Context
+	// resultCount logs the total count image change
+	resultCount int
+}
+
+// Config transforms the data from ConfigMap to SetImage struct
+func (imagetrans *SetImage) Config() error {
```
yuwenma, line 26: Can you verify the previous comment is resolved? See the second example. https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/904#discussion_r953264086

functions/go/set-image/image_util/image_util.go
zyy98: Yes, they are practically the same functions as kustomize, since I need to avoid importing kustomize package, I write my own util package

functions/go/set-image/transformer/images_transformer.go
zyy98: I am not sure which one I missed here, I have given every variable name ` to show the exact name. And I have changed T to t.

zyy98: Oh I see!! you mean change the whole name to `t`.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+		case "name":
+			imagetrans.Image.Name = val
+		case "newName":
+			imagetrans.Image.NewName = val
+		case "newTag":
+			imagetrans.Image.NewTag = val
+		case "digest":
+			imagetrans.Image.Digest = val
+		default:
+			return fmt.Errorf("SubObject has unmatched field type: `data` with wrong field name %v", key)
```
yuwenma, line 38: I think we already talked about this before (can't find my previous comment). but it does not really make sense to the users about what `SubObject` (or `KubeObject`) means and how they can fix it.   Is there any specific reason to use `SubObject` here?

zyy98: I see. I used this phrase to keep the same format with previous unit test result. I agree I should give a better message.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}
+	return nil
+}
+
+// validateInput validates the inputs passed into via the functionConfig
+func (imagetrans *SetImage) validateInput() error {
+	// TODO: support container name and only one argument input in the next PR
+	if imagetrans.Image.Name == "" {
+		return fmt.Errorf("`name` field is missing from image selector")
```
yuwenma, line 48: it is less clear from user's perspective about what "image selector" means. See my previous comment https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/904#discussion_r946242946 about what user message should and should not include.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if imagetrans.Image.NewName == "" && imagetrans.Image.NewTag == "" && imagetrans.Image.Digest == "" {
+		return fmt.Errorf("must specify one of `newName`, `newTag`, or `digest`")
```
yuwenma, line 51: This user message is much better. 👍

functions/go/set-image/image_util/image_util.go
natasha41575: I think if it is the same (or almost the same) as another open source lib, we may need to fork it under a third_party directory with a clear README explaining where the code came from. @yuwenma may have better guidance here

functions/go/set-image/image_util/image_util.go
```diff
@@ -0,0 +1,80 @@
...
+
+var pattern *regexp.Regexp
+
+// IsImageMatched returns true if the value of t is identical to the
+// image name in the full image name and tag as given by s.
+func IsImageMatched(s, t string) bool {
+	// Tag values are limited to [a-zA-Z0-9_.{}-].
+	// Some tools like Bazel rules_k8s allow tag patterns with {} characters.
+	// More info: https://github.com/bazelbuild/rules_k8s/pull/423
+	pattern, _ = regexp.Compile("^" + t + "(@sha256)?(:[a-zA-Z0-9_.{}-]*)?$")
```
natasha41575, line 47: Oh, I misread this (didn't realize that the string t is involved here). Sorry about that, `pattern` actually is changing each time. You can revert the change I suggested (pattern does not need to be global). That's my mistake, sorry about that!

functions/go/set-image/transformer/images_transformer.go
zyy98: I see. The wording suggestion comes from Natasha. Do you have a better idea? Should I just write "must specify `name`"

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-image/image_util"
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+// SetImage supports the set-image workflow, it uses Config to parse functionConfig, Transform to change the image
+type SetImage struct {
+	// Image is the desired image
+	Image image_util.Image `json:"image,omitempty" yaml:"image,omitempty"`
+	// ConfigMap keeps the data field that holds image information
+	ConfigMap map[string]string `json:"data,omitempty" yaml:"data,omitempty"`
```
yuwenma, line 16: ConfigMap is not accurate and can be misleading. you can call it `DataFromDefaultConfig` `DataFromDefault`

functions/go/set-image/image_util/image_util.go
zyy98: Thanks for the reminder, I missed the `t` when I read your comments as well.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+import (
+	"fmt"
+
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-image/custom"
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-image/image_util"
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+// SetImage supports the set-image workflow, it uses Config to parse functionConfig, Transform to change the image
+type SetImage struct {
```
yuwenma, line 12: (I think we've talked about the `SetImage` attributes structure and the long term plans for `additionalImageField` as well?)  
The general ideas are:
1. The core function configs like `newImage` `containerName` `imageName` should live in this file. The main focus of the setImage is for core k8s resources.
2. We keep the `additionalImageField` for backward compatibility reasons. Eventually it will be dropped once we have a solution for CRD resources like supporint OpenAPI schema.

So the code should be structured in a way that can clearly distinguish the `additionalImageFields` and the core k8s resource logics, and dropping the `additionalImageField` interface should not require any code refactoring, or even changes except removing the `additionalImageField` line from `SetImage` struct

yuwenma: Based on that, the code should be look like 
```go
// image_transformer.go
type SetImage struct {
   Image Image `json:"image,omitempty" yaml:"image,omitempty"`

   ... 
   AdditionalImageFields types.FsSlice `json:"additionalImageFields,omitempty" yaml:"additionalImageFields,omitempty"`
}

type Image struct {
    Name string `json:"name,omitempty" yaml:"name,omitempty"`
    NewName string `json:"newName,omitempty" yaml:"newName,omitempty"`
    NewTag string `json:"newTag,omitempty" yaml:"newTag,omitempty"`
    Digest string `json:"digest,omitempty" yaml:"digest,omitempty"`
}

func (r *SetImage) Run(...) {
    if t.AdditionalImageFields != nil {
        custom.SetAdditionalFieldSpec(t.GetMap("image"), items, t.AdditionalImageFields, t.context)
    }
}
```
// custom.go
func SetAdditionalFieldSpec(imageObject fn.SubObject, objects fn.KubeObjects, addImgFields types.FsSlice, ctx *fn.Context) {
    image, additionalImageFields := transformStruct(img, addImgFields)

    for i, obj := range objects {
        objRN, err := yaml.Parse(obj.String())
        if err != nil {
            ctx.ResultErr(err.Error(), obj)
        }
        filter := imagetag.Filter{
            ImageTag: types.Image{
                Name:    imageObject.NestedStringOrDie("name"),
                NewName: imageObject.NestedStringOrDie("newName"),
                NewTag:  imageObject.NestedStringOrDie("newTag"),
                Digest:  imageObject.NestedStringOrDie("digest"),
            },
            FsSlice:  additionalImageFields,
        }

        filter.WithMutationTracker(LogResultCallback(ctx, obj))
        err = filtersutil.ApplyToJSON(filter, objRN)
        if err != nil {
            ctx.ResultErr(err.Error(), obj)
        }
        newObj, err := fn.ParseKubeObject([]byte(objRN.MustString()))
        if err != nil {
            ctx.ResultErr(err.Error(), obj)
        }
        objects[i] = newObj
    }
}
```

File: functions/go/set-image/transformer/images_transformer.go [hunk 38]
```diff
@@ -0,0 +1,170 @@
...
+	// Image is the desired image
+	Image image_util.Image `json:"image,omitempty" yaml:"image,omitempty"`
+	// ConfigMap keeps the data field that holds image information
+	ConfigMap map[string]string `json:"data,omitempty" yaml:"data,omitempty"`
+	// ONLY for kustomize, AdditionalImageFields is the user supplied fieldspec
+	AdditionalImageFields image_util.FsSlice `json:"additionalImageFields,omitempty" yaml:"additionalImageFields,omitempty"`
+	// context logs each detailed result
+	context *fn.Context
```
yuwenma, line 20: nit: let's remove this attribute. You can pass in the context to each function that need to log results.

functions/go/set-image/transformer/images_transformer.go
zyy98: This structure is pretty much what I have here. Except I have the image definition inside a separate package. Because `custom` package need to import `Image` struct, but `transform` package already imports `custom`, to avoid dependency error I have a separate package `image_util` and put `Image` struct here, so that both `transformer` and `custom` can import

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+		if err := t.updateImages(vecObj, &t.Image, o); err != nil {
+			return err
+		}
+	}
+	for _, vecObj := range podSpec.GetSlice("iniContainers") {
+		if err := t.updateImages(vecObj, &t.Image, o); err != nil {
+			return err
+		}
+	}
+	return nil
```
yuwenma, line 71: I saw multiple functions call `updateImages` on "containers" list and "initContainer"list, we should use a function to avoid duplicate code.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if newImage.Digest != "" {
+		tag = "@" + newImage.Digest
+	}
+	newName := name + tag
+	return newName
+}
+
+// updateImages update the image for a given fieldpath
+func (t *SetImage) updateImages(o *fn.SubObject, newImage *image_util.Image, parentO *fn.KubeObject) error {
```
yuwenma: the argument `newImage` is not needed. You should be able to get it from `t.Image`

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	oldValue := o.NestedStringOrDie("image")
+	if !image_util.IsImageMatched(oldValue, newImage.Name) {
+		return nil
+	}
+	newName := getNewImageName(oldValue, newImage)
+	err := o.SetNestedString(newName, "image")
+
+	msg := fmt.Sprintf("updated image from %v to %v", oldValue, newName)
```
yuwenma, line 130: is that possible the old and new image names are the same?  If so, we should not update

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	t.context.ResultInfo(msg, parentO)
+	t.resultCount += 1
+	return err
```
yuwenma, line 133: It's better to log result (info, error or warning) in the same function, otherwise it would be harder to figure out what each function does. 

Also, passing the parentO into the function merely for the `ResultInfo` seems to be an overkill. See previous comment https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/904#discussion_r946248335

yuwenma: Some pseudo code to explain my comments (and other comments about the duplicate code issue around "container" "initContainers", unnecessary argument passed into the function) 
 
```go
func (t *SetImage) updatePodImages(pod *fn.SubObject, ctx *fn.Context) error {
       var containers fn.SliceSubObjects 
       containers = append(containers, pod.GetSlice("initContainers")...)
       containers = append(containers, pod.GetSlice("containers")...)

       for _, container := range containers {
         oldName := o.NestedStringOrDie("image")
         if !IsImageMatched(oldName, t.Image.Name) {
             continue
         }
         newName := getNewImageName(oldName, newImage)
         if newImageName == oldName {
            continue
         }
         err := o.SetNestedString(newName, "image")
         if err != nil {
             ctx.ResultErr()
         }  else {
             ctx.ResultInfo
         }
       }
```

zyy98: For one Object, it can have multiple changes in its images. If I just return a bool, it is not enough to record all the old names that have changed. One way is to keep an internal struct, that tracks all the results, and log them in the end. Another way is to pass in its parent Object, and then I can have the logging right after the update

functions/go/set-image/image_util/image_util.go
yuwenma: I don't think in this setimage use case we should need such a util lib that borrow most of the logic from other libraries. 

If we do, it should be forked. as @natasha41575 suggested

functions/go/set-image/transformer/images_transformer.go
yuwenma: there are two major differences (See my first comment about the general ideas). 
1. The dependency error should not be the reason to duplicate code from another library. Suggest taking a deeper look at the parameters I used in the code, I use `t.GetMap("image")` and regenerate the `types.Image` via fn.SubObject, so it won't trigger dependency loop.
2. "dropping the additionalImageField interface should not require any code refactoring, or even changes". In this PR, the code needs a refactoring because importing "image_utiles" is no longer a necessary workaround for depedency loop but split two closely related structure `SetImage` and `Image` apart.

zyy98: I see your point! One other thing that I would like to ask about. I see you use `types.FsSlice` in `transformer`, but it is best to avoid kustomize pacakges here. Can I have my duplicate definition of `FsSlice` inside `image_util` here?

functions/go/set-image/transformer/images_transformer.go
zyy98: Sorry somehow I missed your previous comments. Will take a look now.

functions/go/set-image/transformer/images_transformer.go
yuwenma: `t.GetMap("image")` --> `fnConfig.GetMap("image")` The fnConfig is the second parameter in the `Run` method.

If the tradeoff is between introducing `image_util` and  defining `types.FnSlice` in `SetImage` struct, I would prefer we choose the latter. For `image_util` the right way is to fork kustomize like @natasha41575 suggested, not borrow the code directly. I don't see value to introduce this complexity TBH.

functions/go/set-image/transformer/images_transformer.go
yuwenma: no return error should be needed. The code is pseudo code

zyy98: I know it is pseudo code and I am not asking about return error. My reply was edited. Do you mind checking my updated comments?

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	template := spec.GetMap("template")
+	if template == nil {
+		return false
+	}
+	podSpec := template.GetMap("spec")
+	return podSpec != nil
+}
+
+func (t *SetImage) hasPodContainers(o *fn.KubeObject) bool {
```
yuwenma, line 100: This should be determined by the resource "Kind", not the existence of a field.

zyy98: The previous version does not check for the Kind, it just applies on all kinds

functions/go/set-image/transformer/images_transformer.go
yuwenma: We talked about that the "image" refers to the "Pod" container image. I didn't read the previous code carefully enough, I think the code does not show this aspect. I'd expect 
1. A "Pod" resource is determined by its "Kind" not the existence of "spec.template.spec". Specifically, the `hasPodContainers` function which checks the existence of `spec` is not right and misleading because all resources have `spec` field, but they are not guaranteed to be a Pod.   
2. Can you verify that in core k8s resources (the same approach as we discussed in set-labels) or kustomize setImage (Maybe @natasha41575 can help giving some guidance here) that only Pod should have containers that need to set the image. 
3. For `SetImages` functions, a more reasonable code structure should be:  for any resources other than Pod, if it contains pod object, get its `Pod` SubObject and then call the Pod `SetContainers` functions.
4. `SetPodImages` --> maybe "ContainerImage" is a more accurate name, at least that's how the code should be structured.

zyy98: I have verified and code is updated. I checked the structs in kubernetes, it seems they will not contain `Pod` as `SubObject`

functions/go/set-image/transformer/images_transformer.go
zyy98: You are right about the kind. I must have misread the previous code. Matching it only by path is so wrong.

functions/go/set-image/transformer/images_transformer.go
zyy98: I apply the same way for `FsSlice`, using `fn.SliceSubObjects`

functions/go/set-image/third_party/README.md
```diff
@@ -0,0 +1,3 @@
+This package is forked from
+`sigs.k8s.io/kustomize/api/`
+with version `api/0.8.11`.
```
natasha41575, line 3: Did you actually fork version `api/0.8.11`, or did you copy from master branch?

zyy98: by `api/0.8.11`, does it mean branch `release-api-v0.8` ? It is the same as that version

natasha41575: Ah, so (1) we should use tags and not branches, and (2) we should fork from the most recent.

So you can use the tag `api/v0.12.1`. Here is the link to the code: https://github.com/kubernetes-sigs/kustomize/tree/api/v0.12.1

zyy98: I see. Thanks a lot!

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if t.AdditionalImageFields != nil {
+		custom.SetAdditionalFieldSpec(functionConfig.GetMap("image"), items, functionConfig.GetSlice("additionalImageFields"), ctx)
+	}
+
+	summary := fmt.Sprintf("summary: updated a total of %v image(s)", t.resultCount)
+	ctx.ResultInfo(summary, nil)
+}
+
+// Config transforms the data from ConfigMap to SetImage struct
```
yuwenma, line 76: typo: `Config` or `configDefaultData`

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,199 @@
...
+func (t *SetImage) validateInput() error {
+	// TODO: support container name and only one argument input in the next PR
+	if t.Image.Name == "" {
+		return fmt.Errorf("must specify `name`")
+	}
+	if t.Image.NewName == "" && t.Image.NewTag == "" && t.Image.Digest == "" {
+		return fmt.Errorf("must specify one of `newName`, `newTag`, or `digest`")
+	}
+	if t.Image.NewTag != "" && t.Image.Digest != "" {
+		return fmt.Errorf("image `newTag` and `digest` both set, set only one")
```
yuwenma, line 105: Shall we support the case where both New Tag and digest are given? This is the feature I mentioned in the 1:1 that kustomize supporting a new use case where the image contains both tag and digest. https://github.com/kubernetes-sigs/kustomize/pull/4406
CC @natasha41575 for more guidance

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,199 @@
...
+	var result []setImageResult
+	for _, o := range containers {
+		oldValue := o.NestedStringOrDie("image")
+		if !image.IsImageMatched(oldValue, t.Image.Name) {
+			continue
+		}
+		newName := getNewImageName(oldValue, t.Image)
+		if oldValue == newName {
+			return nil, nil
+		}
```
yuwenma, line 125: I think the right logic should be using `continue` here rather than `return`.  And we should have a e2e test to cover this.

functions/go/set-image/transformer/images_transformer.go
natasha41575: Yeah, kustomize now supports both tag and digest. I think this is a popular use case. IMO we should support it in set-image. 

But I think it's ok if we do it in a followup if we are running short on time for this PR.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (t SetImage) logResult(ctx *fn.Context, err error, result []setImageResult, o *fn.KubeObject) {
+	if err != nil {
+		ctx.ResultErr(err.Error(), o)
+	}
+	for _, val := range result {
+		msg := fmt.Sprintf("updated image from %v to %v", val.currentValue, val.proposedValue)
+		ctx.ResultInfo(msg, o)
+	}
```
yuwenma, line 198: I might be a little bit out of sync about this. I thought we decide to not include each individual info updates in the `resourcelist.results`， right?

functions/go/set-image/transformer/images_transformer.go
yuwenma: If that's the case, suggest removing this validation check in this PR and add the tag and digest logic in the next PR, so as the additional image fieldspec behavior are aligned with kustomize.

functions/go/set-image/transformer/images_transformer.go
zyy98: Sorry about the confusion. I did not remember we discussed that. Previous version did include this message. So I kept it. I will delete it, and only leave the error messages and total info message

yuwenma: The main problem is that kpt normally truncate the result if it is too long. And if we keep the detailed info for each resource change, more useful information like the summary and error logs will be truncated.

yuwenma: The same problems happens on set-labels and we should fix that as well.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -0,0 +1,196 @@
...
+		return fmt.Errorf("must specify `name`")
+	}
+	if t.Image.NewName == "" && t.Image.NewTag == "" && t.Image.Digest == "" {
+		return fmt.Errorf("must specify one of `newName`, `newTag`, or `digest`")
+	}
+	return nil
+}
+
+// updateContainerImages updates the images inside containers, return potential error, and a list of logging results
+func (t *SetImage) updateContainerImages(pod *fn.SubObject) (error, []setImageResult) {
```
yuwenma, line 108: nit: Normally in the return values, the error is placed at the end. e.g. `([]setImageResult, error)`

functions/go/set-image/transformer/images_transformer.go
zyy98: My bad. I thought truncation would only leave out the early log results, and the total logging would remain.

yuwenma (approved):

PR #908, opened 2022-08-29
remove KUBEVAL_BUILDER_IMAGE from generic defaults

sdowell:
The defaults file is intended for generic default values that can be
shared for Dockerfiles across multiple functions. In this case
KUBEVAL_BUILDER_IMAGE is specific to the kubeval function and can simply
be defined in the dockerfile for that function.


sdowell:
Can I get a review on this?

yuwenma (approved):
looks great! Thanks for the change

PR #909, opened 2022-08-30
fix: `set-image` with more matching rules

zyy98:
to fix issue https://github.com/GoogleContainerTools/kpt/issues/3444
provide `containerName` and `imageName` as input, support `containerName` matching, support `imageName` omission inside input

### DIFF
--- a/functions/go/set-image/transformer/images_transformer.go
+++ b/functions/go/set-image/transformer/images_transformer.go
@@ -9,12 +9,18 @@ import (
 	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
 )
 
-// Image contains an image name, a new name, a new tag or digest, which will replace the original name and tag.
 type Image struct {
-	// Name is a tag-less image name.
+	// DEPRECATED
+	//Name is a tag-less image name. should be deprecate, means image name
 	Name string `json:"name,omitempty" yaml:"name,omitempty"`
 
-	// NewName is the value used to replace the original name.
+	// ContainerName is the name for container
+	ContainerName string `json:"containerName,omitempty" yaml:"containerName,omitempty"`
+
+	// ImageName is the image name
+	ImageName string `json:"imageName,omitempty" yaml:"imageName,omitempty"`
+
+	// NewName is the value used to replace the original name, replace image name
 	NewName string `json:"newName,omitempty" yaml:"newName,omitempty"`
 
 	// NewTag is the value used to replace the original tag.
@@ -35,6 +41,10 @@ type SetImage struct {
 	AdditionalImageFields types.FsSlice `json:"additionalImageFields,omitempty" yaml:"additionalImageFields,omitempty"`
 	// resultCount logs the total count image change
 	resultCount int
+	// currentImageNames records current image name, if no image name is provided as input, there should only be one ImageNames
+	currentImageName string
+	// isImageSelected checks if the input provide the image name or not
+	isImageSelected bool
 }
 
 // Run implements the Runner interface that transforms the resource and log the results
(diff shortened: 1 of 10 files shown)


functions/go/set-image/transformer/images_transformer.go
```diff
@@ -2,19 +2,24 @@ package transformer
...
 	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-image/third_party/sigs.k8s.io/kustomize/api/image"
 	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-image/third_party/sigs.k8s.io/kustomize/api/types"
 	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
 )
 
-// Image contains an image name, a new name, a new tag or digest, which will replace the original name and tag.
 type Image struct {
-	// Name is a tag-less image name.
+	// DEPRECATED
+	//Name is a tag-less image name. should be deprecate, means image name
```
yuwenma, line 13: `// Name is a tag-less image name.` the rest info seems to be redundant.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -2,19 +2,24 @@ package transformer
...
 	Name string `json:"name,omitempty" yaml:"name,omitempty"`
 
-	// NewName is the value used to replace the original name.
+	// ContainerName is the name for container
```
yuwenma, line 16: `// ContainerName is the Pod's container name. It is used to choose the matching containers to update the images.`

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -2,19 +2,24 @@ package transformer
...
+	ContainerName string `json:"containerName,omitempty" yaml:"containerName,omitempty"`
+
+	// ImageName is the name for image
```
yuwenma, line 19: ditto

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -2,19 +2,24 @@ package transformer
...
+	ImageName string `json:"imageName,omitempty" yaml:"imageName,omitempty"`
+
+	// NewName is the value used to replace the original name, replace image name
```
yuwenma, line 22: "replace image name" is ambiguous.
`// NewName is the new image name`

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -43,21 +48,21 @@ func (t SetImage) Run(ctx *fn.Context, functionConfig *fn.KubeObject, items fn.K
 	if err != nil {
 		ctx.ResultErrAndDie(err.Error(), nil)
 	}
-	err = t.validateInput()
+	err = t.validateInput(items)
```
yuwenma, line 51: Suggest adding a sanity check to notify users if the `ResourceList.Items`  is empty. Like what we did in set-labels

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -89,68 +98,147 @@ func (t *SetImage) configDefaultData() error {
...
+// getContainers gets the containers inside kubeObject
+func getContainers(o *fn.KubeObject) fn.SliceSubObjects {
+	switch o.GetKind() {
+	case "Pod":
+		return getPodContainers(o)
+	case "Deployment", "StatefulSet", "ReplicaSet", "DaemonSet", "PodTemplate":
+		return getPodSpecContainers(o)
+	}
 	return nil
 }
```
yuwenma, line 210: suggest adding it back to the main `Run` method. This part belongs to the main logic so it's better to be replaced in the main function

zyy98: `Run` method already uses this function `getContianers`, I have this as a separate function because I feel I need to iterate them twice, once for validating user input, once for update the images. To reduce code duplication, I have this function here.

functions/go/set-image/transformer/images_transformer.go
```diff
@@ -43,21 +48,28 @@ func (t SetImage) Run(ctx *fn.Context, functionConfig *fn.KubeObject, items fn.K
 	if err != nil {
 		ctx.ResultErrAndDie(err.Error(), nil)
 	}
-	err = t.validateInput()
-	if err != nil {
-		ctx.ResultErrAndDie(err.Error(), nil)
+
+	items = items.WhereNot(func(o *fn.KubeObject) bool { return o.IsLocalConfig() })
```
yuwenma, line 52: why not using the `fn.IsLocalConfig` function?
```go
items = items.WhereNot(fn.IsLocalConfig)
```

PR #911, opened 2022-09-09
add wasm functionality to apply-replacements

wmuizelaar:
Like https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/898 - this PR adds wasm functionality to the apply-replacements function.

PR #915, opened 2022-09-16
add wasm functionality to apply-setters and rewrite to new go kpt fn sdk

wmuizelaar:
Just like #911 - this adds wasm functionality to `apply-setters`. As discussed during the latest office-hours I also attempted to port the functionality to the new SDK.

Please review thoroughly and share feedback - I tried to keep as much of the old code in place as possible, but I wasn't able to properly read the 'old' ConfigMap configuration structure, so had to craft a new config format - which is obviously a breaking change. I'm not too happy with that, so ideas on how to prevent that are appreciated ;-)


yuwenma:
Hi, @wmuizelaar, thank you for the contribution. 

We do not want to add new features to the apply-setter functions, as explained in https://github.com/GoogleContainerTools/kpt/issues/3131. For your use cases, I'd recommend using apply-replacements function instead as we talked about in the office hour. I understand the apply-replacement may not be sufficient enough from the feasibility and maintenance perspective. If that's the case, do you mind sharing your examples in https://github.com/GoogleContainerTools/kpt/issues/3131 so we can evaluate and further help.

PR #920, opened 2022-09-27
Fix some incorrect/duplicate go.mod paths

justinsb:
This was causing some IDE warnings.


yuwenma:
@justinsb Thanks for the fix. These functions are placed in this repo but owned by some other teams due to legacy reasons. Not sure how those import issues are introduced but I'm trying to figure out a clean path for these legacy functions. So I won't worry too much about their existing issues.

yuwenma (approved):

PR #922, opened 2022-09-29
Create bind function

justinsb:
Trying a more lightweight approach to creating these functions also,
as the existing process has become quite heavy.


functions/go/bind/main.go
```diff
@@ -0,0 +1,76 @@
...
+	f := Bind{}
+
+	err := f.LoadConfig(rl.FunctionConfig)
+	if err != nil {
+		rl.Results = append(rl.Results, fn.ErrorConfigObjectResult(fmt.Errorf("functionConfig error: %w", err), rl.FunctionConfig))
+		return true, nil
+	}
+
+	if f.Object == nil {
+		return false, fmt.Errorf("`data.object` should not be empty")
```
yuwenma, line 32: Do you mean`fmt.Errorf("functionConfig should not be empty")`?

functions/go/bind/main.go
```diff
@@ -0,0 +1,76 @@
...
+	// Find the matching localconfig object
+	var bindingObject *fn.KubeObject
+	{
+		// TODO: Match GK or GVK?
+		gvk := f.Object.GroupVersionKind()
+
+		matchAnnotations := map[string]string{
+			"config.kubernetes.io/local-config": "binding",
+		}
+		matches := objects.Where(fn.HasAnnotations(matchAnnotations)).Where(fn.IsGVK(gvk.Group, gvk.Version, gvk.Kind))
```
yuwenma, line 57: nit: `.Where(fn.IsGVK(gvk.Group, gvk.Version, gvk.Kind))` --> `.Where(f.Object.IsGVK)`

functions/go/bind/pkg/meta/kinds.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// IsClusterScoped returns true if the specified GVK is cluster scoped, or an error if this cannot be determined.
+func IsClusterScoped(gvk schema.GroupVersionKind) (bool, error) {
+	ki := findKindInfo(gvk)
+	if ki == nil {
+		return false, fmt.Errorf("kind %v not known", gvk)
+	}
+	return ki.ClusterScoped, nil
+}
```
yuwenma, line 30: FYI, the go SDK has `IsClusterScoped` `IsNamespaceScoped` functions that also relies on the hard-coded namespace list https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/fn/internal/namespace.go#L8. I'm happy to extend the list to include CNRM and change the IsClusterScoped to only require GVK args.

functions/go/bind/pkg/meta/namespaceref.go
```diff
@@ -0,0 +1,82 @@
...
+
+type namespaceRef struct {
+	parentObject *fn.SubObject
+
+	name      string
+	fieldPath []string
+}
+
+func buildMetadataNamespaceReference(parentObject *fn.KubeObject) (Ref, error) {
+	namespace := parentObject.GetMap("metadata").GetString("namespace")
```
yuwenma, line 32: `parentObject.GetNamespace()` gives the same result I think. the `GetMap` can return `nil` and then causes panic I think https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/fn/object.go#L962

justinsb:
Merged as part of #923 

We'll deal with the comments here in a separate PR!

PR #925, opened 2022-10-04
Create set-name-prefix function

justinsb:
set-name-prefix function replaces an "old" prefix on all names,
including references to those names.

Where the whole name matches the "old" prefix, it replaces the whole
name.


functions/go/set-name-prefix/setnameprefix.go
```diff
@@ -0,0 +1,100 @@
...
+	return nil
+}
+
+func (f *SetNamePrefix) Transform(objects fn.KubeObjects) error {
+	for _, object := range objects {
+		if object.IsLocalConfig() {
+			continue
+		}
+		if kpt.IsResourceGroup(object) {
+			continue // Should ResourceGroup be marked as local config?
```
yuwenma, line 79: "local-config" might not be accurate because kpt does expect the RG CR in the cluster. On the other side, I think  It will be more helpful for users to read this RG resource as "some client-side deterministic metadata"

yuwenma (approved):

PR #941, opened 2022-11-01
docs: Update tags for go/export-terraform/v0.2.0

yuwenma:


yuwenma:
Some previous function releases do not have the doc updated (latest export-terraform is v0.3.0), this blocks updating the catalog.kpt.dev site.

PR #947, opened 2022-11-10
fix: make 'gcloud' as optional in Makefile for ko push

yuwenma:
fix louhi release pipeline which does not have `gcloud` installed in its build image. https://louhi.dev/?projectId=5662083073179648#/execution-detail/6327051601313792


yuwenma:
cc @justinsb

PR #955, opened 2022-11-22
Skip set-name-prefix branch when generating the function catalog

ChristopherFry:
This allows the site to be generated successfully. Currently `site-generate` make command fails with `Error getting metadata for "set-name-prefix" from "remotes/origin/set-name-prefix/v0.0"` preventing the CI `Generate Site` step from completing successfully.

### DIFF
--- a/scripts/generate_catalog/generate_catalog.go
+++ b/scripts/generate_catalog/generate_catalog.go
@@ -56,6 +56,7 @@ var keywordsPool = []string{
 
 var (
 	branchesToSkip = []string{
+		"set-name-prefix/v0.0",
 		"sops/v0.2",
 	}
 )


scripts/generate_catalog/generate_catalog.go
```diff
@@ -56,6 +56,7 @@ var keywordsPool = []string{
 
 var (
 	branchesToSkip = []string{
+		"set-name-prefix/v0.0",
```
sdowell, line 59: I think this works as a stopgap, but ideally we should maintain a consistent pattern/UX across functions IMO. If the goal is to migrate to a new convention, we should still make sure things are consistent (and not broken) in the transition

sdowell (approved):

scripts/generate_catalog/generate_catalog.go
ChristopherFry: Completely agree. I'll sync with the team after the holidays to confirm the path forward for this function in regards to the documentation.

PR #957, opened 2022-11-22
Fix external link on apply replacements documentation

ChristopherFry:
Fixes the external link on the apply replacements function documentation in reference to  GoogleContainerTools/kpt/issues/3594.

### DIFF
--- a/functions/go/apply-replacements/README.md
+++ b/functions/go/apply-replacements/README.md
@@ -41,4 +41,4 @@ replacements:
 ```
 <!--mdtogo-->
 
-[kustomize replacements]: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/
+[kustomize replacements]: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/ ':target=_blank'


functions/go/apply-replacements/README.md
```diff
@@ -41,4 +41,4 @@ replacements:
 ```
 <!--mdtogo-->
 
-[kustomize replacements]: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/
+[kustomize replacements]: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/ ':target=_blank'
```
sdowell, line 44: I'm not much of a CSS expert so forgive me (I assume this is CSS). From some brief googling it looks like this tells the browser to open this link in a new tab. How does this fix the link?

ChristopherFry: Good question. Essentially we're hitting a cross-origin issue here. We're on domain A, and the browser isn't allowing us to redirect to domain B.

I'm taking a slightly opinionated approach to correct this. Since we're linking to an external domain, I'm choosing to link to it by opening up a new tab. This is consistent with [kpt.dev Contact page](https://kpt.dev/contact/) and [Config Sync documentation](https://cloud.google.com/anthos-config-management/docs/config-sync-overview). Being this is in a new tab, browsing to domain B directly is allowed.

Another way I can resolve this is similar to what we do in the sidebar, and add the `:crossorgin`. This would correct the issue allowing us to browse to the domain B in the same tab.

sdowell: SGTM if this is consistent with other external links

sdowell (approved):

PR #969, opened 2023-01-26
fix: avoid nameclashes by considering the filepath

bluebrown:
Signed-off-by: Nico Braun <rainbowstack@gmail.com>

fix: kpt#3758


bluebrown:
Imo, this implementation is pretty hacky. There is probably a better way.

PR #978, opened 2023-02-20
Improve WASM runtime error messages by adding a function to retrieve errors from resourceList

wleese:
We'd like WASM compiled functions to provide the same information as those run in docker and aims to solve https://github.com/GoogleContainerTools/kpt/issues/3822

This PR also resolves a panic that can occur when an `error` is returned in the constructed javascript function.

In order to actually surface the error messages, the kpt wasm function runner needs to be modified.

This can be tested with:  

Kptfile
```
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: kpt
  annotations:
info:
  description: this package provides the parent package for your application
pipeline:
  mutators:
    - image: gcr.io/kpt-fn/starlark:v0.4.3
      configPath: starlarkconfig.yaml
```
resources.yaml
```
apiVersion: iam.cnrm.cloud.google.com/v1beta1
kind: IAMPolicyMember
metadata:
  name: bucket-read-write-by-service-account-diogotuesday274
  namespace: bb8-tuesdayds
```
starlarkconfig.yaml
```
kind: StarlarkRun
metadata:
  name: apply-team-name-to-iampolicymember
source: |
  team = ctx.resource_list["functionConfig"]["params"]["team"]
```
And running:
`kpt fn render . --allow-alpha-wasm`
and
`KPT_FN_WASM_RUNTIME=nodejs kpt fn render . --allow-alpha-wasm`


functions/go/apply-replacements/run_js.go
```diff
@@ -37,19 +41,40 @@ func applyReplacements(input []byte) ([]byte, error) {
...
 }
 
-func resourceListProcessorWrapper() js.Func {
-	// TODO: figure out a better way to surface a golang error to JS environment.
-	// Currently error is surfaced as a string.
-	jsonFunc := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+func resourceListProcessorWrapper(resourceList *[]byte) js.Func {
+	jsonFunc := js.FuncOf(func(this js.Value, args []js.Value) any {
 		if len(args) != 1 {
 			return "Invalid number of arguments passed"
```
wleese, line 47: Perhaps this entire function could simply either return the resourceList or empty string, always deferring to the other function to get error messages, rather than mixing those approaches. Would require an update to https://github.com/GoogleContainerTools/kpt/pull/3830 if we take that route.

functions/go/apply-setters/generated/docs.go
```diff
@@ -83,7 +81,7 @@ Invoke the function:
 
 Alternatively, setter values can be passed as key-value pairs in the CLI
 
-  $ kpt fn eval --image gcr.io/kpt-fn/apply-setters:unstable -- image=ubuntu replicas=3
+  $ kpt fn eval --image gcr.io/kpt-fn/apply-setters:unstable -- tag=1.16.2 nginx-replicas=3
```
yuwenma, line 84: qq: do we need to change this example?

functions/go/apply-replacements/run_js.go
yuwenma: That's a good point. I think deferring to the other function to get errors makes a lot of sense. Would like to hear @droot @mortent opinion on this.

sdowell:
Thank you for the PR! Could you add some unit/e2e tests to verify and document the behavior of this change?

functions/go/apply-replacements/run_js.go
```diff
@@ -37,19 +41,40 @@ func applyReplacements(input []byte) ([]byte, error) {
...
+	jsonFunc := js.FuncOf(func(this js.Value, args []js.Value) any {
+		rl, err := fn.ParseResourceList(*resourceList)
+		if err != nil {
+			return ""
+		}
+		if len(rl.Results) == 0 {
+			return ""
+		}
+		errorMessages := ""
+		for _, r := range(rl.Results) {
```
yuwenma, line 72: When running in a pipeline, the function only appends `Results` to `ResourceList.Results` but not override the existing ones. So this  `resourceListProcessorErrors ` may get irrelevant error `Results` (from previous functions' results). 

I'd suggest at least adding a comment to notify future contributors about this since it could cause some unexpected failures in certain risk conditions.

yuwenma:
Thank you so much for making the change. The PR makes a lot of sense to me. I just have a comment about the potential risk conditions to pull out the `ResourceList.Results` errors.

Can you also share the your testing steps (test data, command to reproduce)  in the PR description?

yuwenma:
cc @natasha41575

wleese:
> Thank you so much for making the change. The PR makes a lot of sense to me. I just have a comment about the potential risk conditions to pull out the `ResourceList.Results` errors.
> 
> Can you also share the your testing steps (test data, command to reproduce) in the PR description?

I've added a copy & paste from the linked issue at the top of this PR, as those resources can be used to reproduce.

functions/go/apply-setters/generated/docs.go
wleese: (I had no intention of making this change btw. It probably was made when running `make`)

functions/go/apply-replacements/run_js.go
wleese: Added!

wleese:
> Thank you for the PR! Could you add some unit/e2e tests to verify and document the behavior of this change?

I can do this, although it will take some time (don't have a golang background)

natasha41575:
> Thank you for the PR! Could you add some unit/e2e tests to verify and document the behavior of this change?

@wleese If there is a good way to add unit tests, that would be nice, but I don't think we have an existing e2e framework for wasm functions and I don't want to block this PR on that. Adding e2e tests for wasm functions is a separate issue (https://github.com/GoogleContainerTools/kpt/issues/3587) that could take a while to do. So if you find yourself stuck on a good way to add tests, don't worry about spending too much time on it and IMO we can still accept this PR without them.

yuwenma:
> > > Thank you for the PR! Could you add some unit/e2e tests to verify and document the behavior of this change?
> 
> > I can do this, although it will take some time (don't have a golang background)
> 
> @wleese If there is a good way to add unit tests, that would be nice, but I don't think we have an existing e2e framework for wasm functions and I don't want to block this PR on that. Adding e2e tests for wasm functions is a separate issue ([GoogleContainerTools/kpt#3587](https://github.com/GoogleContainerTools/kpt/issues/3587)) that could take a while to do. So if you find yourself stuck on a good way to add tests, don't worry about spending too much time on it and IMO we can still accept this PR without them.

Good catch and suggestion, @sdowell and @natasha41575 Yeah, I think adding the test steps in the PR description is good enough for this PR. We'd like to help and make the contribution experience easier for active users and contributors.

yuwenma (approved):

wleese:
Thanks!

PR #1023, opened 2023-02-27
add local-config annotation to local resources in examples

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/3844

Our examples should have the `local-config` annotation on Kptfile and functionConfigs so that they can be used correctly out of the box.

### DIFF
--- a/examples/apply-replacements-simple/Kptfile
+++ b/examples/apply-replacements-simple/Kptfile
@@ -2,6 +2,8 @@ apiVersion: kpt.dev/v1
 kind: Kptfile
 metadata:
   name: example
+  annotations:
+    config.kubernetes.io/local-config: "true"
 pipeline:
   mutators:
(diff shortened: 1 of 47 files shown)


yuwenma (approved):
Nice Catch!!!

yuwenma (changes requested):
Can you also update the CI e2e test?
Basically the test runs against each example under `./example` folder to verify if the output is expected (under the `.expected/diff.patch`). With the Kptfile changes, the expected output would then no longer matches.

natasha41575:
> Can you also update the CI e2e test?

Done! Thanks for the pointer

natasha41575:
@yuwenma Tests are passing now! Could you please merge when you get the chance?

yuwenma (approved):
thanks for the quick fixes

PR #1024, opened 2023-03-01
fix issue with parsing helm output

natasha41575:
This fixes the issue reported by https://kubernetes.slack.com/archives/C0155NSPJSZ/p1677612509818639.

In the upstream kustomize helm plugin, we recently updated how we parse the output of running `helm template` due to some user-reported issues. This PR updates the render-helm-chart function to pick up that change from the upstream kustomize builtin. I have verified that the reported issue is fixed with this change, which confirms that the issue was due to some incorrect parsing logic.

Reproduction steps:
Build `render-helm-chart` locally. Then create an empty kpt package and run the local function.

```
# from the render-helm-chart directory
$ go install .
$ mkdir foo; kpt pkg init foo
$ kpt fn eval foo --exec=render-helm-chart --results-dir . --output ./output -- name=kube-prometheus-stack repo=https://prometheus-community.github.io/helm-charts version=45.3.0 releaseName=kube-prometheus-stack namespace=monitoring includeCRDs=true
```

You can see that when built from master, there is a parsing error; but when built off my branch here, the function runs successfully.


yuwenma (approved):

yuwenma:
Thanks for the fix! That's fair. I'll bump the render-helm-charts accordingly

PR #1027, opened 2023-03-02
updating helm to 3.11

jracollins:
Just been using the updated `render-helm-chart` version just released in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/1024 (Thanks for the super fast turnaround on this btw!)

I ran into some version issues in the traefik chart and noticed the helm version could do with an update https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/deployment.yaml#L2 |

### DIFF
--- a/functions/go/render-helm-chart/Dockerfile
+++ b/functions/go/render-helm-chart/Dockerfile
@@ -14,7 +14,7 @@ ARG TARGETOS TARGETARCH
 RUN GOOS=$TARGETOS GOARCH=$TARGETARCH go build -o /usr/local/bin/function ./
 
 RUN apk update && apk add --no-cache curl
-ARG HELM_VERSION="v3.8.0"
+ARG HELM_VERSION="v3.11.1"
 RUN curl -fsSL -o /helm-${HELM_VERSION}-${TARGETOS}-${TARGETARCH}.tar.gz https://get.helm.sh/helm-${HELM_VERSION}-${TARGETOS}-${TARGETARCH}.tar.gz && \
      tar -zxvf /helm-${HELM_VERSION}-${TARGETOS}-${TARGETARCH}.tar.gz && \
      mv ./${TARGETOS}-${TARGETARCH}/helm /usr/local/bin/helm


yuwenma:
@jracollins Thank you for the contribution. The change looks good. 

I just to ask some short questions (currently we do not catch up with dependent tool releases due to the releasing overhead and sustainability concerns). 
- A newer version 3.11.1 (instead o f 3.11.0) is released to [patch a CVE fix](https://github.com/helm/helm/releases/tag/v3.11.1) , I'd suggest using that one instead.
- I assume helm has some backward compatibility support for major versions, so using the old version (3.8.0) won't break helm but does a no-op for certain behaviors which are added in the newer versions (3.11.0). Is this correct? 
- For the `traefik` specific version requirement, do you mind elaborating what the issues you encountered? Just to help future customers who use older render-helm-chart versions, and to help us better evaluate the support.

yuwenma (approved):

jracollins:
Not sure how I missed the `3.11.1` - fixed.

So the actual error I ran into was the one I linked in https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/deployment.yaml#L2 rather than anything unexpected and was: 

```
[FAIL] "gcr.io/kpt-fn/render-helm-chart:unstable" in 300ms
  Stderr:
    "failed to evaluate function: failed to run function: failed to run generator: unable to run: 'helm template' with env=[]; \"exit status 1\": \"Error: execution error at (traefik/templates/deployment.yaml:3:8): ERROR: Helm >= 3.9.0 is required\\n\\nUse --debug flag to render out invalid YAML\\n\""
  Exit code: 1
```

I had a look at the changelog and not sure specifically what changes in `3.9` that the traefik chart depends on. [`3.9` is quite old now (May 2022)](https://github.com/helm/helm/releases/tag/v3.9.0) but if adhering to semver properly, it should just be additional features rather than any breaking changes.

yuwenma:
cc @natasha41575 (in case kustomize users run into the same issue)

PR #1037, opened 2023-04-03
docs: Update tags for go/render-helm-chart/v0.2.3

natasha41575:


droot (approved):

natasha41575:
It looks like the CD job for pushing images is broken: https://github.com/GoogleContainerTools/kpt-functions-catalog/actions/workflows/after-tag-with-version.yaml

Will merge this PR after getting guidance on how to push to the image

sdowell:
I think that is expected to fail https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/952/files#r1024392970

cc @justinsb

natasha41575:
@sdowell Do you have any context on why the CD job exists if it is expected to fail?

natasha41575:
@sdowell Thanks for the context! So is it expected to fail but still push the image to the container registry? Seems like the new render-helm-chart function tag didn't get pushed either

sdowell:
IIRC @justinsb was working on a new CD approach that was designed for a subset of functions he was working on. I don't have much more context than that, but it looks like that workflow doesn't work for any of the older functions which use the previous pattern of the Makefile in the parent directory.

I don't recall how exactly the old release flow works, I think there is an internal release pipeline that builds/publishes the images

PR #1064, opened 2023-09-13
[Feature] Set labels conditional spec selector

kemv:
## The problem

I am facing a use-case where i would like to set common labels to all resources in my Kpt pipeline. To do this, i have tried using set-labels.

One of the common labels I want to set on all resources is [app.kubernetes.io/version](http://app.kubernetes.io/version).

The problem is, set-labels also sets .spec.selector labels on deployments, and this field is immutable. This means, that when the version changes, all deployments have to be deleted and created again.

I propose some changes to have a switch to tell set-labels that it shouldn't set `.spec.selectors`. I have achieved this through the `SetLabels` kind already defined:

```yaml
kind: SetLabels
metadata:
  name: set-labels
  annotations:
labels:
  app.kubernetes.io/version: "some-version"
options:
  setSelectorLabels: false # defaults to true
```

## Considerations

I think that keeping backwards compatibility is very important. Therefore the option always defaults to true. Meaning if you change nothing, the functaionality doesn't change, even with this new feature. I haven't added this new functionality to the `ConfigMap` approach. This is because it would be harder to make this approach backwards compatible. The same with the imperative approach.

I have added a new e2e test for this functionality. I don't think this small functionality merits a new example page, but i have expanded the main README.md for the function.


kemv:
@droot @mengqiy @sdowell @yuwenma Seems CODEOWNERS is broken? Or perhaps PRs are closed until migrated to CNCF ?

PR #1144, opened 2025-05-26
Update go version to 1.24 for github workflows

liamfallon:
Github workflows are failing on go 1.19, upgrading to the latest go version 1.24


liamfallon:
Yes, I have to go back to the drawing board on it. I think that the kpt repo will have to be working before we can get this repo working.

PR #1161, opened 2025-08-22
Update push targets to ghcr for latest

efiacor:
Hefty PR. All tested and working on this detached fork - https://github.com/efiacor/krm-functions-catalog

Main focus of this PR:

On push to master branch, gh action triggers a push of ALL go based images to packages section (ghcr) with the following tags
- "latest"
- "$SHORT_GIT_SHA"

Other changes include:

Updating the e2e tests to use the new images and fix invalid expected values.
Refactoring some krm fns to build correctly.
Begin removing support for TypeScript images:
  - Remove CI testing etc.


efiacor:
@CsatariGergely verify-docs job should be disabled.
ts-unit-test-ci has been removed but still remains as "required"

efiacor:
> > @CsatariGergely verify-docs job should be disabled. ts-unit-test-ci has been removed but still remains as "required"
> 
> I think you just need to remove this line https://github.com/kptdev/krm-functions-catalog/blob/master/.github/workflows/verify-docs.yaml#L6 to disable verify-docs for pr-s against master.
> 
> ts-unit-test-ci (ubuntu-latest, 18.x) is still a required check in the settings. I can remove it if you think that it is not needed.

We plan to remove the TypeScript support so we should prob remove it from the required set.

liamfallon (approved):

PR #1227, opened 2026-03-20
update e2e expected outputs for kpt status block changes

aravindtga:
## Summary

Updates all e2e test expected outputs to account for `kpt@main` now writing `status.conditions` into `Kptfile` after `fn render`.

## What changed

`kpt` now emits a `status` block in `Kptfile` after rendering (As part of https://github.com/kptdev/kpt/pull/4418):

```yaml
status:
 conditions:
    - type: Rendered
     status: "True"
     reason: RenderSuccess
```

This caused all e2e tests to fail because the expected `diff.patch` files didn't include this new output.


liamfallon (commented):
I only saw this message on test fails. Just wondering why we only see this error:

```
pkg.render: pkg .:
+        	pipeline.run: already handled error
```

aravindtga:
> I only saw this message on test fails. Just wondering why we only see this error:
> 
> ```
> pkg.render: pkg .:
> +        	pipeline.run: already handled error
> ```

The "already handled error" is an internal sentinel error used by the function runner - when a function fails, the actual error details are printed to CLI output, and this sentinel is returned to signal "already printed, just exit non-zero." 
It was never meant to be serialized to a user-visible string, but now that we write status conditions to the Kptfile, it gets exposed.

We could iterate on this to extract the actual error and populate the condition message. However, the renderStatus field will capture the detailed information about each pipeline step, including the function image, exit code, stderr, and structured error results. So the condition message becomes somewhat redundant for diagnosing failures.  
If required, we can populate this with renderStatus.errorSummary as part of the renderStatus PR https://github.com/kptdev/kpt/pull/4437.

liamfallon:
> > I only saw this message on test fails. Just wondering why we only see this error:
> > ```
> > pkg.render: pkg .:
> > +        	pipeline.run: already handled error
> > ```
> 
> The "already handled error" is an internal sentinel error used by the function runner - when a function fails, the actual error details are printed to CLI output, and this sentinel is returned to signal "already printed, just exit non-zero." It was never meant to be serialized to a user-visible string, but now that we write status conditions to the Kptfile, it gets exposed.
> 
> We could iterate on this to extract the actual error and populate the condition message. However, the renderStatus field will capture the detailed information about each pipeline step, including the function image, exit code, stderr, and structured error results. So the condition message becomes somewhat redundant for diagnosing failures. If required, we can populate this with renderStatus.errorSummary as part of the renderStatus PR [kptdev/kpt#4437](https://github.com/kptdev/kpt/pull/4437).

That's fine @aravindtga , I was just curious.

liamfallon (approved):

liamfallon (approved):

efiacor (approved):

PR #1229, opened 2026-04-02
 Fix create-setters matching setter values in internal kpt annotations

aravindtga:
Fixes https://github.com/kptdev/kpt/issues/4462

### Summary

The create-setters function was incorrectly creating setter comments on `internal.config.kubernetes.io/*` and `config.kubernetes.io/*` annotations. These are kpt runtime metadata injected during fn render (e.g. filesystem paths) and are not user-authored resource content. When a setter value like 4 appeared as a substring in a temp directory path, the function would produce spurious and non-deterministic results.

### Changes

functions/go/create-setters/createsetters/create_setters.go:
- Added a path check in visitScalar() to skip fields under `internal.config.kubernetes.io` and `config.kubernetes.io annotations

functions/go/create-setters/createsetters/create_setters_test.go:
- Added test case skip internal kpt annotations verifying that:
  - `internal.config.kubernetes.io/package-path` containing the setter value is not matched
  - config.kubernetes.io/local-config` is not matched
  - Regular fields like `spec.replicas` still correctly receive setter comments

### DIFF
--- a/functions/go/create-setters/createsetters/create_setters.go
+++ b/functions/go/create-setters/createsetters/create_setters.go
@@ -248,6 +248,12 @@ func (cs *CreateSetters) visitScalar(object *yaml.RNode, path string) error {
 		return nil
 	}
 
+	// skip internal kpt annotations — these are runtime metadata, not user-authored content
+	if strings.Contains(path, ".annotations.internal.config.kubernetes.io") ||
+		strings.Contains(path, ".annotations.config.kubernetes.io") {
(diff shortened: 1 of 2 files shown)


aravindtga:
This should be rebased once https://github.com/kptdev/krm-functions-catalog/pull/1227 is merged.

liamfallon (approved):

mozesl-nokia (approved):

PR #1275, opened 2026-07-09
Update central go version to 1.26.5

mozesl-nokia:
Updated the go version for the e2e test and documentation module to use 1.26.5 and the base docker image will now automatically pick up the latest go patch version.

Also:
- Added default values to BUILDER_IMAGE and BASE_IMAGE to the default Dockerfile to get rid of most warnings (did not do this for all the custom Dockerfiles).
- Added a tool.go file to documentation so that the docsy dependency is not removed randomly by some tool.
- Updated the kpt version of the e2e test

AI was *not* used in the creation of this PR.


liamfallon (dismissed):

liamfallon (approved):

tests/e2etest/go.mod
```diff
@@ -1,25 +1,18 @@
 module github.com/kptdev/krm-functions-catalog/tests/e2etest
 
-go 1.26.3
+go 1.26.5
```
aravindtga, line 3: Should we also consider upgrading the Go version for the KRM functions?

mozesl-nokia: We should do that separately IMO

aravindtga (approved):

PR #1281, opened 2026-07-14
Generate docs for apply-setters v0.3

mozesl-nokia:
Generated docs for the new v0.3 version.

No AI was used to create this PR.


aravindtga (approved):

liamfallon (approved):

efiacor:
Just a query regarding the 0.x version bump. Are there NBC changes that we should document?

mozesl-nokia:
@efiacor I don't think so, I just thought we should bump the minor version since it is quite a big change. The output format did actually change, but only with additions (see the e2e patch changes).

